PR #276, opened 2015-01-22
Minor germline fixes

arahuja:
- Remove count step
- Normalize probabilties before output
- Revert from log-space before computing phred-scaled likelihood

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/276)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/GermlineStandardCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/GermlineStandardCaller.scala
@@ -54,6 +54,7 @@ object GermlineStandard {
       Common.progress(
         "Loaded %,d mapped non-duplicate reads into %,d partitions.".format(readSet.mappedReads.count, readSet.mappedReads.partitions.length))
 
+
       val loci = Common.loci(args, readSet)
       val lociPartitions = DistributedUtil.partitionLociAccordingToArgs(args, loci, readSet.mappedReads)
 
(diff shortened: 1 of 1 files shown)


src/main/scala/org/hammerlab/guacamole/commands/GermlineStandardCaller.scala
```diff
@@ -51,8 +51,6 @@ object GermlineStandard {
 
       val readSet = Common.loadReadsFromArguments(args, sc, Read.InputFilters(mapped = true, nonDuplicate = true))
       readSet.mappedReads.persist()
-      Common.progress(
-        "Loaded %,d mapped non-duplicate reads into %,d partitions.".format(readSet.mappedReads.count, readSet.mappedReads.partitions.length))
```
timodonnell: does removing this speed it up much? I thought (hoped) that since the reads are persisted, counting them should be fast

timodonnell:
Assuming the speed impact of removing the count is significant, this LGTM

src/main/scala/org/hammerlab/guacamole/commands/GermlineStandardCaller.scala
arahuja: There is not that much of a speed-up in dropping , the count was around 30-40 seconds of time, so we can leave it in.

timodonnell: Gotcha, that seems borderline so either way works for me

arahuja:
I added the count back in for now - if it end's being more problematic on large datasets we can remove, or have a debug mode where print intermediate counts.

PR #277, opened 2015-01-26
Fixing check of standard bases

arahuja:
These functions are duplicated.  believe the first, `allStandardBases`, should return a boolean and the second should throw an error on non-standard bases

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/277)


ryan-williams:
lgtm

PR #278, opened 2015-02-16
Add rebuilding of reference from multiple reads and tests

arahuja:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/278)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/reads/MDTagUtils.scala
+++ b/src/main/scala/org/hammerlab/guacamole/reads/MDTagUtils.scala
@@ -1,7 +1,7 @@
 package org.hammerlab.guacamole.reads
 
 import debox.Buffer
-import htsjdk.samtools.{Cigar, CigarOperator}
+import htsjdk.samtools.{ Cigar, CigarOperator }
 import org.bdgenomics.adam.util.MdTag
 import org.hammerlab.guacamole.Bases
 
@@ -18,10 +18,12 @@ object MDTagUtils {
    *
    * @param readSequence The base sequence of the read.
    * @param cigar The cigar for the read.
-   * @param referenceFrom The starting point of this read alignment vs. the reference.
    * @return A sequence of bytes corresponding to the reference overlapping this read.
    */
-  def getReference(mdTag: MdTag, readSequence: Seq[Byte], cigar: Cigar, referenceFrom: Long, allowNBase: Boolean): Seq[Byte] = {
+  def getReference(mdTag: MdTag,
+                   readSequence: Seq[Byte],
+                   cigar: Cigar,
+                   allowNBase: Boolean): Seq[Byte] = {
 
     var referencePos = mdTag.start
     var readPos = 0
@@ -79,8 +81,9 @@ object MDTagUtils {
     reference.toArray
   }
 
-  def getReference(read: MappedRead, allowNBase: Boolean): Seq[Byte] = {
-    getReference(read.mdTag, read.sequence, read.cigar, read.start, allowNBase)
+  def getReference(read: MappedRead, allowNBase: Boolean): (Long, Seq[Byte]) = {
+    val readReferenceStart = read.mdTag.start
+    (readReferenceStart, getReference(read.mdTag, read.sequence, read.cigar, allowNBase))
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/scala/org/hammerlab/guacamole/reads/MDTagUtils.scala
```diff
@@ -75,7 +76,42 @@ object MDTagUtils {
...
+    reference.toArray
+  }
+
+  def getReference(read: MappedRead, allowNBase: Boolean): Seq[Byte] = {
+    getReference(read.mdTag, read.sequence, read.cigar, read.start, allowNBase)
+  }
+
+  /**
+   * Rebuilds the reference from a set of overlapping and sorted reads
+   * Fill in N if their is a gap in the reads
```
ryan-williams: `s/their/there/`

arahuja: woops... will fix, also not sure why the test are not passing, but doesn't seem like any fail it just dies.  will push a force fix so it should re-test

src/main/scala/org/hammerlab/guacamole/reads/MDTagUtils.scala
```diff
@@ -75,7 +76,42 @@ object MDTagUtils {
...
-    reference.elems
+    reference.toArray
+  }
+
+  def getReference(read: MappedRead, allowNBase: Boolean): Seq[Byte] = {
+    getReference(read.mdTag, read.sequence, read.cigar, read.start, allowNBase)
+  }
+
+  /**
+   * Rebuilds the reference from a set of overlapping and sorted reads
```
timodonnell: Some comments:
 * It feels weird that this function doesn't take a start and an end locus to return the reference sequence for. Is the idea that the caller would look at the reads first to know where it starts? That seems potentially confusing e.g. in the case of clipped reads.
 * If you do move to taking a start and an end locus, might as well build up the sequence directly in an array (instead of a buffer), to avoid the copy.
 * If multiple reads are overlapping a locus, it feels like we should assert that they agree on the reference bases.
 * Consider an assertion to force that the reads are sorted

timodonnell:
aside from my comment LGTM

arahuja:
Can you explain the first comment more? Right now, this builds the reference from the reads available, so it should start on the first reference base the first read covers.  Do you mean figuring out where it starts would be confusing in the case of clipped reads?

Also on the third comment - `If multiple reads are overlapping a locus, it feels like we should assert that they agree on the reference bases.` this shouldn't ever happen right?  This would really indicate an issue with the md tag?

timodonnell:
I'd expect that for at least some uses of this function, I wouldn't want the whole reference sequence spanned, but rather a specific interval of the reference. Having this function return the whole reference sequence spanned is less efficient (especially consider the possibility of long reads eventually), but more importantly forces the caller to do some math to figure out how to offset to a particular desired interval of loci. To me, it would seem clearer to take a start and an end loci and always return that exact number of bases. The caller can always pass in the full range of loci spanned by the reads if that is what is desired. If you don't see it this way, that's fine though -- just a suggestion.

Regarding the third comment, couldn't that happen if the reads are corrupt somehow (e.g. aligned to different references, or a bug in the aligner)? That seems like a good case to catch with an assert.

ryan-williams:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/278)







---

<sup>**[src/main/scala/org/hammerlab/guacamole/reads/MDTagUtils.scala, line 91 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/278#-JjBHVW7w82dprSM_lKm)** ([raw file](https://github.com/hammerlab/guacamole/blob/e1dedc8eb6e3e59cc55a8dade1be044b9a08239d/src/main/scala/org/hammerlab/guacamole/reads/MDTagUtils.scala#L91)):</sup>
`s/read/reads/`

---

ryan-williams:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/278)







---

<sup>**[src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala, line 60 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/278#-JjBTMRkUiL7nnlO3rjp)** ([raw file](https://github.com/hammerlab/guacamole/blob/e1dedc8eb6e3e59cc55a8dade1be044b9a08239d/src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala#L60)):</sup>
I think the `G` in `1C5G2` matches the reference here; is that intentional?

Also I'm going to send you a PR with ASCII pileups of some of these added as comments :) lmk if that's a bad idea

---

ryan-williams:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/278)







---

<sup>**[src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala, line 70 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/278#-JjBTsMmrcEodhZoAq5Q)** ([raw file](https://github.com/hammerlab/guacamole/blob/e1dedc8eb6e3e59cc55a8dade1be044b9a08239d/src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala#L70)):</sup>
should this `2M3I7M` be `2M3I8M`?

---

ryan-williams:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/278)







---

<sup>**[src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala, line 80 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/278#-JjBa9zXzZdhNgsP7klQ)** ([raw file](https://github.com/hammerlab/guacamole/blob/e1dedc8eb6e3e59cc55a8dade1be044b9a08239d/src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala#L80)):</sup>
ditto, `s/7M/8M/`, also the `A` and `G` in `1A5G2` seem to match the reference?

---

ryan-williams:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/278)



---

I left a few comments about the tests seeming to unexpectedly not be testing actual deviations from the reference; let me know if I'm off-base there, I stopped and filed #281 with a few of them. I can fix the rest if you don't want to, lmk

---

arahuja:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/278)







---

<sup>**[src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala, line 60 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/278#-JjBTMRkUiL7nnlO3rjp)** ([raw file](https://github.com/hammerlab/guacamole/blob/e1dedc8eb6e3e59cc55a8dade1be044b9a08239d/src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala#L60)):</sup>
Replied on your PR, but this is wrong, but in the other way.  It should 1A5G2 as the MDtag should show whats in the reference.

---

<sup>**[src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala, line 70 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/278#-JjBTsMmrcEodhZoAq5Q)** ([raw file](https://github.com/hammerlab/guacamole/blob/e1dedc8eb6e3e59cc55a8dade1be044b9a08239d/src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala#L70)):</sup>
Yea it should

---

<sup>**[src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala, line 80 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/278#-JjBa9zXzZdhNgsP7klQ)** ([raw file](https://github.com/hammerlab/guacamole/blob/e1dedc8eb6e3e59cc55a8dade1be044b9a08239d/src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala#L80)):</sup>
Yea correct on s/7M/8M and see above on the mdtag should match the reference

---

ryan-williams:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/278)



---

to be clear, you still have more changes you're going to make here right?

---

arahuja:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/278)



---

Yes, for sure, going to incorporate Tim's comments and still need to fix some of issues you pointed out.

---

arahuja:
@timodonnell @ryan-williams this now updated
- takes a reference start and end position
- assert to check for sorted reads
- updated diagrams/md tags for all the tests

ryan-williams:
lgtm!

ryan-williams:
merge when you're ready @timodonnell @arahuja

PR #279, opened 2015-02-17
Add overlaps for other reference region

arahuja:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/279)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/HasReferenceRegion.scala
+++ b/src/main/scala/org/hammerlab/guacamole/HasReferenceRegion.scala
@@ -51,5 +51,14 @@ trait HasReferenceRegion {
   def overlapsLocus(locus: Long, halfWindowSize: Long = 0): Boolean = {
     start - halfWindowSize <= locus && end + halfWindowSize > locus
   }
+
+  /**
+   * Does the region overlap another reference region
+   * @param other another region on the genome
+   * @return True if the the regions overlap
+   */
+  def overlaps(other: HasReferenceRegion): Boolean = {
+    other.referenceContig == referenceContig && (overlapsLocus(other.start) || overlapsLocus(other.end - 1))
+  }
 }
 
--- a/src/main/scala/org/hammerlab/guacamole/HasReferenceRegion.scala
+++ b/src/main/scala/org/hammerlab/guacamole/HasReferenceRegion.scala
@@ -58,7 +58,7 @@ trait HasReferenceRegion {
    * @return True if the the regions overlap
    */
   def overlaps(other: HasReferenceRegion): Boolean = {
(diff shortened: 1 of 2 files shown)


src/main/scala/org/hammerlab/guacamole/HasReferenceRegion.scala
```diff
@@ -51,5 +51,14 @@ trait HasReferenceRegion {
...
     start - halfWindowSize <= locus && end + halfWindowSize > locus
   }
+
+  /**
+   * Does the region overlap another reference region
+   * @param other another region on the genome
+   * @return True if the the regions overlap
+   */
+  def overlaps(other: HasReferenceRegion): Boolean = {
+    other.referenceContig == referenceContig && (overlapsLocus(other.start) || overlapsLocus(other.end - 1))
```
timodonnell: what if the other region completely encloses the current one? does this give the wrong result in that case?

e.g.
```
other: ------------XXXXXXXXXXXX----------
self:  -----------------XXXX--------------------

**Comment by timodonnell** [comment]
sorry for neglecting this for so long. if the issue I noted is indeed a bug, then I think we should add a test for that case in addition to fixing it. beside that LGTM

**Comment by arahuja** [comment]
Yea thanks - I had missed that, added a test and fixed the function

File: src/test/scala/org/hammerlab/guacamole/HasReferenceRegionSuite.scala [hunk 2]
```diff
@@ -0,0 +1,53 @@
...
+    val read2 = TestUtil.makeRead("TCGATCGA", cigar = "8M", mdtag = "8", start = 5L)
+
+    read1.overlaps(read2) should be(true)
+    read2.overlaps(read1) should be(true)
+
+  }
+
+  test("overlapping reads, different contigs") {
+    val read1 = TestUtil.makeRead("TCGATCGA", cigar = "8M", mdtag = "8", start = 1L)
+    val read2 = TestUtil.makeRead("TCGATCGA", cigar = "8M", mdtag = "8", start = 5L, chr = "chr2")
```
ryan-williams: there's an implied `{1-8,5-13} --> true` case here that it probably makes sense to add a case explicitly of

arahuja: @ryan-williams sorry for not looking at this sooner.  I don't think I understand what you mean - isn't the 1-8, 5-13 the test above where they are on the same chromosome?

ryan-williams: huh, yea.. I guess I misread / not sure what I was thinking. disregard

PR #281, opened 2015-02-27
a few testing tweaks / comments

ryan-williams:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/281)

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala
@@ -36,6 +36,13 @@ class MDTagUtilsSuite extends FunSuite with Matchers {
   test("rebuild reference from multiple reads") {
     val originalReference = "AAATTGATACTCGAACGA"
 
+    /*
+        ref: AAATTGATACTCGAACGA
+         r1: AAATTGATAC
+         r2:      GATACTCGAA
+         r3:         ACTCGAACGA
+     */
+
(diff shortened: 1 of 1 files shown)


src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala
```diff
@@ -56,8 +71,15 @@ class MDTagUtilsSuite extends FunSuite with Matchers {
...
+
+    /*
+        ref: AAATTGATACTCGAACGA
+         r1: AAATTGATAC
+         r2:      GCTACTCAAA
+         r3:         ACTCGAACGA
+     */
     val firstRead = TestUtil.makeRead(originalReference.substring(0, 10), "10M", "10", start = 0)
-    val secondRead = TestUtil.makeRead("GCGGGTACTCGAA", "10M", "1C5G2", start = 5)
+    val secondRead = TestUtil.makeRead("GCGGGTACTCAAA", "10M", "1C5A2", start = 5)
```
arahuja: This is actually my mistake the first time, the MDTag should be 1A5G2.  It should say what the reference bases are.

arahuja:
I'm going to merge this into my branch and we can continue there - thanks for the diagrams, makes it much easier to see!

src/test/scala/org/hammerlab/guacamole/reads/MDTagUtilsSuite.scala
ryan-williams: ah yea good cal

PR #282, opened 2015-03-15
Fix scripts/guacamole to run correct jar

timodonnell:
Invoke target/guacamole-with-dependencies-*-.jar, not target/guacamole-*.jar.

Otherwise, we were seeing errors when running scripts/guacamole.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/282)


arahuja:
Yea, the jars were renamed after release PR so `guacamole.jar` is smaller and can be uploaded to maven.

Thanks!

Issue #287, opened 2015-05-03
read evidence analysis


timodonnell:
It could be useful to have an analysis that takes in a collection of BAM files and variants (specified in a set of VCF files) and writes out a csv with, for each BAM file, locus, the number of reads with each allele. This would be after applying some configurable read filters.

Currently, I have a Python [script](https://github.com/hammerlab/ovarian-cancer/blob/master/projects/pt189/read_evidence_plot.py) for generating this data for pt189 (something like 23 BAMs and ~1k variants). I ran it on all pt189 datasets like this:

```
time python read_evidence_plot.py \
    datasets.json \
    --evidence-summary-out out_full.csv \
    --extra-loci-offsets -1 \
    --sefara-set box_path_prefix=~/Box\ Sync/ \
```

(the `--extra-loci-offsets` argument tells it to, besides querying each locus with a variant, also query the locus before the variant, so we can plot non-variant loci near each variant).

It took 87 minutes to generate this 143k line output file:

```
(venv-2.7)[odonnt02@hammerlab-dev3 ~/sinai/git/ovarian-cancer/projects/pt189]$ head /hpc/users/odonnt02/sinai/git/ovarian-cancer/projects/pt189/out_full.csv
sample,contig,start,end,allele,count
bam_normal_exome_illumina,2,91673898,91673899,A,19
bam_normal_exome_illumina,2,91673897,91673898,T,19
bam_normal_exome_illumina,2,92315419,92315420,A,2
bam_normal_exome_illumina,2,92315419,92315420,C,4
bam_normal_exome_illumina,2,92315419,92315420,T,2629
bam_normal_exome_illumina,2,92315419,92315420,G,72
bam_normal_exome_illumina,2,92315418,92315419,A,1
bam_normal_exome_illumina,2,92315418,92315419,C,2
bam_normal_exome_illumina,2,92315418,92315419,T,92
(venv-2.7)[odonnt02@hammerlab-dev3 ~/sinai/git/ovarian-cancer/projects/pt189]$
```

I'd be interested if a Spark analysis for this could be significantly faster than 87 minutes.

arahuja:
I'd be up for taking this on.

One question I have is why the out put is `bamfile`, `pos`, `counts` as opposed to each header being a BAM file?  I understand its pretty each to pivot it, but if this 100k+ variants, and you easily want to grab all the data for a single variant, having it pre-aggregated by position may be useful?

timodonnell:
[...]

PR #288, opened 2015-05-05
Variant support/read evidence module

arahuja:
This computes the count per allele at a set of variant sites.

Future TODOs:
- Support filtering the VCF to only passing variants
- Support not requiring non-MDTag reads. 

The second one will require changing other pieces of code as well, so I will do that in a second PR and after some discussion of what that looks like.  Simplest thing is an `Option[MdTag]`, but that usually means that each pileup element also will have some option for the `referenceBases` in `Allele`.

Sample output:
```
Normal_Exome, chr1, 10340, A, A, 8
Normal_Exome, chr1, 13011, G, G, 18
Normal_Exome, chr1, 13115, T, G, 4
Normal_Exome, chr1, 13115, T, T, 2
Normal_Exome, chr1, 13117, A, G, 4
Normal_Exome, chr1, 13117, A, A, 2
Normal_Exome, chr1, 13280, C, C, 10
Normal_Exome, chr1, 13301, C, C, 7
```

This takes about 5min our cluster on 2 exome BAMs.  There are some smaller overhead that we can look into (there is a variant count print (~20s)) and if we had pre-converted BAMs to parquet we would be able to filter to the correct loci efficiently (without a full-scan/conversion of each BAM record).

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/288)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/Common.scala
+++ b/src/main/scala/org/hammerlab/guacamole/Common.scala
@@ -84,7 +84,7 @@ object Common extends Logging {
     }
 
     /** Argument for writing output genotypes. */
-    trait Output extends Base {
+    trait GenotypeOutput extends Base {
       @Opt(name = "--out", metaVar = "VARIANTS_OUT", required = false,
         usage = "Variant output path. If not specified, print to screen.")
       var variantOutput: String = ""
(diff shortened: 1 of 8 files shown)


ryan-williams:
seems reasonable to me!

hammer:
@arahuja anything preventing the merge? Also, file the TODOs as issues?

arahuja:
Mainly the travis issue - keeps hanging for some reason in the middle the tests but runs through locally.  Wanted to look into that.

arahuja:
Crazy now its happy after re-running
Closes #287

PR #299, opened 2015-06-18
Remove getMappedReadOpt property

arahuja:
Not sure if there was another reason to have this, but for it's current uses we can check the type.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/299)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/reads/Read.scala
+++ b/src/main/scala/org/hammerlab/guacamole/reads/Read.scala
@@ -61,13 +61,13 @@ trait Read {
   val isDuplicate: Boolean
 
   /** Is this read mapped? */
-  final val isMapped: Boolean = getMappedReadOpt.isDefined
+  final val isMapped: Boolean = this match {
+    case r: MappedRead   => true
+    case r: UnmappedRead => false
+  }
 
   /** The sample (e.g. "tumor" or "patient3636") name. */
   val sampleName: String
-
-  /** Returns this Read as a MappedRead iff isMapped=true, otherwise None. */
-  def getMappedReadOpt: Option[MappedRead] = None
 
   /** Whether the read failed predefined vendor checks for quality */
   val failedVendorQualityChecks: Boolean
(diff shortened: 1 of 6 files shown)


src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -61,14 +61,14 @@ trait Read {
   val isDuplicate: Boolean
 
   /** Is this read mapped? */
-  final val isMapped: Boolean = getMappedReadOpt.isDefined
+  final val isMapped: Boolean = this match {
```
ryan-williams: semi-related thought: should we `@transient` some/all fields in classes that get ser'd on the critical path, e.g. `Read`, `Pileup`?

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -119,40 +119,27 @@ object Read extends Logging {
...
     failedVendorQualityChecks: Boolean = false,
     isPositiveStrand: Boolean = true,
-    matePropertiesOpt: Option[MateProperties] = None): Read = {
+    matePropertiesOpt: Option[MateProperties] = None) = {
 
     val sequenceArray = sequence.map(_.toByte).toArray
     val qualityScoresArray = baseQualityStringToArray(baseQualities, sequenceArray.length)
 
-    if (referenceContig.isEmpty) {
-      UnmappedRead(
```
ryan-williams: does this mean we only allow `MappedRead`s now?

arahuja: Yes, actually it does, but we were only using it to make `MappedReads` anyways.

`UnmappedRead`'s constructor could be used directly instead.

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
arahuja: What effect would that have? Or more as a note to developers?

ryan-williams: I was thinking that it would improve performance since we would serialize less, but I guess we're manually controlling what gets kryo-serialized so nvm

ryan-williams:
Lgtm

PR #300, opened 2015-06-18
Remove mapping complexity and regional analysis filters

arahuja:
Mostly code deletion, after a lot of the validation tests these filters did not add much and actually were wrong in many cases.  These were throwing out variants based on the synthetic rules in SMC challenge that we should not do.

This may result in some tests failing in the SomaticStandardSuite, but those tests can also be cleaned up.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/300)


ryan-williams:
LGTM

PR #301, opened 2015-06-22
Small fixups

danvk:
Two small changes:

- Delete the `.reviewboardrc` file. We've never really used it.
- Switch to out-of-line links in `README.md`, which we seem to prefer these days.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/301)


ryan-williams:
thx @danvk, lgtm

PR #307, opened 2015-06-24
Check for output file existence first

### LINKED ISSUE #305: Check if output directory exists before running the job
I ran an entire Guacamole job and had it fail because the output directory (as specified by `--out`) already existed. This happened because a previous invocation left a temporary file there.

It would be better if Guacamole warned about this before kicking off any tasks on the cluster.

danvk:
This moves the failure before the entire job runs, rather than after.

It wasn't clear to me if this was an issue for JSON and Parquet output, so I'm only doing the check for VCF output. @arahuja @ryan-williams should I be doing the check for those, too?

I could move the `Common.validateArguments` call into `Command.scala` to reduce duplication. Thoughts?

Fixes #305 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/307)


ryan-williams:
lgtm!

PR #308, opened 2015-06-25
Extract read name from Alignments

danvk:
This is partially pulled from the Cornell code.

You get paired reads by grouping by read name, so this is a pre-req for doing structural variant calling.

Two questions:
- Are there performance concerns about parsing read names?
- I'm surprised how many files I had to modify to do this (5 main + 5 test). What's the point of the case classes and kryo serialization? Why not use the ADAM Avro format everywhere?

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/308)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/reads/MappedRead.scala
+++ b/src/main/scala/org/hammerlab/guacamole/reads/MappedRead.scala
@@ -45,7 +45,8 @@ case class MappedRead(
     mdTagString: String,
     failedVendorQualityChecks: Boolean,
     isPositiveStrand: Boolean,
-    matePropertiesOpt: Option[MateProperties]) extends Read with HasReferenceRegion {
+    matePropertiesOpt: Option[MateProperties],
+    readName: String) extends Read with HasReferenceRegion {
 
   assert(baseQualities.length == sequence.length,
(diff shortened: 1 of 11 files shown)


arahuja:
Some of this contrasts with what I was doing with `MateProperties` essentially, removing it from `MappedRead` and shifting it to a new class.

The string here will actually add a lot of memory overhead per read.  Is it necessary to group on readPair if you have the mate information?  Could we use add this to the `ReadPair` class here: https://github.com/hammerlab/guacamole/blob/remove-mate-prop/src/main/scala/org/hammerlab/guacamole/reads/ReadPair.scala

On the question re: Kryo serialization 
- Essentially allows us to manage what get's serialized - for example on the `MappedRead` we save/serialize the `mdTagString` since we often don't parse that before the shuffle, but not the `MdTag` itself.
- Using Avro may be fine in some cases.  I would be for moving MappedRead to Avro perhaps.  Some things are ugly through that, mostly handling `null` everywhere in the Java/Scala back and forth.  The ADAM Avro record has too much overhead in it compared to the `MappedRead`

danvk:
(Abandoning this in favor of `matePropertiesOpt`)

Issue #318, opened 2015-07-07
port varcode read evidence tests for ReadEvidence command


timodonnell:
The read evidence (aka "variant-support" -- we should actually change those two names to agree) command here: https://github.com/hammerlab/guacamole/blob/master/src/main/scala/org/hammerlab/guacamole/commands/ReadEvidence.scala has no tests.

The command itself is trivial, but it would be good to have some tests for it as a proxy for our pileup building infrastructure. Especially for RNA, I wouldn't be surprised if we're missing or messing up some cigar operators.

For the read evidence module in varcode, we put together a test set covering a lot of corner cases based on hand-comparing to IGV: https://github.com/hammerlab/varcode/blob/master/test/test_read_evidence.py

We should port those varcode tests over to Guacamole

arahuja:
@timodonnell Do you have a preference on a name, `ReadEvidence` vs `VariantSupport`? I had though `ReadEvidence` didn't say anything about evidence for what.  Maybe `VariantReadEvidence`

timodonnell:
I don't have a strong preference on the name, and I agree "ReadEvidence" is kind of meaningless. `VariantSupport` works for me.

PR #320, opened 2015-07-08
Test for RNA read data

arahuja:
As mentioned in #318, we do not have any tests for RNA read processing.

This PR:
- removes the `UNKNOWN_MAPPING_QUALITY` condition from `MappedRead`
- fixes rebuilding the reference when there are `N` CIGAR operators
- adds a sample RNA file and some tests for pileup processing

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/320)


timodonnell:
LGTM!

PR #322, opened 2015-07-14
Variant support tests

### LINKED ISSUE #318: port varcode read evidence tests for ReadEvidence command

arahuja:
Fixes #318 

This ports over _most_ of the tests from the `varcode` read evidence module.  `varcode` currently has more filter features, which are not available here so not all of the tests translated 1 to 1.

I tried the RNA tests as well, but they do not pass due to https://github.com/bigdatagenomics/adam/issues/726

I will add the RNA tests once that is fixed

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/322)


timodonnell:
LGTM -- thanks for doing this, @arahuja !

PR #323, opened 2015-07-14
VAF HIstogram improvements

### LINKED ISSUE #315: vaf histogram improvements
Some ideas:

 * output to csv with fields sample, start vaf, end vaf, count e.g. pt189_normal, 0.0, 0.1, 1000
 * accept an argument like --min-depth to control the minimum number of reads that must overlap a locus in order for it to be included

arahuja:
Fixes #315

- Improves VAF histogram output
- Adds filtering for pileups

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/323)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
@@ -1,5 +1,7 @@
 package org.hammerlab.guacamole.commands
 
+import java.io.{ BufferedWriter, FileWriter }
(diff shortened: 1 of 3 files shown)


arahuja:
@timodonnell Made some updates here:
- local output option with header
- some tests for the histogram creation
- for now made bins an option between 1 and 100, but we can update this later if you see a need for more precision

timodonnell:
Just ran like a charm for me. LGTM. Thanks @arahuja !

PR #324, opened 2015-07-15
Support pileup flatmapping over any number of RDDs

### LINKED ISSUE #317: support any number of RDDs in `windowTaskFlatMapMultipleRDDs`
see: https://github.com/hammerlab/guacamole/blob/master/src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala#L494

we'll need this for multisample joint somatic calling

### LINKED ISSUE #321: incorrect comment / can't run single unit test
We say:
```
// You can use a line like the following to turn on only a particular unit test for debugging.
  // TestUtil.runOnly = "partitionLociByApproximateReadDepth chr20 synth1 subset"
```
here:
https://github.com/hammerlab/guacamole/blob/master/src/test/scala/org/hammerlab/guacamole/DistributedUtilSuite.scala#L35
and also: https://github.com/hammerlab/guacamole/blob/master/src/test/scala/org/hammerlab/guacamole/util/TestUtil.scala#L46

however, setting neither of these for me limits the tests to a single case.

We should either fix this or delete those lines.

timodonnell:
Generalizes `windowTaskFlatMapMultipleRDDs` to handle any number of RDDs.

Closes #317. Also closes #321.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/324)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
+++ b/src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
@@ -341,7 +341,7 @@ object DistributedUtil extends Logging {
         val advancedPileups = maybePileups match {
           case Some(existingPileups) => {
             existingPileups.zip(windows).map(
-              pileup_and_window => initOrMovePileup(Some(pileup_and_window._1), pileup_and_window._2))
+              pileupAndWindow => initOrMovePileup(Some(pileupAndWindow._1), pileupAndWindow._2))
           }
           case None => windows.map(initOrMovePileup(None, _))
         }
--- a/src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
+++ b/src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
@@ -21,14 +21,13 @@ package org.hammerlab.guacamole
 import org.apache.commons.math3
 import org.apache.spark.rdd.RDD
 import org.apache.spark.{ Partitioner, AccumulatorParam, SparkConf, Logging }
-import org.apache.spark.SparkContext._
 import org.apache.spark.broadcast.Broadcast
 import org.apache.spark.serializer.JavaSerializer
 import org.hammerlab.guacamole.Common.Arguments.{ Base, Loci }
 import org.hammerlab.guacamole.Common._
 import org.hammerlab.guacamole.pileup.Pileup
 import org.hammerlab.guacamole.reads.MappedRead
-import org.hammerlab.guacamole.windowing.{ SlidingWindow }
+import org.hammerlab.guacamole.windowing.{ SplitIterator, SlidingWindow }
 import org.kohsuke.args4j.{ Option => Args4jOption }
 
 import scala.collection.mutable.{ HashMap => MutableHashMap }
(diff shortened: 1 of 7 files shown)


timodonnell:
This changes the code path for all the variations of `pileupFlatMap`, which is performance-sensitive code. I've run some basic tests on the cluster and things seem ok. But we should keep an eye out for any slowdowns or other unexpected effects of this change.

src/main/scala/org/hammerlab/guacamole/windowing/SplitIterator.scala
```diff
@@ -0,0 +1,62 @@
...
+   * i'th element in the sequence is an iterator over the values v where k = i.
+   *
+   * @param num number of keys. Each key k must be 0 <= k < num.
+   * @param iterator source iterator of (key, value) pairs. It will be consumed as the result iterators are read.
+   * @tparam V
+   * @return a buffered iterator for each key in 0 .. num.
+   */
+  def split[V](num: Int, iterator: Iterator[(Int, V)]): Seq[BufferedIterator[V]] = {
+    val buffers = (0 until num).map(_ => new collection.mutable.Queue[V])
+    (0 until num).map(index => new SplitIterator[V](index, buffers, iterator))
```
arahuja: Actually, I think this same functionality is possible with `.partition` on an Iterator?

```scala
var otherIterator = iterator
(0 until num).map(index => { 
       indexIterator, otherIterator = otherIterator.partition(_._1 == index)
       indexIterator
})
```

I think this does the same thing, but I did not/run test it.

timodonnell: Yeah, good call. IIRC we used to use `partition` in this way a while ago to split out each reads for each chromosome into a separate iterator, but the overhead of having to nest a `partition` for each key value was substantial (if there are N keys, each element can get buffered N times).

timodonnell: I think this is the PR that got rid of that pattern, so we have an example of what exactly was slow:

https://github.com/hammerlab/guacamole/commit/f557a3fd37b9808235889e4a4693914e20bb8a4e#diff-525c141f106b5f061f77693e79b5e974L558

src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
```diff
@@ -323,6 +323,34 @@ object DistributedUtil extends Logging {
...
+      readsRDDs,
+      lociPartitions,
+      skipEmpty,
+      halfWindowSize = 0L,
+      initialState = None,
+      function = (maybePileups: Option[Seq[Pileup]], windows: PerSample[SlidingWindow[MappedRead]]) => {
+        val advancedPileups = maybePileups match {
+          case Some(existingPileups) => {
+            existingPileups.zip(windows).map(
+              pileup_and_window => initOrMovePileup(Some(pileup_and_window._1), pileup_and_window._2))
```
arahuja: `pileup_and_window` style?

timodonnell: good call will fix

timodonnell:
updated to address comment

arahuja:
Sorry for the delay, LGTM

timodonnell:
rebased to master

PR #325, opened 2015-07-16
Move MateProperties to separate ReadPair class

### LINKED ISSUE #136: Ensure paired read information in type-safe way
Many of pair information fields are stored in options or with flags letting us know the read is paired - can we improve that cc @ryan-williams to expand.

arahuja:
- This moves out the information on the mate to a separate class `ReadPair`
- `ReadPair` is a `MappedRead` or `UnmappedRead` with the additional mate information.

This allows us to project just the `MappedRead` elements when we don't need the pair information and save on memory.  Also, this closes #136

cc @danvk: this might change your in-flight work a bit

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/325)


src/main/scala/org/hammerlab/guacamole/reads/ReadPair.scala
```diff
@@ -0,0 +1,41 @@
...
+import htsjdk.samtools.SAMRecord
+
+/**
+ * *
+ * @param read Unmapped or MappedRead base read
+ * @param isFirstInPair Whether the read is earlier that the the mate read
+ * @param mateAlignmentProperties Alignment location of the mate if it the mate is aligned
+ * @tparam T UnmappedRead or MappedRead
+ */
+case class ReadPair[+T <: Read](read: T,
```
timodonnell: This class name seems confusing to me since I'd expect a `ReadPair` object to contain two reads. What does this class represent? I'd suggest a header line in the doc summarizing what this is, and potentially renaming it to something clearer (e.g. `PairedRead`?)

arahuja: Yes, I agree `PairedRead` is a better name. It is a `Mapped/UnmappedRead` that is a paired end read and the alignment information of the mate.

timodonnell:
LGTM

src/main/scala/org/hammerlab/guacamole/reads/MateAlignmentProperties.scala
```diff
@@ -0,0 +1,51 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
```
danvk: We're licensing Guacamole code to BDG?

arahuja: That's what the rest of the code is so far at least

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -62,8 +62,9 @@ trait Read {
 
   /** Is this read mapped? */
   final val isMapped: Boolean = this match {
-    case r: MappedRead   => true
-    case r: UnmappedRead => false
+    case r: MappedRead       => true
```
danvk: Is there a reason to do it this way vs. implementing `isMapped` in each subclass? (that would be the more OO approach)

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -170,6 +169,7 @@ object Read extends Logging {
   def fromSAMRecordOpt(record: SAMRecord,
```
danvk: Not strictly related to this change, but what does `Opt` mean here? Optimized? Option? When would it return `None`?

src/main/scala/org/hammerlab/guacamole/reads/MateAlignmentProperties.scala
```diff
@@ -0,0 +1,51 @@
...
+import htsjdk.samtools.SAMRecord
+
+/**
+ * Details of the mate read alignment
+ * @param mateReferenceContig Contig/chromosome of the mate read
+ * @param mateStart 0-based start position of the mate read
+ * @param inferredInsertSize Insert size between the reads if defined
+ * @param isMatePositiveStrand Whether the mate is on the positive strand
+ */
+case class MateAlignmentProperties(mateReferenceContig: String,
```
danvk: Something that's bugged me while working with this type: why include `mate` in all the field names when it's clear from the type? I do sometimes wonder if these values are referring to the mate's mate's start.

danvk:
Looks good. This will require some changes in my code but it's for the best. Having a type for paired reads will simplify things.

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
arahuja: No, not really, I think that makes sense as well. I think we did have that at some point, not sure what was the reasoning behind this way.

 Will follow on with that

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
arahuja: Here it means it returns an Option - I think @ryan-williams introduced the convention - 

@ryan-williams this is something you had thought was pretty standard?

It returns `None` right now for `MappedRead` without an MDTag, which I think @timodonnell wants to address

src/main/scala/org/hammerlab/guacamole/reads/MateAlignmentProperties.scala
arahuja: Ah interesting, good point! Most of these are holdovers from `SAMRecord` but happy to drop.

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
danvk: The docstring should mention that behavior then. In general our [style guide][1] says to avoid abbrevs in method names.

[1]: https://hammerlab.atlassian.net/wiki/display/MSL/Software+Development#SoftwareDevelopment-General

PR #328, opened 2015-07-16
Standardize opt to Args4jOption

### LINKED ISSUE #306: Standardize on @Opt or @Option
I ran `git grep @Option` to find all the command line flags. It took me a minute to realize that some classes alias `Option` to `Opt`. Saving three characters doesn't seem particularly useful—we should just use one or the other.

arahuja:
This moves all `Option` from `Args4J` to `Args4jOption` to avoid colliding with scala `Option` and to standardize across the codebase.

Closes #306

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/328)


timodonnell:
LGTM

PR #334, opened 2015-07-22
Find regions with abnormally large inserts.

danvk:
This is pretty simple and will conflict with @arahuja's #325, but it is at least able to find interesting regions  in synth4.

For example, here's one of the output lines when I run it on synth4 tumor:

    GenomeRange(21,10744200,10760575)

and the corresponding view in IGV:

large gaps

it's sorted by insert size, so you're just seeing the big gaps. But there does seem to be some evidence for a large, consistent deletion.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/334)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -54,8 +54,8 @@ object StructuralVariant {
       r.mateAlignmentProperties.flatMap(_.inferredInsertSize).map(x => x * sgn)
     }
 
-    // Coalesce sequences of values separated by blockSize into single (inclusive) ranges.
-    // The sequence of values must be sorted and consist of multiples of blockSize.
+    // Given a sorted sequence of positions (all of which are multiples of @blockSize), coalesce ranges of
+    // consecutive multiples of @blockSize into tuples representing the start and end (inclusive) of each range.
     // For example, (0, 10, 20, 30, 70, 80) --> [(0, 30), (70, 80)]
     def coalesceAdjacent(xs: IndexedSeq[Long], blockSize: Long): Iterator[(Long, Long)] = {
       val rangeStarts = ArrayBuffer(0)
@@ -78,7 +78,7 @@ object StructuralVariant {
       if (!r.isMateMapped) {
         Seq[Long]()
       } else {
-        val mateStart = r.mateAlignmentProperties.map(mp => mp.start).getOrElse(r.read.start)
+        val mateStart = r.mateAlignmentProperties.map(_.start).getOrElse(r.read.start)
         val start = Math.min(r.read.start, mateStart)
         val roundedStart = start / BLOCK_SIZE * BLOCK_SIZE
         val insertSize = orientedInsertSize(r).getOrElse(0)
@@ -117,14 +117,13 @@ object StructuralVariant {
         .map(_._1)
 
       // Group adjacent blocks on the same contig
-      val exceptionalRanges = exceptionalBlocks
-        .groupBy(_.contig)
-        .flatMap {
-          case (contig, locations) => {
-            val positions = locations.map(_.position).toArray.sorted
-            coalesceAdjacent(positions, BLOCK_SIZE)
-              .map(x => GenomeRange(contig, x._1, x._2))
-          }
+      val exceptionalRanges =
+        for {
+          (contig, locations) <- exceptionalBlocks.groupBy(_.contig)
+          positions = locations.map(_.position).toArray.sorted
+          (start, end) <- coalesceAdjacent(positions, BLOCK_SIZE)
+        } yield {
+          GenomeRange(contig, start, end)
         }
 
       println("Number of exceptional blocks: " + exceptionalBlocks.count())
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -14,6 +14,10 @@ import scala.collection.mutable.ArrayBuffer
  * This currently looks for regions with abnormally large insert sizes. This typically indicates
  * that there's been a large deletion.
  *
+ * It does this by finding the average insert size of all mate pairs which overlap each 25-base pair "block" of the
+ * genome. If this average is significantly different than the average insert size genome-wide, then it's evidence that
+ * there may be a structural variant.
+ * 
  * The output is a text file of genomic ranges where there's some evidence of an SV.
  */
 object StructuralVariant {
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -3,7 +3,7 @@ package org.hammerlab.guacamole.commands
 import org.apache.spark.SparkContext
 import org.apache.spark.util.StatCounter
 import org.hammerlab.guacamole.{ DistributedUtil, SparkCommand, Common }
-import org.hammerlab.guacamole.reads.{ Read, MappedRead }
+import org.hammerlab.guacamole.reads.{ PairedRead, Read, MappedRead }
 import org.kohsuke.args4j.{ Option => Args4JOption }
 
 import scala.collection.mutable.ArrayBuffer
@@ -30,31 +30,24 @@ object StructuralVariant {
   }
 
   case class Location(
-    contig: String,
-    position: Long) {
+      contig: String,
+      position: Long) {
   }
 
   case class GenomeRange(
     contig: String,
     start: Long,
-    stop: Long
-  )
+    stop: Long)
 
   object Caller extends SparkCommand[Arguments] {
     override val name = "structural-variant"
     override val description = "Find structural variants, i.e. large insertions or deletions, inversions and translocations."
 
-    def isFirstInPair(read: MappedRead) = {
-      read.matePropertiesOpt match {
-        case Some(mateProps) => mateProps.isFirstInPair
-        case None            => false
-      }
-    }
-
     // The sign of inferredInsertSize follows the orientation of the read. This returns
     // an insert size which is positive in the common case, regardless of the orientation.
-    def orientedInsertSize(r: MappedRead): Option[Int] = {
-      r.matePropertiesOpt.flatMap(_.inferredInsertSize.map(x => x * (if (r.isPositiveStrand) { 1 } else { -1 })))
+    def orientedInsertSize(r: PairedRead[MappedRead]): Option[Int] = {
+      val sgn = if (r.isPositiveStrand) { 1 } else { -1 }
+      r.mateAlignmentProperties.flatMap(_.inferredInsertSize).map(x => x * sgn)
     }
 
     // Coalesce sequences of values separated by blockSize into single ranges.
@@ -76,19 +69,22 @@ object StructuralVariant {
     }
 
     // Returns a sequence of all blocks which overlap the read, its mate or the insert between them.
-    def matedReadBlocks(r: MappedRead): Seq[Long] = {
-      val mateStart = r.matePropertiesOpt.map(_.mateStart.getOrElse(r.start)).getOrElse(r.start)
-      val start = List(r.start, r.end, mateStart).min
-      val roundedStart = (1.0 * start / BLOCK_SIZE).toLong * BLOCK_SIZE
-      val insertSize = orientedInsertSize(r).getOrElse(0)
-
-      (roundedStart to (roundedStart + insertSize) by BLOCK_SIZE)
+    def matedReadBlocks(r: PairedRead[MappedRead]): Seq[Long] = {
+      if (!r.isMateMapped) {
+        Seq[Long]()
+      } else {
+        val mateStart = r.mateAlignmentProperties.map(mp => mp.start).getOrElse(r.read.start)
+        val start = List(r.read.start, r.read.end, mateStart).min
+        val roundedStart = (1.0 * start / BLOCK_SIZE).toLong * BLOCK_SIZE
+        val insertSize = orientedInsertSize(r).getOrElse(0)
+        (roundedStart to (roundedStart + insertSize) by BLOCK_SIZE)
+      }
     }
 
     override def run(args: Arguments, sc: SparkContext): Unit = {
       val readSet = Common.loadReadsFromArguments(args, sc, Read.InputFilters(nonDuplicate = true))
-      val pairedMappedReads = readSet.mappedReads.filter(_.isPaired)
-      val firstInPair = pairedMappedReads.filter(isFirstInPair)
+      val pairedMappedReads = readSet.mappedPairedReads
+      val firstInPair = pairedMappedReads.filter(_.isFirstInPair)
 
       // Pare down to mate pairs which aren't highly displaced & aren't inverted.
       val readsInRange = firstInPair.filter(r => {
@@ -105,7 +101,7 @@ object StructuralVariant {
       // TODO: take greater advantage of data locality than .groupByKey() does?
       val insertsByBlock = readsInRange.flatMap(r => {
         val insertSize = orientedInsertSize(r).getOrElse(0)
-        matedReadBlocks(r).map(pos => (Location(r.referenceContig, pos), insertSize))
+        matedReadBlocks(r).map(pos => (Location(r.read.referenceContig, pos), insertSize))
       }).groupByKey()
 
       // Find blocks where the average insert size is 2 sigma from the average for the whole genome.
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -1,0 +1,133 @@
+package org.hammerlab.guacamole.commands
+
+import org.apache.spark.SparkContext
+import org.apache.spark.util.StatCounter
+import org.hammerlab.guacamole.{ DistributedUtil, SparkCommand, Common }
+import org.hammerlab.guacamole.reads.{ Read, MappedRead }
+import org.kohsuke.args4j.{ Option => Args4JOption }
+
+import scala.collection.mutable.ArrayBuffer
+
+/**
+ * Structural Variant caller
+ *
+ * This currently looks for regions with abnormally large insert sizes. This typically indicates
+ * that there's been a large deletion.
+ *
+ * The output is a text file of genomic ranges where there's some evidence of an SV.
+ */
+object StructuralVariant {
+
+  val MAX_INSERT_SIZE = 25000
+  val BLOCK_SIZE = 25
+
+  protected class Arguments extends Common.Arguments.Reads with DistributedUtil.Arguments {
+    @Args4JOption(name = "--filter-contig", usage = "Filter to alignments where either mate is in this contig.")
+    var filterContig: String = ""
+
+    @Args4JOption(name = "--output", usage = "Path to output text file.")
+    var output: String = ""
+  }
+
+  case class Location(
+    contig: String,
+    position: Long) {
+  }
+
+  case class GenomeRange(
+    contig: String,
+    start: Long,
+    stop: Long
+  )
+
+  object Caller extends SparkCommand[Arguments] {
+    override val name = "structural-variant"
+    override val description = "Find structural variants, i.e. large insertions or deletions, inversions and translocations."
+
+    def isFirstInPair(read: MappedRead) = {
+      read.matePropertiesOpt match {
+        case Some(mateProps) => mateProps.isFirstInPair
+        case None            => false
+      }
+    }
+
+    // The sign of inferredInsertSize follows the orientation of the read. This returns
+    // an insert size which is positive in the common case, regardless of the orientation.
+    def orientedInsertSize(r: MappedRead): Option[Int] = {
+      r.matePropertiesOpt.flatMap(_.inferredInsertSize.map(x => x * (if (r.isPositiveStrand) { 1 } else { -1 })))
+    }
+
+    // Coalesce sequences of values separated by blockSize into single ranges.
+    // The sequence of values must be sorted.
+    def coalesceAdjacent(xs: IndexedSeq[Long], blockSize: Long): Iterator[(Long, Long)] = {
+      val rangeStarts = ArrayBuffer(0)
+      for (i <- 1 until xs.length) {
+        if (xs(i - 1) + blockSize != xs(i)) {
+          rangeStarts += i
+        }
+      }
+      rangeStarts += xs.length
+
+      rangeStarts.sliding(2).map(items => {
+        val startIdx = items(0)
+        val nextStartIdx = items(1)
+        (xs(startIdx), xs(nextStartIdx - 1))
+      })
+    }
+
+    // Returns a sequence of all blocks which overlap the read, its mate or the insert between them.
+    def matedReadBlocks(r: MappedRead): Seq[Long] = {
+      val mateStart = r.matePropertiesOpt.map(_.mateStart.getOrElse(r.start)).getOrElse(r.start)
+      val start = List(r.start, r.end, mateStart).min
+      val roundedStart = (1.0 * start / BLOCK_SIZE).toLong * BLOCK_SIZE
+      val insertSize = orientedInsertSize(r).getOrElse(0)
+
+      (roundedStart to (roundedStart + insertSize) by BLOCK_SIZE)
+    }
+
+    override def run(args: Arguments, sc: SparkContext): Unit = {
+      val readSet = Common.loadReadsFromArguments(args, sc, Read.InputFilters(nonDuplicate = true))
+      val pairedMappedReads = readSet.mappedReads.filter(_.isPaired)
+      val firstInPair = pairedMappedReads.filter(isFirstInPair)
+
+      // Pare down to mate pairs which aren't highly displaced & aren't inverted.
+      val readsInRange = firstInPair.filter(r => {
+        val insertSize = orientedInsertSize(r).getOrElse(0)
+        (insertSize > 0) && (insertSize < MAX_INSERT_SIZE)
+      })
+      readsInRange.persist()
+
+      val insertSizes = readsInRange.flatMap(orientedInsertSize)
+      val insertStats = insertSizes.stats()
+      println("Stats on inferredInsertSize: " + insertStats)
+
+      // Create a map from (block) -> [insert sizes]
+      // TODO: take greater advantage of data locality than .groupByKey() does?
+      val insertsByBlock = readsInRange.flatMap(r => {
+        val insertSize = orientedInsertSize(r).getOrElse(0)
+        matedReadBlocks(r).map(pos => (Location(r.referenceContig, pos), insertSize))
+      }).groupByKey()
+
+      // Find blocks where the average insert size is 2 sigma from the average for the whole genome.
+      val exceptionalThreshold = insertStats.mean + 2 * insertStats.stdev
+      val exceptionalBlocks = insertsByBlock.filter(pair => {
+        val inserts = pair._2
+        val stats = new StatCounter()
+        stats.merge(inserts.map(_.toDouble))
+        (stats.mean > exceptionalThreshold)
+      })
+
+      // Group adjacent blocks on the same contig
+      val exceptionalRanges = exceptionalBlocks.groupBy(_._1.contig).flatMap(kv => {
+        val contig = kv._1
+        val locations = kv._2.map(_._1.position).toArray.sorted
+        coalesceAdjacent(locations, BLOCK_SIZE).map(x => GenomeRange(contig, x._1, x._2))
+      })
+
+      println("Number of exceptional blocks: " + exceptionalBlocks.count())
+      exceptionalRanges.coalesce(1).saveAsTextFile(args.output)
+    }
+
+  }
+
+}
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -17,7 +17,7 @@ import scala.collection.mutable.ArrayBuffer
  * It does this by finding the average insert size of all mate pairs which overlap each 25-base pair "block" of the
  * genome. If this average is significantly different than the average insert size genome-wide, then it's evidence that
  * there may be a structural variant.
- * 
+ *
  * The output is a text file of genomic ranges where there's some evidence of an SV.
  */
 object StructuralVariant {
@@ -45,7 +45,7 @@ object StructuralVariant {
 
   object Caller extends SparkCommand[Arguments] {
     override val name = "structural-variant"
-    override val description = "Find structural variants, i.e. large insertions or deletions, inversions and translocations."
+    override val description = "Find structural variants, e.g. large deletions."
 
     // The sign of inferredInsertSize follows the orientation of the read. This returns
     // an insert size which is positive in the common case, regardless of the orientation.
@@ -54,8 +54,9 @@ object StructuralVariant {
       r.mateAlignmentProperties.flatMap(_.inferredInsertSize).map(x => x * sgn)
     }
 
-    // Coalesce sequences of values separated by blockSize into single ranges.
-    // The sequence of values must be sorted.
+    // Coalesce sequences of values separated by blockSize into single (inclusive) ranges.
+    // The sequence of values must be sorted and consist of multiples of blockSize.
+    // For example, (0, 10, 20, 30, 70, 80) --> [(0, 30), (70, 80)]
     def coalesceAdjacent(xs: IndexedSeq[Long], blockSize: Long): Iterator[(Long, Long)] = {
       val rangeStarts = ArrayBuffer(0)
       for (i <- 1 until xs.length) {
@@ -65,11 +66,11 @@ object StructuralVariant {
       }
       rangeStarts += xs.length
 
-      rangeStarts.sliding(2).map(items => {
-        val startIdx = items(0)
-        val nextStartIdx = items(1)
+      for {
+        Seq(startIdx, nextStartIdx) <- rangeStarts.sliding(2)
+      } yield {
         (xs(startIdx), xs(nextStartIdx - 1))
-      })
+      }
     }
 
     // Returns a sequence of all blocks which overlap the read, its mate or the insert between them.
@@ -78,10 +79,10 @@ object StructuralVariant {
         Seq[Long]()
       } else {
         val mateStart = r.mateAlignmentProperties.map(mp => mp.start).getOrElse(r.read.start)
-        val start = List(r.read.start, r.read.end, mateStart).min
-        val roundedStart = (1.0 * start / BLOCK_SIZE).toLong * BLOCK_SIZE
+        val start = Math.min(r.read.start, mateStart)
+        val roundedStart = start / BLOCK_SIZE * BLOCK_SIZE
         val insertSize = orientedInsertSize(r).getOrElse(0)
-        (roundedStart to (roundedStart + insertSize) by BLOCK_SIZE)
+        roundedStart to (roundedStart + insertSize) by BLOCK_SIZE
       }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+    start: Long,
+    stop: Long
+  )
+
+  object Caller extends SparkCommand[Arguments] {
+    override val name = "structural-variant"
+    override val description = "Find structural variants, i.e. large insertions or deletions, inversions and translocations."
+
+    def isFirstInPair(read: MappedRead) = {
+      read.matePropertiesOpt match {
```
danvk: Is this a reasonable way to get at `isFirstInPair`, or is there a simpler way to do it with `Option` types?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+      val readSet = Common.loadReadsFromArguments(args, sc, Read.InputFilters(nonDuplicate = true))
+      val pairedMappedReads = readSet.mappedReads.filter(_.isPaired)
+      val firstInPair = pairedMappedReads.filter(isFirstInPair)
+
+      // Pare down to mate pairs which aren't highly displaced & aren't inverted.
+      val readsInRange = firstInPair.filter(r => {
+        val insertSize = orientedInsertSize(r).getOrElse(0)
+        (insertSize > 0) && (insertSize < MAX_INSERT_SIZE)
+      })
+      readsInRange.persist()
```
danvk: Is there a rule of thumb about which RDDs you're supposed to `.persist()`?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+
+      val insertSizes = readsInRange.flatMap(orientedInsertSize)
+      val insertStats = insertSizes.stats()
+      println("Stats on inferredInsertSize: " + insertStats)
+
+      // Create a map from (block) -> [insert sizes]
+      // TODO: take greater advantage of data locality than .groupByKey() does?
```
danvk: I'm curious to hear your thoughts on this. The idea is that for an input which is sorted by position, nearby records will have similar blocks. So a combiner that operates on individual partitions would be very effective. I'm not sure if `.groupByKey` does that.

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
arahuja: I think using the collection functions is preferred over matching on `Some`/`None` 

```
read.matePropertiesOpt.exists(_.isPaired)
```

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
arahuja: If you are going to use the RDD twice, it might make sense to persist.  I rarely do as most are constructed once and then either output or shuffled, both operations which make no use of the persistance

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * This currently looks for regions with abnormally large insert sizes. This typically indicates
+ * that there's been a large deletion.
+ *
+ * The output is a text file of genomic ranges where there's some evidence of an SV.
+ */
+object StructuralVariant {
+
+  val MAX_INSERT_SIZE = 25000
+  val BLOCK_SIZE = 25
```
arahuja: Can you define/explain a 'block' somewhere. I read this initially thinking it had something to do with the BAM file.

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
danvk: I use `readsInRange` on both lines 96 (for getting stats on insert sizes) and on line 102 (for building the per-block list of insert sizes).

If I didn't have the `.persist()` here, would it have to re-run the `mappedPairedReads` and `.filter(_.isFirstInPair)` steps?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
arahuja: You may be looking for something like `combineByKey` then which you can two functions, a per partition combiner and then a per-key combiner. A `Combiner` and `Reducer` from the map-reduce world.

If possible, those are usually better to use anyways, since `groupByKey` pulls all the values of a key into memory.

You could also use `repartitionAndSortWithinPartitions` if you want your values in sorted order when processing.  

Haven't looked through what this is doing yet though to see if this applies.

danvk: I think I may want `aggregateByKey`, which lets you specify a reduce function, so that all the values don't have to go into memory at once.

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
danvk: Sure, I added a clarifying comment in the scaladoc for the caller.

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+    @Args4JOption(name = "--output", usage = "Path to output text file.")
+    var output: String = ""
+  }
+
+  case class Location(
+      contig: String,
+      position: Long) {
+  }
+
+  case class GenomeRange(
```
ryan-williams: Just noting that equivalents of these two case classes seem to proliferate around genomics codebases.

Using BDG [`Feature`](https://github.com/bigdatagenomics/bdg-formats/blob/de290eb5e5e237e13ee20ca9b162744888721041/src/main/resources/avro/bdg.avdl#L710)s / [`ReferenceRegion`](https://github.com/bigdatagenomics/adam/blob/a68a7e1ad3d970111d007b6a4011365e15e213a6/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala#L125)s would give you writing-to-Parquet for free, instead of a text file, fwiw.

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+    contig: String,
+    start: Long,
+    stop: Long)
+
+  object Caller extends SparkCommand[Arguments] {
+    override val name = "structural-variant"
+    override val description = "Find structural variants, i.e. large insertions or deletions, inversions and translocations."
```
ryan-williams: it doesn't do all of this yet, does it?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+
+    // The sign of inferredInsertSize follows the orientation of the read. This returns
+    // an insert size which is positive in the common case, regardless of the orientation.
+    def orientedInsertSize(r: PairedRead[MappedRead]): Option[Int] = {
+      val sgn = if (r.isPositiveStrand) { 1 } else { -1 }
+      r.mateAlignmentProperties.flatMap(_.inferredInsertSize).map(x => x * sgn)
+    }
+
+    // Coalesce sequences of values separated by blockSize into single ranges.
+    // The sequence of values must be sorted.
```
ryan-williams: A weird apparent pre-condition of this is also that no values in `xs` are separated by less than `blockSize`?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+      })
+    }
+
+    // Returns a sequence of all blocks which overlap the read, its mate or the insert between them.
+    def matedReadBlocks(r: PairedRead[MappedRead]): Seq[Long] = {
+      if (!r.isMateMapped) {
+        Seq[Long]()
+      } else {
+        val mateStart = r.mateAlignmentProperties.map(mp => mp.start).getOrElse(r.read.start)
+        val start = List(r.read.start, r.read.end, mateStart).min
```
ryan-williams: can `r.read.end` be less than `r.read.start`?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
timodonnell: A bit more detail on what the input and especially output are of this function would be helpful.

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+        val roundedStart = (1.0 * start / BLOCK_SIZE).toLong * BLOCK_SIZE
```
ryan-williams: why not just `start / BLOCK_SIZE * BLOCK_SIZE`? integer division ftw…

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+        val insertSize = orientedInsertSize(r).getOrElse(0)
+        (roundedStart to (roundedStart + insertSize) by BLOCK_SIZE)
```
ryan-williams: intellij gives me an "unnecessary parends" warning here and on L117, if we care about those

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+      }
+    }
+
+    override def run(args: Arguments, sc: SparkContext): Unit = {
+      val readSet = Common.loadReadsFromArguments(args, sc, Read.InputFilters(nonDuplicate = true))
+      val pairedMappedReads = readSet.mappedPairedReads
+      val firstInPair = pairedMappedReads.filter(_.isFirstInPair)
+
+      // Pare down to mate pairs which aren't highly displaced & aren't inverted.
+      val readsInRange = firstInPair.filter(r => {
```
ryan-williams: perhaps:
```
firstInPair.filter(r => orientedInsertSize(r).exists(_ < MAX_INSERT_SIZE))
```
or
```
firstInPair.filter(r => orientedInsertSize(r).exists(s => 0 < s && s < MAX_INSERT_SIZE))
```
depending whether you actually care about the `== 0` case or were just using `0` as a sentinel…

ryan-williams: more golfed:

```
firstInPair.flatMap(orientedInserSize).filter(_ < MAX_INSERT_SIZE)
```

ryan-williams: and while we're here, I increasingly find `for`-comps the most readable
```
for {
  r <- firstInPair
  s <- orientedInsertSize(r)
  if 0 < s && s < MAX_INSERT_SIZE
} yield s
```

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+        val insertSize = orientedInsertSize(r).getOrElse(0)
+        (insertSize > 0) && (insertSize < MAX_INSERT_SIZE)
+      })
+      readsInRange.persist()
+
+      val insertSizes = readsInRange.flatMap(orientedInsertSize)
```
ryan-williams: this {c,sh}ould be declared above L94; it would make that block cleaner (my 3rd comment there used this)

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+      val insertStats = insertSizes.stats()
+      println("Stats on inferredInsertSize: " + insertStats)
+
+      // Create a map from (block) -> [insert sizes]
+      // TODO: take greater advantage of data locality than .groupByKey() does?
+      val insertsByBlock = readsInRange.flatMap(r => {
+        val insertSize = orientedInsertSize(r).getOrElse(0)
```
ryan-williams: again, maybe want to just start out with `insertSizes` here?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
ryan-williams: `insertSizes.filter(_ < MAX_INSERT_SIZE)`

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
ryan-williams: ah nvm, my mistake.

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
ryan-williams:
```
(for {
  r <- readsInRange
  pos <- matedReadBlocks(r)
  insertSize <- orientedInsertSize(r)
} yield {
  Location(r.read.referenceContig, pos) -> insertSize
}).groupByKey
```

Tricky thing about `for`s and chained `flatMap`s and `map`s in Scala: you can't `flatMap` over an `Option` then a list; e.g. this would give a not-very-descriptive error about inferred type something somethings:
```
  r <- readsInRange
  insertSize <- orientedInsertSize(r)
  pos <- matedReadBlocks(r)
```

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+      val rangeStarts = ArrayBuffer(0)
+      for (i <- 1 until xs.length) {
+        if (xs(i - 1) + blockSize != xs(i)) {
+          rangeStarts += i
+        }
+      }
+      rangeStarts += xs.length
+
+      rangeStarts.sliding(2).map(items => {
+        val startIdx = items(0)
```
ryan-williams: intellij suggests `items.head` fwiw

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+    def coalesceAdjacent(xs: IndexedSeq[Long], blockSize: Long): Iterator[(Long, Long)] = {
+      val rangeStarts = ArrayBuffer(0)
+      for (i <- 1 until xs.length) {
+        if (xs(i - 1) + blockSize != xs(i)) {
+          rangeStarts += i
+        }
+      }
+      rangeStarts += xs.length
+
+      rangeStarts.sliding(2).map(items => {
```
ryan-williams: ICYMI:
```
for { 
  (startIdx, nextStartIdx) <- rangeStarts.sliding(2) 
} yield { 
  (xs(startIdx), xs(nextStartIdx - 1)) 
}
```

ryan-williams: good work using `sliding` though, btw :) I was looking for a way to suggest you use it on L62

ryan-williams: another syntax that's useful sometimes:
```
rangeStarts.sliding(2).map {
  case (startIdx, nextStartIdx) => (xs(startIdx), xs(nextStartIdx - 1))
}
```

ryan-williams: OK, in today's edition of "you can always do it in one pass using `foldLeft`", I think this works:

```
    def coalesceAdjacent(xs: Array[Long], blockSize: Long): Iterable[(Long, Long)] = {
      xs match {
        case Array(first, rest @ _*) =>
          val (list, last) =
            rest.foldLeft(
              (ListBuffer[(Long, Long)](), (first, first))
            )(
                (p, pos) => {
                  val (arr, (curStart, curEnd)) = p
                  if (curEnd + blockSize == pos)
                    (arr, (curStart, pos))
                  else
                    (arr :+(curStart, curEnd), (pos, pos))
                }
              )
          list :+ last
        case _ => Nil
      }
    }
```

Not claiming this is cleaner or better necessarily, but worth a look :)

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+    // The sign of inferredInsertSize follows the orientation of the read. This returns
+    // an insert size which is positive in the common case, regardless of the orientation.
+    def orientedInsertSize(r: PairedRead[MappedRead]): Option[Int] = {
+      val sgn = if (r.isPositiveStrand) { 1 } else { -1 }
+      r.mateAlignmentProperties.flatMap(_.inferredInsertSize).map(x => x * sgn)
+    }
+
+    // Coalesce sequences of values separated by blockSize into single ranges.
+    // The sequence of values must be sorted.
+    def coalesceAdjacent(xs: IndexedSeq[Long], blockSize: Long): Iterator[(Long, Long)] = {
```
ryan-williams: `IndexedSeq[Long]` seems like an unusual type to see here; any reason for it? Seems like you only pass an `Array[Long]` here

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
ryan-williams: Also probably worth commenting that the ranges returned are inclusive pairs of "block start" loci, if I have that right?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
ryan-williams: Yea, I think either one will work; if I have it right, `combineByKey` is more general, but for this case all you need is `aggregateByKey`.

You are going from an `RDD[(K, V)]` to an `RDD[(K, U)]`. You have:
* a `zeroValue: U`,
* a `seqOp: (U, V) => U`, and
* a `combOp: (U, U) => U`,

which are the necessities for the `aggregateByKey` API.

`combineByKey` generalizes further by substituting a `createCombiner: V => U` instead of `zeroValue: U`.

`K == Location`, `V == Long`, `U == StatsCounter`, `zeroValue == new StatCounter()`, and `seqOp` and `combOp` are overloaded versions of `StatsCounter.merge` that take a `Double` and a `StatsCounter`, resp.

ryan-williams:
```
      val exceptionalBlocks: RDD[Location] =
        (for {
          r <- readsInRange
          insertSize <- orientedInsertSize(r)
          pos <- matedReadBlocks(r)
        } yield {
          Location(r.read.referenceContig, pos) -> insertSize
        }).aggregateByKey(
          new StatCounter()
        )(
          seqOp = _ merge _,
          combOp = _ merge _
        )
        .filter(_._2.mean > exceptionalThreshold)
        .map(_._1)
```

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -0,0 +1,133 @@
...
+      val exceptionalThreshold = insertStats.mean + 2 * insertStats.stdev
+      val exceptionalBlocks = insertsByBlock.filter(pair => {
+        val inserts = pair._2
+        val stats = new StatCounter()
+        stats.merge(inserts.map(_.toDouble))
+        (stats.mean > exceptionalThreshold)
+      })
+
+      // Group adjacent blocks on the same contig
+      val exceptionalRanges = exceptionalBlocks.groupBy(_._1.contig).flatMap(kv => {
```
ryan-williams: I am guessing this will work for you with the size the array you anticipate dealing with, but the `groupBy`, plus the `.toArray.sorted` below, feels painful.

I *think* we can use some `{aggregate,combine}ByKey` trickery like I gave an example of above, as well as some thinking along the lines of my `foldLeft` example even further above, and do all of this in `RDD`-space without having to pull the whole values list into memory.

I'll see if I can get an example together for you.

ryan-williams: hm, yea not coming up with much... to do anything smarter you'd have to go into full "Region Join" territory, I think. Anyways you have at most a few million `Long`s in an array here, so it should be OK.

src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala
```diff
@@ -0,0 +1,16 @@
...
+class StructuralVariantCallerSuite extends GuacFunSuite with Matchers {
+  test("coalesce intervals") {
+    val nums = Array(0L, 10L, 20L, 30L,
+      50L,
+      70L, 80L,
+      100L, 110L, 120L,
+      150L)
+    val ranges = StructuralVariant.Caller.coalesceAdjacent(nums, 10)
+    assert(ranges.toList === List((0L, 30L), (50L, 50L), (70L, 80L), (100L, 120L), (150L, 150L)))
+  }
```
ryan-williams: a few more test cases would be nice. One I did in my shell was:

```
scala> coalesceAdjacent(Array(25,50,75,101,126,200,250,275), 0)
res2: Iterable[(Long, Long)] = ListBuffer((25,75), (101,126), (200,200), (250,275))
```

ryan-williams: also some tests of the calling would be good

danvk:
Thanks for all the feedback Ryan, Arun & Tim. I believe I've addressed all the comments, please take another look!

---

Review status: 0 of 3 files reviewed at latest revision, 16 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 41 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-41)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L41)):</sup>
But mine are subtly different :)

Point taken. Both those types look like they're for regions, though, rather than individual loci?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 48 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-48)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L48)):</sup>
Done.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 49 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r1-48)** ([raw file](https://github.com/hammerlab/guacamole/blob/5949189588c7e80069019ac15f1a25567a1998ea/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L49)):</sup>
(I deleted this code after rebasing on Arun's PairedRead refactor)

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 58 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-58)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L58)):</sup>
I expanded the comment. Let me know if you guys think there's a clearer way of structuring this; code whose explanation is as long as the code is usually no good.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 59 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-59)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L59)):</sup>
`IndexedSeq[Long]` is the more general type, right?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 68 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-68)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L68)):</sup>
I went with the for comprehension version. Had to make it `Seq(startIdx, nextStartIdx)` though.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 69 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-69)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L69)):</sup>
(this is obsolete now that I'm using the for comprehension)

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 81 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-81)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L81)):</sup>
Signs point to no. I switched this to use `Math.min` on the two possibilities.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 82 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-82)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L82)):</sup>
Done.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 84 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-84)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L84)):</sup>
Done.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 94 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-94)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L94)):</sup>
I do need to filter for the case where `orientedInsertSize < 0`. It corresponds to an inversion. 

I like the for comprehension version.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 105 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r1-105)** ([raw file](https://github.com/hammerlab/guacamole/blob/5949189588c7e80069019ac15f1a25567a1998ea/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L105)):</sup>
I used the for comprehension mostly as you suggested. Thanks for the code!

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 121 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-121)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L121)):</sup>
I rewrote this use `case` parameter matching. Could it be done more cleanly with another for comprehension?

---

<sup>**[src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala, line 15 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7_--r3-15)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala#L15)):</sup>
Hmm, I wouldn't consider `blockSize` of zero and values which aren't multiples of `blockSize` to be valid inputs.

I'd like to add some tests. Do you have advice about how to craft BAM files from scratch? Just type out lines of a SAM file and convert?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/334)

danvk:
Friendly ping

ryan-williams:
Looking now, sorry for the delay

ryan-williams:
Review status: 0 of 3 files reviewed at latest revision, 11 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 41 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-41)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L41)):</sup>
The exact equivalents would be ADAM's [ReferencePosition](https://github.com/bigdatagenomics/adam/blob/adam-parent_2.10-0.17.0/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferencePosition.scala#L86) and [ReferenceRegion](https://github.com/bigdatagenomics/adam/blob/adam-parent_2.10-0.17.0/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala#L125).


Any comment on the point about writing Features to Parquet instead of bespoke text file? "Sounds like a maybe a good idea later" is a fine response, j/w.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 57 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jv-dGykX0gmviyOmDKj)** ([raw file](https://github.com/hammerlab/guacamole/blob/d07e54906ac51da590e4eea75b3d5cfc31d46b2d/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L57)):</sup>
Still a little confusing imho, but not a big deal. One attempt at rephrasing fwiw:

Given a sorted sequence of positions (all of which are multiples of @blockSize), coalesce ranges of consecutive multiples of @blockSize into tuples representing the start and end (inclusive) of each range.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 81 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jv-f6xo_GfRiUQzwyI0)** ([raw file](https://github.com/hammerlab/guacamole/blob/d07e54906ac51da590e4eea75b3d5cfc31d46b2d/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L81)):</sup>
`s/mp => mp/_/`

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 115 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jv-gIYL-7eNLUWcMPh9)** ([raw file](https://github.com/hammerlab/guacamole/blob/d07e54906ac51da590e4eea75b3d5cfc31d46b2d/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L115)):</sup>
heh, didn't want to go for `(seqOp = _ merge _, combOp = _ merge _)`? this lgtm too, j/w

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 120 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-121)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L120)):</sup>
Sure. My earlier comment was about trying to avoid the `.toArray` below, but in response to your q:
```
val exceptionalRanges =
  for {
    (contig, locations) <- exceptionalBlocks.groupBy(_.contig)
    positions = locations.map(_.position).toArray.sorted
    (start, end) <- coalesceAdjacent(positions, BLOCK_SIZE)
  } yield {
    GenomeRange(contig, start, end)
  }
```

As a rule, any time you have a `map` inside one or more `flatMap`s, you can make a `for` out of it; you don't have to, jfyi.

Also this rule is good enough that I remember it and not even why it's true without thinking about it for a minute :)

---

<sup>**[src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala, line 15 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7_--r3-15)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala#L15)):</sup>
Oh, sorry, I didn't mean that `0` for `blockSize` above; I had hard-coded `25` in the impl of `coalesceAdjacent` in my repl, so that `0` was just a placeholder. With that example I just meant to show you a more complicated test of the functionality. I see now that it breaks the precondition of everything being a multiple of `blockSize`; feel free to disregard.

re: calling tests... yea we don't have a great story there. Every time I'm in this position I put N units of work into improving that story and generate issues related to it worth 2N units of work.

Open to using this to try to improve the story there, but also open to punting to a later PR.

One answer is make a SAM file with many copies of one read, then manually change each one's start/end and mate info to create the test you want, and remove as many extraneous fields as you can.

Another (probably better) answer is to just construct your reads (with mate info) manually in scala in your test case.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/334)

danvk:
Addressed all comments, please take another look.

---

Review status: 0 of 3 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 41 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-41)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L41)):</sup>
Eventually it should output [`StructuralVariant`][1] records, but this is all pretty in flux right now.

[1]: https://github.com/bigdatagenomics/bdg-formats/blob/de290eb5e5e237e13ee20ca9b162744888721041/src/main/resources/avro/bdg.avdl#L373

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 57 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jv-dGykX0gmviyOmDKj)** ([raw file](https://github.com/hammerlab/guacamole/blob/d07e54906ac51da590e4eea75b3d5cfc31d46b2d/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L57)):</sup>
Done.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 58 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-58)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L58)):</sup>
Done.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 81 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jv-f6xo_GfRiUQzwyI0)** ([raw file](https://github.com/hammerlab/guacamole/blob/d07e54906ac51da590e4eea75b3d5cfc31d46b2d/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L81)):</sup>
Done.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 115 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jv-gIYL-7eNLUWcMPh9)** ([raw file](https://github.com/hammerlab/guacamole/blob/d07e54906ac51da590e4eea75b3d5cfc31d46b2d/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L115)):</sup>
Yeah, paren-less function calling scares me a little bit.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 120 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-121)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L120)):</sup>
Cool. The destructuring assignment seems like a big win of using `for` comprehensions.

---

<sup>**[src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala, line 15 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7_--r3-15)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala#L15)):</sup>
Pure end-to-end tests of MapReduces tend to be really brittle. Any little change you make will change the output in complex ways.

The best way to test pipelines like this is to break them down into isolated bits that have more well-defined behavior, e.g. "load the mated, mapped reads with insert sizes in a given range", "find the median insert size for a collection of reads" and "group these locations into contiguous ranges" and test each of those independently.

I'd like to do that once this code settles down a bit (it's going to change dramatically as I implement something more DELLY-like).

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/334)

arahuja:
Review status: 0 of 3 files reviewed at latest revision, 11 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 120 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7Zz-r3-121)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L120)):</sup>
Sorry for the late add here, and this can probably be handled later, but the positions can be sorted without loading them into an array first, but it's a little annoying

i.e. 
```
class ContigPartitioner(tasks: Int) extends Partitioner {
   def partition(key: Any) {
        case key: Location(contig, start, end) => contig //or something to map it to an int
    }
}


exceptionalBlocks.keyBy(somethingThatIsLocation).repartitionAndSort(new ContigPartitioner())

exceptionalBlocks.mapPartitions( things => {
  while (things)
       while (notCoalesced) {
          keepCoalescing
       }
       yield coalesced
   }
 } 
```
etc ... written in Reviewable/not Intellij  and Location would need an ordering

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/334)

ryan-williams:
Review status: 0 of 3 files reviewed at latest revision, 11 unresolved discussions, all commit checks successful.

---

<sup>**[src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala, line 15 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/334#-Jur141mDoaPaQaHo7_--r3-15)** ([raw file](https://github.com/hammerlab/guacamole/blob/86aec297b422922dae2c9c224b1e25ad7304d4f2/src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala#L15)):</sup>
I'm saying that you can (and we do typically when we test stuff like this) create a dummy `RDD[GenomeRange]` or `RDD[MappedRead]` in scala pretty easily.

```
sc.parallelize(List(MappedRead(…), MappedRead(…), MappedRead(…)))
```

Then you can write some very simple tests of just the logic you're adding here, without worrying about I/O.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/334)

PR #336, opened 2015-07-23
 Add referenceName to Pileup and fix warning

### LINKED ISSUE #335: make Pileup.referenceName work even when the pileup is empty
We currently load the reference name (i.e. the contig) from the first read in the pileup: https://github.com/hammerlab/guacamole/blob/master/src/main/scala/org/hammerlab/guacamole/pileup/Pileup.scala#L43

This makes an assertion error if the pileup is empty. We should carry around the reference name in the pileup so it works in this case.

As a separate issue, `referenceName` is kind of an awkward name for this. Maybe we should rename it to `contig` or `chromosome`?

timodonnell:
* Add referenceName to Pileup as an attribute. This makes it possible to access this field even when the Pileup is empty. Closes #335.
* Fix a compile warning  in `ReadSet`.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/336)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/ReadSet.scala
+++ b/src/main/scala/org/hammerlab/guacamole/ReadSet.scala
@@ -53,8 +53,8 @@ case class ReadSet(
 
   lazy val mappedPairedReads: RDD[PairedRead[MappedRead]] = reads.flatMap(read =>
     read match {
-      case rp if rp.isMapped => Some(rp.asInstanceOf[PairedRead[MappedRead]])
-      case _                 => None
+      case rp: PairedRead[_] if rp.isMapped => Some(rp.asInstanceOf[PairedRead[MappedRead]])
+      case _                                => None
     }
   )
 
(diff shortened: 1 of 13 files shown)


src/main/scala/org/hammerlab/guacamole/ReadSet.scala
```diff
@@ -53,8 +53,8 @@ case class ReadSet(
 
   lazy val mappedPairedReads: RDD[PairedRead[MappedRead]] = reads.flatMap(read =>
     read match {
-      case rp: PairedRead[MappedRead] => Some(rp)
-      case _                          => None
+      case rp if rp.isMapped => Some(rp.asInstanceOf[PairedRead[MappedRead]])
```
arahuja: This doesn't check that `rp` is a `ReadPair` though, it may just be a `MappedRead`

timodonnell: Good catch, fixed.

PR #337, opened 2015-07-24
Construct a DELLY-like Structural Variant graph.

danvk:
This finds paired reads with exceptionally large insert sizes. It then builds a graph where compatible pairs are connected.

Highlights:

- Introduces a `PairedMappedRead` case class for when both a Read and its mate are mapped.
- Implements the "simple line-sweep algorithm" described in the [DELLY paper][1].
- Breaks the pipeline into a few more easily testable pieces and adds unit tests.

[1]: http://bioinformatics.oxfordjournals.org/content/28/18/i333

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/337)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -2,12 +2,15 @@ package org.hammerlab.guacamole.commands
 
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.RDD
-import org.apache.spark.util.StatCounter
 import org.hammerlab.guacamole.{ DistributedUtil, SparkCommand, Common }
-import org.hammerlab.guacamole.reads.{ PairedMappedRead, PairedRead, Read, MappedRead }
+import org.hammerlab.guacamole.reads.{ PairedMappedRead, Read }
 import org.kohsuke.args4j.{ Option => Args4JOption }
 
-import scala.collection.mutable.ArrayBuffer
+import scalax.collection.Graph
+import scalax.collection.edge._
+import scalax.collection.mutable.{ Graph => MutableGraph }
+import scalax.collection.GraphPredef._
+import scalax.collection.edge.Implicits._
 
 /**
  * Structural Variant caller
@@ -47,12 +50,11 @@ object StructuralVariant {
     median: Double,
     mad: Double)
 
-  // An undirected graph of evidence for structural variants
-  // TODO: find an appropriate Scala graph library and use that.
-  case class PairedReadGraph(
-    reads: IndexedSeq[PairedMappedRead], // vertices
-    edges: Seq[(Int, Int, Double)] // (vertex1, vertex2, weight)
-    )
+  // An undirected, weighted graph of evidence for structural variants.
+  // Nodes are paired reads.
+  // Edges represent compatibility between the induced structural variants,
+  // with the weight being the strength of compatibility.
+  type PairedReadGraph = Graph[PairedMappedRead, WUnDiEdge]
 
   object Caller extends SparkCommand[Arguments] {
     override val name = "structural-variant"
@@ -154,7 +156,7 @@ object StructuralVariant {
     // For more details, see the DELLY paper.
     def buildVariantGraph(exceptionalReads: Iterable[PairedMappedRead], maxNormalInsertSize: Int): PairedReadGraph = {
       val reads = exceptionalReads.toArray.sortBy(r => Math.min(r.read.start, r.mate.start))
-      val edges = new ArrayBuffer[(Int, Int, Double)]()
+      val graph = MutableGraph[PairedMappedRead, WUnDiEdge]()
 
       for {
         i <- 0 until reads.length
@@ -171,7 +173,7 @@ object StructuralVariant {
             done = true
           } else {
             if (!areReadsIncompatible(read, nextRead, maxNormalInsertSize)) {
-              edges += ((i, j, 1.0))
+              graph += (read ~ nextRead) % 1
             }
           }
           j += 1
@@ -179,7 +181,7 @@ object StructuralVariant {
       }
 
       // TODO: exclude isolated reads from the graph
-      PairedReadGraph(reads, edges)
+      graph
     }
 
     override def run(args: Arguments, sc: SparkContext): Unit = {
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -1,9 +1,10 @@
 package org.hammerlab.guacamole.commands
 
 import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
 import org.apache.spark.util.StatCounter
 import org.hammerlab.guacamole.{ DistributedUtil, SparkCommand, Common }
-import org.hammerlab.guacamole.reads.{ PairedRead, Read, MappedRead }
+import org.hammerlab.guacamole.reads.{ PairedMappedRead, PairedRead, Read, MappedRead }
 import org.kohsuke.args4j.{ Option => Args4JOption }
 
 import scala.collection.mutable.ArrayBuffer
@@ -34,14 +35,24 @@ object StructuralVariant {
   }
 
   case class Location(
-      contig: String,
-      position: Long) {
-  }
+    contig: String,
+    position: Long)
 
   case class GenomeRange(
     contig: String,
     start: Long,
     stop: Long)
+
+  case class MedianStats(
+    median: Double,
+    mad: Double)
+
+  // An undirected graph of evidence for structural variants
+  // TODO: find an appropriate Scala graph library and use that.
+  case class PairedReadGraph(
+    reads: IndexedSeq[PairedMappedRead], // vertices
+    edges: Seq[(Int, Int, Double)] // (vertex1, vertex2, weight)
+    )
 
   object Caller extends SparkCommand[Arguments] {
     override val name = "structural-variant"
@@ -49,85 +60,141 @@ object StructuralVariant {
 
     // The sign of inferredInsertSize follows the orientation of the read. This returns
     // an insert size which is positive in the common case, regardless of the orientation.
-    def orientedInsertSize(r: PairedRead[MappedRead]): Option[Int] = {
-      val sgn = if (r.isPositiveStrand) { 1 } else { -1 }
-      r.mateAlignmentProperties.flatMap(_.inferredInsertSize).map(x => x * sgn)
+    def orientedInsertSize(r: PairedMappedRead): Int = {
+      val sgn = if (r.read.isPositiveStrand) { 1 } else { -1 }
+      r.inferredInsertSize * sgn
     }
 
-    // Given a sorted sequence of positions (all of which are multiples of @blockSize), coalesce ranges of
-    // consecutive multiples of @blockSize into tuples representing the start and end (inclusive) of each range.
-    // For example, (0, 10, 20, 30, 70, 80) --> [(0, 30), (70, 80)]
-    def coalesceAdjacent(xs: IndexedSeq[Long], blockSize: Long): Iterator[(Long, Long)] = {
-      val rangeStarts = ArrayBuffer(0)
-      for (i <- 1 until xs.length) {
-        if (xs(i - 1) + blockSize != xs(i)) {
-          rangeStarts += i
+    // Extract median and MAD (Median Absolute Deviation) from an unordered Array.
+    def medianStats[T <% Double](xs: Array[T])(implicit ord: Ordering[T]): MedianStats = {
+      if (xs.isEmpty) {
+        return MedianStats(0.0, 0.0)
+      }
+
+      // Extract the median of a sorted array of numbers
+      def getMedian[T <% Double](nums: Array[T]): Double = {
+        val len = nums.length
+        if (len % 2 == 0) {
+          0.5 * (nums(len / 2 - 1) + nums(len / 2))
+        } else {
+          1.0 * nums(len / 2)
         }
       }
-      rangeStarts += xs.length
 
-      for {
-        Seq(startIdx, nextStartIdx) <- rangeStarts.sliding(2)
-      } yield {
-        (xs(startIdx), xs(nextStartIdx - 1))
-      }
+      val nums = xs.sorted
+      val median = getMedian(nums)
+      val residuals = nums.map(x => Math.abs(1.0 * x - median)).sorted
+      val mad = getMedian(residuals)
+
+      MedianStats(median, mad)
     }
 
-    // Returns a sequence of all blocks which overlap the read, its mate or the insert between them.
-    def matedReadBlocks(r: PairedRead[MappedRead]): Seq[Long] = {
-      if (!r.isMateMapped) {
-        Seq[Long]()
-      } else {
-        val mateStart = r.mateAlignmentProperties.map(_.start).getOrElse(r.read.start)
-        val start = Math.min(r.read.start, mateStart)
-        val roundedStart = start / BLOCK_SIZE * BLOCK_SIZE
-        val insertSize = orientedInsertSize(r).getOrElse(0)
-        roundedStart to (roundedStart + insertSize) by BLOCK_SIZE
+    case class ExceptionalReadsReturnType(
+      val readsInRange: RDD[PairedMappedRead],
+      val insertSizes: RDD[Int],
+      val insertStats: MedianStats,
+      val maxNormalInsertSize: Int,
+      val exceptionalReads: RDD[PairedMappedRead])
+
+    def getExceptionalReads(reads: RDD[PairedMappedRead]): ExceptionalReadsReturnType = {
+      // Pare down to mate pairs which aren't highly displaced & aren't inverted or translocated.
+      val readsInRange = reads.filter(r => {
+        val s = r.inferredInsertSize
+        r.read.referenceContig == r.mate.referenceContig &&
+          r.read.isPositiveStrand != r.mate.isPositiveStrand &&
+          s < MAX_INSERT_SIZE
+      })
+      readsInRange.persist()
+
+      val insertSizes = readsInRange.map(orientedInsertSize)
+      val insertStats = medianStats(insertSizes.takeSample(false, 100000))
+      println("insert stats: " + insertStats)
+
+      // Find blocks where the average insert size differs significantly from the genome-wide median.
+      // Following DELLY, we use a threshold of 5 median absolute deviations above the median.
+      val maxNormalInsertSize = (insertStats.median + 5 * insertStats.mad).toInt
+
+      val exceptionalReads = readsInRange.filter(_.inferredInsertSize > maxNormalInsertSize)
+
+      ExceptionalReadsReturnType(
+        readsInRange,
+        insertSizes,
+        insertStats,
+        maxNormalInsertSize,
+        exceptionalReads
+      )
+    }
+
+    // Given two pairs of reads, is there a deletion which would make both of them have normal insert sizes?
+    // TODO: this assumes that both reads have read.start.pos < read.mate.pos
+    def areReadsIncompatible(read1: PairedMappedRead, read2: PairedMappedRead, maxNormalInsertSize: Long): Boolean = {
+      if (read1.read.start < read2.read.start) return areReadsIncompatible(read2, read1, maxNormalInsertSize)
+
+      // This matches the DELLY logic; the last five lines are copy/paste. See
+      // https://github.com/tobiasrausch/delly/blob/7464cacfe1d420ac5bd7ed40a3093a80a93a6964/src/tags.h#L441-L445
+      val pair1Min = Math.min(read1.read.start, read1.mate.start)
+      val pair1Max = pair1Min + read1.inferredInsertSize
+      val pair1ReadLength = read1.read.sequence.length
+
+      val pair2Min = Math.min(read2.read.start, read2.mate.start)
+      val pair2Max = pair2Min + read2.inferredInsertSize
+      val pair2ReadLength = read2.read.sequence.length
+
+      if ((pair2Min + pair2ReadLength - pair1Min) > maxNormalInsertSize) return true;
+      if ((pair2Max < pair1Max) && ((pair1Max + pair1ReadLength - pair2Max) > maxNormalInsertSize)) return true;
+      if ((pair2Max >= pair1Max) && ((pair2Max + pair2ReadLength - pair1Max) > maxNormalInsertSize)) return true;
+      if ((pair1Max < pair2Min) || (pair2Max < pair1Min)) return true;
+      return false;
+    }
+
+    // Given a list of reads with abnormally large inserts, construct a graph representation of the structural variants
+    // where:
+    // Vertices = paired reads
+    // Edges = two sets of paired reads could represent the same structural variant
+    // For more details, see the DELLY paper.
+    def buildVariantGraph(exceptionalReads: Iterable[PairedMappedRead], maxNormalInsertSize: Int): PairedReadGraph = {
+      val reads = exceptionalReads.toArray.sortBy(r => Math.min(r.read.start, r.mate.start))
+      val edges = new ArrayBuffer[(Int, Int, Double)]()
+
+      for {
+        i <- 0 until reads.length
+      } {
+        val read = reads(i)
+        val start = Math.min(read.read.start, read.mate.start)
+        var j = i + 1
+        var done = false
+        // Loop through subsequent reads until it's impossible to find another compatible one.
+        while (j < reads.length && !done) {
+          val nextRead = reads(j)
+          val nextStart = Math.min(nextRead.read.start, nextRead.mate.start)
+          if (Math.abs(nextStart + nextRead.read.sequence.length - start) > maxNormalInsertSize) {
+            done = true
+          } else {
+            if (!areReadsIncompatible(read, nextRead, maxNormalInsertSize)) {
+              edges += ((i, j, 1.0))
+            }
+          }
+          j += 1
+        }
       }
+
+      // TODO: exclude isolated reads from the graph
+      PairedReadGraph(reads, edges)
     }
 
     override def run(args: Arguments, sc: SparkContext): Unit = {
       val readSet = Common.loadReadsFromArguments(args, sc, Read.InputFilters(nonDuplicate = true))
       val pairedMappedReads = readSet.mappedPairedReads
-      val firstInPair = pairedMappedReads.filter(_.isFirstInPair)
+      val firstInPair = pairedMappedReads.filter(_.isFirstInPair).flatMap(PairedMappedRead(_))
 
-      // Pare down to mate pairs which aren't highly displaced & aren't inverted.
-      val readsInRange = for {
-        r <- firstInPair
-        s <- orientedInsertSize(r)
-        if 0 < s && s < MAX_INSERT_SIZE
-      } yield r
-      readsInRange.persist()
+      val packed = getExceptionalReads(firstInPair)
+      val exceptionalReads = packed.exceptionalReads
+      val maxNormalInsertSize = packed.maxNormalInsertSize
 
-      val insertSizes = readsInRange.flatMap(orientedInsertSize)
-      val insertStats = insertSizes.stats()
-      println("Stats on inferredInsertSize: " + insertStats)
+      val readsByContig = exceptionalReads.groupBy(_.read.referenceContig)
+      val svGraph = readsByContig.mapValues(buildVariantGraph(_, maxNormalInsertSize))
 
-      // Find blocks where the average insert size is 2 sigma from the average for the whole genome.
-      val exceptionalThreshold = insertStats.mean + 2 * insertStats.stdev
-      val exceptionalBlocks = (for {
-        r <- readsInRange
-        pos <- matedReadBlocks(r)
-        insertSize <- orientedInsertSize(r)
-      } yield {
-        Location(r.read.referenceContig, pos) -> insertSize
-      })
-        .aggregateByKey(new StatCounter())(seqOp = _.merge(_), combOp = _.merge(_))
-        .filter { case (loc, stats) => stats.mean > exceptionalThreshold }
-        .map(_._1)
-
-      // Group adjacent blocks on the same contig
-      val exceptionalRanges =
-        for {
-          (contig, locations) <- exceptionalBlocks.groupBy(_.contig)
-          positions = locations.map(_.position).toArray.sorted
-          (start, end) <- coalesceAdjacent(positions, BLOCK_SIZE)
-        } yield {
-          GenomeRange(contig, start, end)
-        }
-
-      println("Number of exceptional blocks: " + exceptionalBlocks.count())
-      exceptionalRanges.coalesce(1).saveAsTextFile(args.output)
+      svGraph.coalesce(1).saveAsTextFile(args.output)
     }
   }
 }
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -93,11 +93,11 @@ object StructuralVariant {
     }
 
     case class ExceptionalReadsReturnType(
-      val readsInRange: RDD[PairedMappedRead],
-      val insertSizes: RDD[Int],
-      val insertStats: MedianStats,
-      val maxNormalInsertSize: Int,
-      val exceptionalReads: RDD[PairedMappedRead])
+      readsInRange: RDD[PairedMappedRead],
+      insertSizes: RDD[Int],
+      insertStats: MedianStats,
+      maxNormalInsertSize: Int,
+      exceptionalReads: RDD[PairedMappedRead])
 
     def getExceptionalReads(reads: RDD[PairedMappedRead]): ExceptionalReadsReturnType = {
       // Pare down to mate pairs which aren't highly displaced & aren't inverted or translocated.
@@ -129,9 +129,9 @@ object StructuralVariant {
     }
 
     // Given two pairs of reads, is there a deletion which would make both of them have normal insert sizes?
-    def areReadsIncompatible(read1: PairedMappedRead, read2: PairedMappedRead, maxNormalInsertSize: Long): Boolean = {
+    def areReadsCompatible(read1: PairedMappedRead, read2: PairedMappedRead, maxNormalInsertSize: Long): Boolean = {
       if (read1.minPos > read2.minPos) {
-        areReadsIncompatible(read2, read1, maxNormalInsertSize)
+        areReadsCompatible(read2, read1, maxNormalInsertSize)
       } else {
         // This matches the DELLY logic; the last five lines are copy/paste. See
         // https://github.com/tobiasrausch/delly/blob/7464cacfe1d420ac5bd7ed40a3093a80a93a6964/src/tags.h#L441-L445
@@ -141,10 +141,10 @@ object StructuralVariant {
         val (pair2Min, pair2Max) = read2.ends
         val pair2ReadLength = read2.read.sequence.length
 
-        (((pair2Min + pair2ReadLength - pair1Min) > maxNormalInsertSize) ||
-         ((pair2Max < pair1Max) && ((pair1Max + pair1ReadLength - pair2Max) > maxNormalInsertSize)) ||
-         ((pair2Max >= pair1Max) && ((pair2Max + pair2ReadLength - pair1Max) > maxNormalInsertSize)) ||
-         ((pair1Max < pair2Min) || (pair2Max < pair1Min)))
+        !(((pair2Min + pair2ReadLength - pair1Min) > maxNormalInsertSize) ||
+          ((pair2Max < pair1Max) && ((pair1Max + pair1ReadLength - pair2Max) > maxNormalInsertSize)) ||
+          ((pair2Max >= pair1Max) && ((pair2Max + pair2ReadLength - pair1Max) > maxNormalInsertSize)) ||
+          ((pair1Max < pair2Min) || (pair2Max < pair1Min)))
       }
     }
 
@@ -168,7 +168,7 @@ object StructuralVariant {
           if (Math.abs(nextStart + nextRead.readLength - start) > maxNormalInsertSize) {
             done = true
           } else {
-            if (!areReadsIncompatible(read, nextRead, maxNormalInsertSize)) {
+            if (areReadsCompatible(read, nextRead, maxNormalInsertSize)) {
               graph += (read ~ nextRead) % 1
             }
           }
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -69,11 +69,8 @@ object StructuralVariant {
 
     // Extract median and MAD (Median Absolute Deviation) from an unordered Array.
     def medianStats[T <% Double](xs: Array[T])(implicit ord: Ordering[T]): MedianStats = {
-      if (xs.isEmpty) {
-        return MedianStats(0.0, 0.0)
-      }
 
-      // Extract the median of a sorted array of numbers
+      // Extract the median of a sorted, non-empty array of numbers
       def getMedian[T <% Double](nums: Array[T]): Double = {
         val len = nums.length
         if (len % 2 == 0) {
@@ -83,12 +80,16 @@ object StructuralVariant {
         }
       }
 
-      val nums = xs.sorted
-      val median = getMedian(nums)
-      val residuals = nums.map(x => Math.abs(1.0 * x - median)).sorted
-      val mad = getMedian(residuals)
+      if (xs.isEmpty) {
+        MedianStats(0.0, 0.0)
+      } else {
+        val nums = xs.sorted
+        val median = getMedian(nums)
+        val residuals = nums.map(x => Math.abs(1.0 * x - median)).sorted
+        val mad = getMedian(residuals)
 
-      MedianStats(median, mad)
+        MedianStats(median, mad)
+      }
     }
 
     case class ExceptionalReadsReturnType(
@@ -128,25 +129,23 @@ object StructuralVariant {
     }
 
     // Given two pairs of reads, is there a deletion which would make both of them have normal insert sizes?
-    // TODO: this assumes that both reads have read.start.pos < read.mate.pos
     def areReadsIncompatible(read1: PairedMappedRead, read2: PairedMappedRead, maxNormalInsertSize: Long): Boolean = {
-      if (read1.read.start < read2.read.start) return areReadsIncompatible(read2, read1, maxNormalInsertSize)
+      if (read1.minPos > read2.minPos) {
+        areReadsIncompatible(read2, read1, maxNormalInsertSize)
+      } else {
+        // This matches the DELLY logic; the last five lines are copy/paste. See
+        // https://github.com/tobiasrausch/delly/blob/7464cacfe1d420ac5bd7ed40a3093a80a93a6964/src/tags.h#L441-L445
+        val (pair1Min, pair1Max) = read1.ends
+        val pair1ReadLength = read1.readLength
 
-      // This matches the DELLY logic; the last five lines are copy/paste. See
-      // https://github.com/tobiasrausch/delly/blob/7464cacfe1d420ac5bd7ed40a3093a80a93a6964/src/tags.h#L441-L445
-      val pair1Min = Math.min(read1.read.start, read1.mate.start)
-      val pair1Max = pair1Min + read1.inferredInsertSize
-      val pair1ReadLength = read1.read.sequence.length
+        val (pair2Min, pair2Max) = read2.ends
+        val pair2ReadLength = read2.read.sequence.length
 
-      val pair2Min = Math.min(read2.read.start, read2.mate.start)
-      val pair2Max = pair2Min + read2.inferredInsertSize
-      val pair2ReadLength = read2.read.sequence.length
-
-      if ((pair2Min + pair2ReadLength - pair1Min) > maxNormalInsertSize) return true;
-      if ((pair2Max < pair1Max) && ((pair1Max + pair1ReadLength - pair2Max) > maxNormalInsertSize)) return true;
-      if ((pair2Max >= pair1Max) && ((pair2Max + pair2ReadLength - pair1Max) > maxNormalInsertSize)) return true;
-      if ((pair1Max < pair2Min) || (pair2Max < pair1Min)) return true;
-      return false;
+        (((pair2Min + pair2ReadLength - pair1Min) > maxNormalInsertSize) ||
+         ((pair2Max < pair1Max) && ((pair1Max + pair1ReadLength - pair2Max) > maxNormalInsertSize)) ||
+         ((pair2Max >= pair1Max) && ((pair2Max + pair2ReadLength - pair1Max) > maxNormalInsertSize)) ||
+         ((pair1Max < pair2Min) || (pair2Max < pair1Min)))
+      }
     }
 
     // Given a list of reads with abnormally large inserts, construct a graph representation of the structural variants
@@ -155,21 +154,18 @@ object StructuralVariant {
     // Edges = two sets of paired reads could represent the same structural variant
     // For more details, see the DELLY paper.
     def buildVariantGraph(exceptionalReads: Iterable[PairedMappedRead], maxNormalInsertSize: Int): PairedReadGraph = {
-      val reads = exceptionalReads.toArray.sortBy(r => Math.min(r.read.start, r.mate.start))
+      val reads = exceptionalReads.toArray.sortBy(_.minPos)
       val graph = MutableGraph[PairedMappedRead, WUnDiEdge]()
 
-      for {
-        i <- 0 until reads.length
-      } {
-        val read = reads(i)
-        val start = Math.min(read.read.start, read.mate.start)
+      for { (read, i) <- reads.zipWithIndex } {
+        val start = read.minPos
         var j = i + 1
         var done = false
         // Loop through subsequent reads until it's impossible to find another compatible one.
         while (j < reads.length && !done) {
           val nextRead = reads(j)
-          val nextStart = Math.min(nextRead.read.start, nextRead.mate.start)
-          if (Math.abs(nextStart + nextRead.read.sequence.length - start) > maxNormalInsertSize) {
+          val nextStart = nextRead.minPos
+          if (Math.abs(nextStart + nextRead.readLength - start) > maxNormalInsertSize) {
             done = true
           } else {
             if (!areReadsIncompatible(read, nextRead, maxNormalInsertSize)) {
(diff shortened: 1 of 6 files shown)


src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +35,167 @@ object StructuralVariant {
...
-    def matedReadBlocks(r: PairedRead[MappedRead]): Seq[Long] = {
-      if (!r.isMateMapped) {
-        Seq[Long]()
-      } else {
-        val mateStart = r.mateAlignmentProperties.map(_.start).getOrElse(r.read.start)
-        val start = Math.min(r.read.start, mateStart)
-        val roundedStart = start / BLOCK_SIZE * BLOCK_SIZE
-        val insertSize = orientedInsertSize(r).getOrElse(0)
-        roundedStart to (roundedStart + insertSize) by BLOCK_SIZE
+    case class ExceptionalReadsReturnType(
```
danvk: This feels awkward. Suggestions on better ways to structure this would be appreciated!

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +35,167 @@ object StructuralVariant {
...
     contig: String,
     start: Long,
     stop: Long)
 
+  case class MedianStats(
+    median: Double,
+    mad: Double)
+
+  // An undirected graph of evidence for structural variants
+  // TODO: find an appropriate Scala graph library and use that.
```
danvk: Suggestions here would be appreciated.

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +38,165 @@ object StructuralVariant {
...
-    def coalesceAdjacent(xs: IndexedSeq[Long], blockSize: Long): Iterator[(Long, Long)] = {
-      val rangeStarts = ArrayBuffer(0)
-      for (i <- 1 until xs.length) {
-        if (xs(i - 1) + blockSize != xs(i)) {
-          rangeStarts += i
-        }
+    // Extract median and MAD (Median Absolute Deviation) from an unordered Array.
+    def medianStats[T <% Double](xs: Array[T])(implicit ord: Ordering[T]): MedianStats = {
+      if (xs.isEmpty) {
+        return MedianStats(0.0, 0.0)
```
ryan-williams: Not sure of others' thoughts on this but I think `return`s like this are somewhat un-idiomatic in Scala; I remember hearing in particular that `return` is implemented by `throw`ing, which is perf-costly.

```
if (xs.isEmpty) {
  MedianStats(0.0, 0.0)
} else {
  …
}
```

or

```
xs match {
  case Array(head, rest @ _*) => 
    …
  case _ => 
    MedianStats(0.0, 0.0)
}
```

maybe better?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +38,165 @@ object StructuralVariant {
...
+      val insertSizes: RDD[Int],
+      val insertStats: MedianStats,
+      val maxNormalInsertSize: Int,
+      val exceptionalReads: RDD[PairedMappedRead])
+
+    def getExceptionalReads(reads: RDD[PairedMappedRead]): ExceptionalReadsReturnType = {
+      // Pare down to mate pairs which aren't highly displaced & aren't inverted or translocated.
+      val readsInRange = reads.filter(r => {
+        val s = r.inferredInsertSize
+        r.read.referenceContig == r.mate.referenceContig &&
```
ryan-williams: the comment on `PairedMappedRead` says they're always on the same contig...

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +38,165 @@ object StructuralVariant {
...
+          r.read.isPositiveStrand != r.mate.isPositiveStrand &&
+          s < MAX_INSERT_SIZE
+      })
+      readsInRange.persist()
+
+      val insertSizes = readsInRange.map(orientedInsertSize)
+      val insertStats = medianStats(insertSizes.takeSample(false, 100000))
```
ryan-williams: thoughts on getting the actual median / MAD? cf. [SO for RDD's median](http://stackoverflow.com/a/28160731/544236)

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +38,165 @@ object StructuralVariant {
...
+        insertStats,
+        maxNormalInsertSize,
+        exceptionalReads
+      )
+    }
+
+    // Given two pairs of reads, is there a deletion which would make both of them have normal insert sizes?
+    // TODO: this assumes that both reads have read.start.pos < read.mate.pos
+    def areReadsIncompatible(read1: PairedMappedRead, read2: PairedMappedRead, maxNormalInsertSize: Long): Boolean = {
+      if (read1.read.start < read2.read.start) return areReadsIncompatible(read2, read1, maxNormalInsertSize)
```
ryan-williams: cf. earlier comment about `return`

ryan-williams: also doesn't this fix the `TODO` on L131?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +38,165 @@ object StructuralVariant {
...
+
+      // This matches the DELLY logic; the last five lines are copy/paste. See
+      // https://github.com/tobiasrausch/delly/blob/7464cacfe1d420ac5bd7ed40a3093a80a93a6964/src/tags.h#L441-L445
+      val pair1Min = Math.min(read1.read.start, read1.mate.start)
+      val pair1Max = pair1Min + read1.inferredInsertSize
```
ryan-williams: have we settled on "insert size" including the reads' lengths? that is the more confusing interpretation to me, I'd love for us to come with a different name that means that, that we can port in and out of at our interfaces with whichever tool is using it in the confusing way

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +38,165 @@ object StructuralVariant {
...
+      val pair1ReadLength = read1.read.sequence.length
+
+      val pair2Min = Math.min(read2.read.start, read2.mate.start)
+      val pair2Max = pair2Min + read2.inferredInsertSize
+      val pair2ReadLength = read2.read.sequence.length
+
+      if ((pair2Min + pair2ReadLength - pair1Min) > maxNormalInsertSize) return true;
```
ryan-williams: can you comment these lines explaining what they do? the lack of obvious a pair1⟺pair2 symmetry is disconcerting.

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +38,165 @@ object StructuralVariant {
...
-      // Group adjacent blocks on the same contig
-      val exceptionalRanges =
-        for {
-          (contig, locations) <- exceptionalBlocks.groupBy(_.contig)
-          positions = locations.map(_.position).toArray.sorted
-          (start, end) <- coalesceAdjacent(positions, BLOCK_SIZE)
-        } yield {
-          GenomeRange(contig, start, end)
-        }
+      val packed = getExceptionalReads(firstInPair)
```
ryan-williams: possibly:
```
val ExceptionalReadsReturnType(_, _, _, maxNormalInsertSize, exceptionalReads) = getExceptionalReads(firstInPair)
```

rather than these three lines

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +38,165 @@ object StructuralVariant {
...
+        i <- 0 until reads.length
+      } {
+        val read = reads(i)
+        val start = Math.min(read.read.start, read.mate.start)
+        var j = i + 1
+        var done = false
+        // Loop through subsequent reads until it's impossible to find another compatible one.
+        while (j < reads.length && !done) {
+          val nextRead = reads(j)
+          val nextStart = Math.min(nextRead.read.start, nextRead.mate.start)
```
ryan-williams: could be worth making this a `def`, `val`, or `lazy val` on `PairedMappedRead`

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +38,165 @@ object StructuralVariant {
...
+          if (Math.abs(nextStart + nextRead.read.sequence.length - start) > maxNormalInsertSize) {
```
ryan-williams: what does `nextStart + nextRead.read.sequence.length` mean if `nextStart` is actually the mate's start pos?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -34,100 +38,165 @@ object StructuralVariant {
...
+    // Given a list of reads with abnormally large inserts, construct a graph representation of the structural variants
+    // where:
+    // Vertices = paired reads
+    // Edges = two sets of paired reads could represent the same structural variant
+    // For more details, see the DELLY paper.
+    def buildVariantGraph(exceptionalReads: Iterable[PairedMappedRead], maxNormalInsertSize: Int): PairedReadGraph = {
+      val reads = exceptionalReads.toArray.sortBy(r => Math.min(r.read.start, r.mate.start))
+      val graph = MutableGraph[PairedMappedRead, WUnDiEdge]()
+
+      for {
```
ryan-williams: fwiw:
```
reads.zipWithIndex.map {
  case (read, idx) =>
    reads
      .drop(idx + 1)
      .takeWhile(next => next.minStart + next.read.sequence.length - read.minStart <= maxNormalInsertSize)
      .filter(next => !areReadsIncompatible(read, next, maxNormalInsertSize))
}
```

may be less efficient by a constant, e.g. if the `.drop` means you always traverse the whole second array

src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala
```diff
@@ -1,16 +1,75 @@
...
+    assert(medianStats(nums2) == MedianStats(1.5, 0.5))
+
+    // Pathological cases
+    assert(medianStats(Array(1)) == MedianStats(1.0, 0.0))
+    assert(medianStats(Array()) == MedianStats(0.0, 0.0))
+  }
+
+  sparkTest("read filtering") {
+    val reads = Seq(
+      // A few reads with an insert size of 100 to set the median and MAD
```
ryan-williams: 1. a little confusing that this says 100 but they're all 88 below.
2. does insert size mean "length of first read + the size between the reads"?! that's a 3rd definition that i did not expect.

ryan-williams: does this set the median to 100 and the MAD to 0?

ryan-williams:
cool, back to you @danvk. I would like us to clear up what we're calling insert sizes, and what the core logic is around read-pairs' pairwise-"compatibility"

danvk:
Back to you, @ryan-williams.

Here's a diagram showing what "insert size" means (in this case `read.start` < `mate.start`):
img_0994

I think some of this logic would be clearer if we had names for the four points of interest in a paired read.

---

Review status: 0 of 6 files reviewed at latest revision, 12 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 73 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-73)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L73)):</sup>
I restructured to use an `if` statement. Do you have a citation for `return` being implemented using exceptions? That's somewhat surprising to me, but it would make the "`return` breaks out of anonymous functions" behavior less surprising.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 105 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-105)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L105)):</sup>
Good point. I removed that requirement from `PairedMappedRead`; paired reads which map to different contigs are indicative of translocations.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 112 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-112)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L112)):</sup>
Sorting the entire RDD to get the precise median seems like overkill, as does implementing the linear-time median selection algorithm. It would be quite surprising if the median of a 100,000 element sample of insert sizes differed that much from the true median.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 133 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-133)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L133)):</sup>
I reworked it to not use `return`. One thing I really like about early-out `return` statements is that they let you add quick checks to the top of a function without forcing you to indent everything below them.

It doesn't actually fix the TODO: this `if` checks that the first paired read is before the second paired read, whereas the TODO is about the position of the main read/mate read within each pair. I reworked it to fix the TODO.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 138 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-138)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L138)):</sup>
[This discussion][1] indicates that "insert size" is the distance between the sequencing primers, i.e. the distance between the first & last mapped base pair. I agree that this is confusing, but it does seem to be a standard term. The distance between the (inner) ends of the two paired reads is the "gap distance".

I'm pretty open to moving some of this logic into `PairedMappedRead` or adding a step that forces the pairs into a particular orientation. Also open to discussions about how to name the start/end of each read.

[1]: http://seqanswers.com/forums/showthread.php?t=13098

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 145 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-145)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L145)):</sup>
The `read1.read.start < read2.read.start` requirement at the top of the function is what breaks the symmetry.

The high-level explanation is in the comment for this function: "Given two pairs of reads, is there a deletion which would make both of them have normal insert sizes?" The detailed explanation is in a few pages of math I wrote out last week. Here's my diagram of the scenarios that each of the four lines of this `if` statement correspond to:

img_0995

What were you imagining? Some ASCII art?

I think that this could be written much more directly by finding the largest deletion that matches both reads and determining whether that's in range. It's just that then I'd lose parity with DELLY.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 161 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-161)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L161)):</sup>
I switched to use `reads.zipWithIndex`. It looks like `Array.drop` makes a copy of the remainder of the `Array`, so it would be quite inefficient. Does Scala have an equivalent of Python's slices which returns an iterator without copying? (`.slice` also makes a copy)

Also, wouldn't use `map` here produce a giant `Array` of `Unit`?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 171 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-171)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L171)):</sup>
I added a few `def`s to clean this up. Is memory blowup a concern with `val` or `lazy val`?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 172 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-172)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L172)):</sup>
The sequence lengths for each end should be the same. If they're not, there's no simple way to get at the end position since it isn't denormalized into the SAM record. 

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 192 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-192)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L192)):</sup>
Done. Is there any way to make this an anonymous type? Ideally I wouldn't have to define it explicitly, Scala could just figure it out from what I return in my function.

---

<sup>**[src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala, line 32 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppX-r2-32)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala#L32)):</sup>
Insert size is length of first read + gap between reads + length of second read. Since `TestUtil` has a default sequence length of 12, this works out to `97 - 9 + 12 = 100`, etc. I added a comment.

These five reads collectively set the median and the MAD since they have the same insert size and outnumber all the other reads. I updated the comment to spell it out more.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/337)

ryan-williams:
Cool, definitely looking better and I think there's some important points for us to hash out. More comments below.

---

Review status: 0 of 6 files reviewed at latest revision, 14 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 71 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-73)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L71)):</sup>
[This](http://stackoverflow.com/a/12335710/544236) is the best thing I could find in a few mins of googling. lmk if you are not persuaded. Feels like a gotcha that we might as well avoid, but I don't feel strongly.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 97 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvGlUYjBmGU1Ty5QPlo)** ([raw file](https://github.com/hammerlab/guacamole/blob/cb595ed45d718d67e393876f0cb03340217e09ac/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L97)):</sup>
IJ says http://cl.ly/image/3n0F1N1t401i

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 113 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-112)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L113)):</sup>
Sure, but you are going to run a Spark job to do this `taksSample` as it is, and several others in the course of running this caller, so I'm not convinced that there's much downside.. a maybe-only-marginally longer Spark job vs. possibly-premature optimization, I guess.

Are you sure that you won't get a biased sample? Is the MAD also resilient enough to downsampling, possibly by several OoMs?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 133 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-133)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L133)):</sup>
Understood re: indenting, but there's an argument that the indenting is good: there are two branches and they are equally indented. It boils down to a functional vs. imperative kind of subjectivity, I guess.

Anyway, given that, per your comment below, this line breaks symmetry in what still feels to me like an unnecessarily confusing set of `if`s later, I'm less and less a fan of doing it this way.

If this 5-`if`s is what took you hours to grok from DELLY, then we should do better if at all possible. See my query below about why we can't just intersect the gaps and subtract some lengths.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 138 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-138)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L138)):</sup>
OK, insert vs. gap makes sense. The rest we can address on other comments. Feel free to ack this.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 139 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvGlh09Q2RX5RmuhDRm)** ([raw file](https://github.com/hammerlab/guacamole/blob/cb595ed45d718d67e393876f0cb03340217e09ac/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L139)):</sup>
IJ says http://cl.ly/image/2O3H0Q3o0U0Y

you should probably just remove the parends on the method declarations to fix this.

or we can declare that we don't care and I can turn off those warnings

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 144 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-145)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L144)):</sup>
In what way would you lose parity? If DELLY is doing what it means to do correctly, and we do it correctly in a way that we find easier to grok, everything's still OK right?

My instinct is that this can be formulated more simply: you have 2 intervals (the 2 gaps); intersect them, subtract the intersection's length from each gap's length, and compare the difference to the cutoff. Is that not what you're doing?

I think that a simple  sketch, possibly with a very small amount of ASCII-art, (re)iterating the meanings of insert size, gap size, and stating what assumptions you're making and not making about each read being before its mate or the other pair, would be useful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 160 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-161)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L160)):</sup>
Ah you're right re: the return being wrong, though I think it would return an `Array[Array[PairedMappedRead]]` as written; I don't think `Array[Unit]` is correct?

Needs a `flatMap`:
```
reads.zipWithIndex.flatMap {
  case (read, idx) =>
    reads
      .drop(idx + 1)
      .takeWhile(next => next.minStart + next.read.sequence.length - read.minStart <= maxNormalInsertSize)
      .filter(next => !areReadsIncompatible(read, next, maxNormalInsertSize))
      .map(next => (read ~ next) % 1)
}
```

Obligatory "there's a `map` inside a `.flatMap`, plus a destructuring assignment, let's `for` it":

```
for {
  (read, idx) <- reads.zipWithIndex
  next <- reads.drop(idx + 1).takeWhile(next => next.minStart + next.read.sequence.length - read.minStart <= maxNormalInsertSize)
  if !areReadsIncompatible(read, next, maxNormalInsertSize)
} yield {
  (read ~ next) % 1
}
```

Good digging re: the `Array.drop` and `.slice`, though I think as I said, it's `O(N^2)` either way, right, so a constant worse? Anyway, I'm fine with the loops, just wanted to discuss.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 167 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-171)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L167)):</sup>
It can be. Could be worth running it both ways and looking at the memory usage stats in Spree or Spark UI :) This is a kind of question we bump up against a lot and we need to come up with easy and rigorous ways of answering them.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 168 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-172)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L168)):</sup>
eesh. Should we be worried about soft- and/or hard-clipping then? A mate with a clipped start could mess with our calculations here. Probably an edge case, but... at the very least, maybe worth commenting since the answer is not actually that the code is immune to the possibility.

You'd have to have a `MappedReadPair` data structure that had both reads on it :)

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 171 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvGiiYIW459TmeKtE-C)** ([raw file](https://github.com/hammerlab/guacamole/blob/cb595ed45d718d67e393876f0cb03340217e09ac/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L171)):</sup>
I'm not a huge fan of the double-negative semantics on this

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 188 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-192)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L188)):</sup>
You mean, like, `Tuple5`? :) You could do that. FWIW it sounds like I find this less troublesome than you do.

Maybe just renaming it: `ReadInsertStats`, `InsertSizedReads`, etc. `ReturnType` is the awkward thing imho.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/reads/PairedMappedRead.scala, line 46 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvGmFNx-Sql6d78iSz4)** ([raw file](https://github.com/hammerlab/guacamole/blob/cb595ed45d718d67e393876f0cb03340217e09ac/src/main/scala/org/hammerlab/guacamole/reads/PairedMappedRead.scala#L46)):</sup>
I have IJ set to always add a newline on save

---

<sup>**[src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala, line 32 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppX-r2-32)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala#L32)):</sup>
Cool, thanks. Thoughts on making the MAD more like 1? Making it 0 feels like a degenerate case. Not a strong ask though.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/337)

danvk:
Back to you. The main remaining point is whether to improve on DELLY from the get-go vs. match it first & then improve on it.

---

Review status: 0 of 6 files reviewed at latest revision, 12 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 71 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-73)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L71)):</sup>
From reading that SO it sounds like `return` is implemented with exceptions only if it's used inside a closure. So performance doesn't seem like a concern. If it's considered good Scala style to avoid it, though, I'm happy to not rock the boat as a new Scala dev.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 97 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvGlUYjBmGU1Ty5QPlo)** ([raw file](https://github.com/hammerlab/guacamole/blob/cb595ed45d718d67e393876f0cb03340217e09ac/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L97)):</sup>
Done.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 113 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-112)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L113)):</sup>
In what way would it be biased? I'd expect median/MAD to be less affected by downsampling than mean/stdev, since they're robust to outliers.

I guess the optimization is doing one pass over the data (to get the sample) vs at least four (collect the insert sizes, sort them, extract the median, calculate the deviations, sort those, extract the median). It may be even worse if `.lookup` is implemented as another full linear scan over the `RDD`.

LMK what you think, I can run some benchmarks & numbers on synth4 tomorrow.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 133 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-133)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L133)):</sup>
I agree 100% that we should improve this logic. And if it were correct, I'd be happy to do it right now. But I don't think it is. It's imprecise in very particular ways which your suggestion would correct.

I'd prefer to try and match DELLY precisely and _then_ improve it, rather than imitating it and making changes to its algorithm in one go.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 139 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvGlh09Q2RX5RmuhDRm)** ([raw file](https://github.com/hammerlab/guacamole/blob/cb595ed45d718d67e393876f0cb03340217e09ac/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L139)):</sup>
I removed the parens on the declarations. Didn't know that was something you could do!

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 144 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-145)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L144)):</sup>
LMK what you think re: my point above about matching DELLY & then improving on it.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 160 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-161)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L160)):</sup>
We expect each node to be connected to relatively few other nodes relative to the size of the whole graph. In other words, the set of reads remaining after your `.drop(...).takeWhile(...)` is expected to be O(1). Then the loop version I wrote is expected to be O(N), since it does O(1) checks for each of the N elements. In the worst case it could be O(N^2) if the entire graph is fully-connected.

`.drop` makes a copy of the entire remaining portion of the array, which is O(N). So the version with the for comprehension is guaranteed to be O(N^2). If `.drop` were implemented as a O(1) view on the Array, then the two versions would be equivalent.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 168 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-172)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L168)):</sup>
I wonder if I should throw out clipped reads for this analysis. They'll be useful for the next step, which is to look for short reads which cross the deletion. If a read is soft-clipped, it might be because it runs up against the deleted portion of the genome, which gives you a good indication of precisely where the deletion starts/stops.

I believe the only way to get CIGAR strings & sequences for both reads in a pair would be to parse the read name & join on it. @arahuja wasn't very enthusiastic about that approach, since parsing read names would blow up the memory requirements.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 171 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvGiiYIW459TmeKtE-C)** ([raw file](https://github.com/hammerlab/guacamole/blob/cb595ed45d718d67e393876f0cb03340217e09ac/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L171)):</sup>
Switched to `areReadsCompatible`

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 188 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-192)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L188)):</sup>
Pairs were considered a code smell at Google, so Tuple5 definitely doesn't appeal to me. `_1`, `_2` and their kin seem pretty common in Scala though.

I like the anonymous type because it would prevent it from being used as anything but the return type of this function. I think of `ReturnType` as second best because you'd think twice about ever typing that in a declaration other than the return type of the function.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/reads/PairedMappedRead.scala, line 46 \[r3\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvGmFNx-Sql6d78iSz4)** ([raw file](https://github.com/hammerlab/guacamole/blob/cb595ed45d718d67e393876f0cb03340217e09ac/src/main/scala/org/hammerlab/guacamole/reads/PairedMappedRead.scala#L46)):</sup>
Done.

---

<sup>**[src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala, line 32 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppX-r2-32)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala#L32)):</sup>
Sure, done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/337)

ryan-williams:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 133 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-133)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L133)):</sup>
OK, matching DELLY then iterating is fine, but now I'm confused about "our" (really: your) belief about the correctness of the code. Do we think it's incorrect, know it's incorrect, or think it's correct, or know it's correct? If it's incorrect, do we have examples? 

In either case, this feels like something we should have tests and/or docs for. This is really dense logic that it seems like we're agreeing is not grok-able in a reasonable amount of time. If it's wrong on corner cases, leaving that here knowingly without docs or tests seems unfair to future-us.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 142 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvI6VFia5RP68WeGzc7)** ([raw file](https://github.com/hammerlab/guacamole/blob/6fe131c35d98b0379723f4d70aca1b177eb98286/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L142)):</sup>
nit: `.readLength` above but not here; any reason?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 145 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvI78xOPJ2gedfUp8J3)** ([raw file](https://github.com/hammerlab/guacamole/blob/6fe131c35d98b0379723f4d70aca1b177eb98286/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L145)):</sup>
one example of a thing I'm uneasy with about this: I did not expect to see `pair1Max` and `pair1ReadLength` here with the same sign (likewise with `pair2` below). What does one read length beyond the end of the rightmost read even mean? Do we know? Can we document it?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 188 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-192)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L188)):</sup>
tuples with destructuring assignment seem much less bad to me than tuples where `_N`'s are thrown around.

I don't think I know what the anonymous types you speak of would look like here..?

Anyways this lgtm

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/337)

danvk:
Back to you, Ryan. I've added a bunch of tests for `areReadsCompatible` and marked the ones which I believe DELLY gets incorrect.

---

Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 133 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-133)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L133)):</sup>
I believe that if DELLY says two reads are incompatible, then they are incompatible. If it says that they're compatible (or, really "not incompatible") then they may or may not actually be compatible.

I've added some tests cases which illustrate this and clearly marked the ones which I believe should pass but don't. The point about future us is well-taken.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 142 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvI6VFia5RP68WeGzc7)** ([raw file](https://github.com/hammerlab/guacamole/blob/6fe131c35d98b0379723f4d70aca1b177eb98286/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L142)):</sup>
Done.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 145 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvI78xOPJ2gedfUp8J3)** ([raw file](https://github.com/hammerlab/guacamole/blob/6fe131c35d98b0379723f4d70aca1b177eb98286/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L145)):</sup>
Argh, this is something that I inadvertently changed in translation. `pair1Max` should be the start of the latter read, so that `pair1Max + pair1ReadLength` is the end of the read, i.e. what I currently have defined as `pair1Max`. See what I mean about how having names for these points would be helpful?

I fixed the bug by creating a new method, `startsAndStops`, which returns all four numbers.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 188 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-192)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L188)):</sup>
I was thinking something like the first example [here][1]. I had trouble getting that to work because of this error:

```
scala> val foo = "hello"
foo: String = hello
scala> new { val foo = foo }
<console>:20: error: recursive value foo needs type
              new { val foo = foo }
                              ^
scala> new { val foo: String = foo }
<console>:20: warning: value foo does nothing other than call itself recursively
              new { val foo: String = foo }
```

[1]: http://daily-scala.blogspot.com/2009/12/labelled-multiple-returns.html

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/337)

ryan-williams:
Thanks for bearing with me, i feel much better about this. OTOH i feel less good about using reviewable on mobile!

Just a couple of other ideas below

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 145 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvI78xOPJ2gedfUp8J3)** ([raw file](https://github.com/hammerlab/guacamole/blob/6fe131c35d98b0379723f4d70aca1b177eb98286/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L145)):</sup>
Cool, i like making it clear which of the 4 points you're using.


Suggested naming convention for the four points here / elsewhere: `pairMin`, `gapMin`, `gapMax`, `pairMax`. s/Min/Start/ and s/Max/End/g` also wfm. It's a nit but calling [the end of the gap] the [max of the pair] is still confusing me.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 188 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-192)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L188)):</sup>
Send like that's because `foo` in the definition of the `val` just trivially recursively refers to the `val` being defined? Anyway, whatever you want to do here wfm

---

<sup>**[src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala, line 46 \[r5\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvLjBkSVJv58Ipm8JdF)** ([raw file](https://github.com/hammerlab/guacamole/blob/fd6f0532d8fd75851585fb7d690c7a76fc6a3ae0/src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala#L46)):</sup>
Thanks for doing this and making it clear what DELLY does and doesn't do.

One possible nit: uncomment the "wrong" ones, put DELLY's wrong answers as the "expected" values, and add a comment on such cases explaining that DELLY is wrong on those. That has the advantage of being even more convincing that DELLY gets those wrong bc the tests would be failing otherwise :)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/337)

danvk:
Back to you, Ryan. Almost there!

---

Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 145 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvI78xOPJ2gedfUp8J3)** ([raw file](https://github.com/hammerlab/guacamole/blob/6fe131c35d98b0379723f4d70aca1b177eb98286/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L145)):</sup>
Went with `pair1Min`, `pair1GapMin`, `pair1GapMax`, `pair1Max`.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 188 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-192)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L188)):</sup>
Yeah, that's definitely what's happening. How do I make it not happen (other than by renaming my variables)?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/337)

danvk:
Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

<sup>**[src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala, line 46 \[r5\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-JvLjBkSVJv58Ipm8JdF)** ([raw file](https://github.com/hammerlab/guacamole/blob/fd6f0532d8fd75851585fb7d690c7a76fc6a3ae0/src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala#L46)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/337)

ryan-williams:
lgtm thanks Dan

---

Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 188 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/337#-Jv0X46RwwPuFMBf-ppW-r2-192)** ([raw file](https://github.com/hammerlab/guacamole/blob/e5f2bde65dca9b6acc9ae4d81334ddc2c12bc2b4/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L188)):</sup>
Hm i think you rename your variables (just in the function when you're computing them though, right?)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/337)

PR #338, opened 2015-07-24
Skip variant calls in the middle of the a deletion pileup

### LINKED ISSUE #275: MidDeletion lead to genotypes with a blank allele
There are two problems with deletions that overlap with SNV, as in the image below

image

The pileup preceding the SNV will evaluate the deletion and possibly call a deletion, however this leads to two problems in the next position
1. the next position contains both MidDeletion and Mismatch elements.  MidDeletion elements do not define reference or alternate alleles so both are empty (but equal) but labeled as "reference"
2. whether or not we call deletion before should affect whether or not we call the SNV at the next position

arahuja:
Fixes #275

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/338)


arahuja:
@ryan-williams @timodonnell @danvk mind taking a look at this one?

timodonnell:
LGTM

PR #339, opened 2015-07-25
Remove qc-metrics as a snapshot dependency

arahuja:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/339)


ryan-williams:
great, thx

PR #340, opened 2015-07-27
Join output variants with DBSNP to retrieve rsID

### LINKED ISSUE #326: Add dbSNP input to variant callers
This input can be used to:
1. Label variants as they are called
2. Add option to filter out variants

arahuja:
Fixes #326 

This adds an optional rsID to `CalledSomaticGenotype`

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/340)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala
@@ -20,6 +20,8 @@ package org.hammerlab.guacamole.commands
 
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.RDD
+import org.bdgenomics.adam.rdd.ADAMContext
+import org.bdgenomics.formats.avro.DatabaseVariantAnnotation
 import org.hammerlab.guacamole.Common.Arguments.SomaticCallerArgs
 import org.hammerlab.guacamole.filters.PileupFilter.PileupFilterArguments
 import org.hammerlab.guacamole.filters.SomaticGenotypeFilter.SomaticGenotypeFilterArguments
@@ -28,7 +30,6 @@ import org.hammerlab.guacamole.likelihood.Likelihood
 import org.hammerlab.guacamole.pileup.Pileup
 import org.hammerlab.guacamole.reads.Read
 import org.hammerlab.guacamole.variants.{ AlleleConversions, AlleleEvidence, CalledSomaticAllele }
-import org.hammerlab.guacamole.windowing.SlidingWindow
 import org.hammerlab.guacamole.{ Common, DelayedMessages, DistributedUtil, SparkCommand }
 import org.kohsuke.args4j.{ Option => Args4jOption }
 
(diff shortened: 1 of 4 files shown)


timodonnell:
LGTM

src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala
```diff
@@ -113,6 +117,28 @@ object SomaticStandard {
           args.minTumorAlternateReadDepth
         )
 
+      if (args.dbSnpVcf != "") {
+        val adamContext = new ADAMContext(sc)
+        val dbSnpVariants = adamContext.loadVariantAnnotations(args.dbSnpVcf)
+        potentialGenotypes
```
ryan-williams: does this command have a side effect that is useful? I'm missing how it's having any effect otherwise

src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala
```diff
@@ -113,6 +117,28 @@ object SomaticStandard {
...
+          .keyBy(_.adamVariant)
+          .leftOuterJoin(dbSnpVariants.keyBy(_.variant))
+          .map(_._2).map({
+            case (calledAllele: CalledSomaticAllele, dbSnpVariant: Option[DatabaseVariantAnnotation]) =>
+              CalledSomaticAllele(
+                sampleName = calledAllele.sampleName,
```
ryan-williams: any reason not to use `calledAllele.copy(rsID = dbSnpVariant.map(_.getDbSnpId))` here?

src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala
```diff
@@ -113,6 +117,28 @@ object SomaticStandard {
           args.minTumorAlternateReadDepth
         )
 
+      if (args.dbSnpVcf != "") {
+        val adamContext = new ADAMContext(sc)
+        val dbSnpVariants = adamContext.loadVariantAnnotations(args.dbSnpVcf)
+        potentialGenotypes
+          .keyBy(_.adamVariant)
+          .leftOuterJoin(dbSnpVariants.keyBy(_.variant))
```
ryan-williams: `s/_.variant/_.getVariant/`? IJ telling me that the direct accessor is deprecated

ryan-williams:
Catching up on some PRs that I missed, a couple possibly-dumb questions here @arahuja if you have a minute

src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala
arahuja: You're right! This should have been `potentialGenotypes = ...` I couldn't get this into the output, since that needs to come in ADAM, but I should fix this.

src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala
arahuja: Yes, that's essentially what I wanted, didn't see that before.

src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala
arahuja: Yes, will update

PR #341, opened 2015-07-28
Add metric for avg number of mismatches per read

arahuja:
This adds some new fields to `AlleleEvidence` to debug some variants

I also removed the custom Kryo serializers as I didn't see any change in performance with running without them and maintaining that code was a pain.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/341)


ryan-williams:
I'm assuming I shouldn't review this until tests are passing

arahuja:
Oh sorry, didn't notice! I rebased this last night after pulling in the other PR's and didn't notice - will look!

ryan-williams:
Np, I'll look in the meantime anyway, just wasn't sure if it was somewhat
WIP or not

arahuja:
@ryan-williams Looks like it an import was dropped when rebasing, should build/pass now

ryan-williams:
This looks great, 2 small q's

---

Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 218 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-JvSHIg1qhKyxCPBy5Jz)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L218)):</sup>
why this change, OOC?

---

<sup>**[src/test/scala/org/hammerlab/guacamole/variants/AlleleEvidenceSuite.scala, line 42 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-JvSJGDGmz4ZsK7FCj1g)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/test/scala/org/hammerlab/guacamole/variants/AlleleEvidenceSuite.scala#L42)):</sup>
aren't the mismatches here `0`, `1`, `0`, for a median of `0`?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/341)

arahuja:
Review status: 0 of 12 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 218 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-JvSHIg1qhKyxCPBy5Jz)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L218)):</sup>
This (and the other one) might confusing for the same reason.  `AlleleEvidence` now computes the evidence for one specific allele.  In the tumor genotype, it's the variant allele and in the normal, the germline allele.  Hence the stats are only for a subset of the pileup elements.  We may want to know the normal's "variant" allele stats, but most often it doesn't have one (0 reads that support it)

---

<sup>**[src/test/scala/org/hammerlab/guacamole/variants/AlleleEvidenceSuite.scala, line 42 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-JvSJGDGmz4ZsK7FCj1g)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/test/scala/org/hammerlab/guacamole/variants/AlleleEvidenceSuite.scala#L42)):</sup>
It's only computing the evidence, `AlleleEvidence` for the `A/C` allele which is only the second read.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/341)

ryan-williams:
Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 216 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-Jw7xIkUKTQfi3KWis4U)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L216)):</sup>
`allele <- mostLikelyTumorGenotype.getNonReferenceAlleles.find(_.altBases.nonEmpty).toSeq`

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 218 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-JvSHIg1qhKyxCPBy5Jz)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L218)):</sup>
So we're comparing the likelihood of a specific allele in the tumor to the likelihood of the reference allele in the normal? That feels sort of weird to me. What about the allele's evidence in the normal? What if the normal has very little reference evidence but lots of reference for the "somatic" allele in question here?

I understand that on some level at this part of the code we've already decided that `somaticOdds` is high enough to warrant creating a `CalledSomaticAllele`, but we do our filter passes after this, and they are sensitive to the specific values in `normalEvidence`, `tumorEvidence`, and `phredScaledSomaticLikelihood`, so I'm not fully getting why this is what we want here.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/341)

arahuja:
Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 216 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-Jw7xIkUKTQfi3KWis4U)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L216)):</sup>
Is a there reason to `filter` to `find` or are you just suggesting the `!isEmpty` to `nonEmpty`?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 218 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-JvSHIg1qhKyxCPBy5Jz)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L218)):</sup>
My thought behind this was you want to compare the evidence the called allele for each - the variant allele for the tumor and the reference allele for the normal, with the implication that `normalEvidence` covers the normal reference and `tumorEvidence` the variant

Yes, we could actually save all four, and maybe we want to.  Mostly, I wanted to the tumor's average mapping quality etc to based on just the variant reads.  The normal's could as well, but most of the time, the normal does not have any variant reads, just making that `NaN`.  

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/341)

ryan-williams:
Review status: 0 of 12 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 216 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-Jw7xIkUKTQfi3KWis4U)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L216)):</sup>
there's two IJ warnings here, one of which I think is worth changing and one I'll defer to you on though I am marginally inclined to agree with IJ:

* `.filter(…).headOption` → `.find(…)`
* `!_.isEmpty` → `_.nonEmpty`

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 218 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-JvSHIg1qhKyxCPBy5Jz)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L218)):</sup>
OK, if what you have here is what you want for your purposes atm then I guess that's ok, but I'm still curious what your feelings are about how this code path would treat a situation where normal and tumor have lots of evidence for the same alt allele, i.e. a locus that is clearly a germline, not somatic, mutation. Don't we want to filter those out? Is that being taken care of somewhere else?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/341)

arahuja:
@ryan-williams sorry for the delay on this! I made the `filter -> find` and `!isEmpty -> nonEmpty` change and renamed some of the evidence classes/documentation that might help clarify what I wanted to do.

src/main/scala/org/hammerlab/guacamole/variants/CalledAllele.scala
```diff
@@ -41,50 +38,4 @@ case class CalledAllele(sampleName: String,
...
-      allele,
-      evidence,
-      rsId,
-      length
-    )
-
-  }
-}
-
+}
```
ryan-williams: fwiw you can have intellij add newlines to ends of files on save:

src/main/scala/org/hammerlab/guacamole/Serialization.scala
```diff
@@ -44,9 +44,6 @@ class GuacamoleKryoRegistrator extends ADAMKryoRegistrator {
     kryo.register(classOf[LociSet.SingleContig], new LociSetSingleContigSerializer)
     kryo.register(classOf[LociMap[Long]], new LociMapLongSerializer)
     kryo.register(classOf[LociMap.SingleContig[Long]], new LociMapLongSingleContigSerializer)
-    kryo.register(classOf[CalledAllele], new CalledAlleleSerializer)
```
ryan-williams: do we not need these to be serialized ever now?

ryan-williams:
Woops, should have kept discussion over on reviewable; luckily it imports from here.

I think I have [at least one question](https://reviewable.io/reviews/hammerlab/guacamole/341#-JvSHIg1qhKyxCPBy5Jz) over there still; want to go "ack" or respond to everything there and then I'll have another look?

arahuja:
> (Reviewable was unable to map this GitHub line comment to the right spot—sorry!)

> do we not need these to be serialized ever now?

They do need to be serialized, but we don't need the custom serializers (there are simple case classes which has a default kryo serializers.  We could still register them, as that would save writing the FQN and instead write a class id

---

Review status: 0 of 12 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 216 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-Jw7xIkUKTQfi3KWis4U)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L216)):</sup>
Cool, thanks for explaining, will do both.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 218 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/341#-JvSHIg1qhKyxCPBy5Jz)** ([raw file](https://github.com/hammerlab/guacamole/blob/32163c2a87885c8b0c4cb9820c747fbf0f7c939d/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L218)):</sup>
> where normal and tumor have lots of evidence for the same alt allele

Yes, that is filtered out in the line above, where at this point we are assuming it is a somatic mutation since the likelihood is much higher in the tumor than the normal,  see L210

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/341)

arahuja:
For some reason Reviewable won't let me comment on the newline thing, but thanks for the tip, I'll add that

ryan-williams:
cool, this lgtm

Issue #124, opened 2014-07-28
Apply assembly algorithms to variant-heavy regions


ryan-williams:
(summarizing @arahuja on the subject)

**Simple Version**
When many SNVs are found close together, consider dropping them under the assumption that something may have been wrong with the reads there. @arahuja has a PoC of this.

**Complex Version**
In regions where SNVs/SVs are found, run more advanced assembly algorithms, e.g. Heng Li's [fermi](https://github.com/lh3/fermi), [contrail](http://sourceforge.net/projects/contrail-bio/), other de-Bruijn's-based ones.

arahuja:
Related references:
1. Rimmer, A. et al. Integrating mapping-, assembly- and haplotype-based approaches for calling variants in clinical sequencing applications. Nat. Genet. 1–9 (2014). doi:10.1038/ng.3036
2. Rizk, G., Gouin, A., Chikhi, R. & Lemaitre, C. MindTheGap : integrated detection and assembly of short and long insertions. Bioinformatics 30, 1–7 (2014).
3. Narzisi, G. et al. Scalpel: Accurate detection of de novo and transmitted INDELs within exome-capture data using micro-assembly. bioRxiv (2013). doi:10.1101/001370
4. Mose, L. E., Wilkerson, M. D., Hayes, D. N., Perou, C. M. & Joel, S. ABRA : improved coding indel detection via assembly based re- alignment. 1–3 (2014).

PR #342, opened 2015-07-28
Simple alignment utilities

arahuja:
This is the first of a few pieces of #124, I wanted to break them off into smaller components to make it easier to review.

These adds a simple aligner and tests to ensure the correct CIGAR is genereated from the aligner.  There are probably many ways to make this a bit faster, but I will come back to that once the essential pieces are in.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/342)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
+++ b/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
@@ -1,0 +1,157 @@
+package org.hammerlab.guacamole.alignment
+
+import breeze.linalg.{ DenseMatrix, argmin }
+import org.hammerlab.guacamole.alignment.AlignmentState.AlignmentState
+
+import scala.annotation.tailrec
+
+object AffineGapPenaltyAlignment {
+
+  /**
+   * Retraces the optimal alignment path from a matrix of scores and transitions
+   *
+   * @param alignmentStates sequenceLength x referenceLength matrix of alignment score at each step
+   * @param alignmentScores sequenceLength x referenceLength matrix of best transition at each step
+   * @param sequenceLength Length of the input sequence to align
+   * @param referenceLength Length of reference sequence to align against
+   * @return An alignment path and score
+   */
+  def findBestPath(alignmentStates: DenseMatrix[AlignmentState],
+                   alignmentScores: DenseMatrix[Double],
+                   sequenceLength: Int,
+                   referenceLength: Int): ReadAlignment = {
+
+    @tailrec
+    def recFindBestPath(sequenceIdx: Int,
+                        referenceIdx: Int,
+                        currentPath: List[AlignmentState]): List[AlignmentState] = {
+
+      val lastMove = alignmentStates(sequenceIdx, referenceIdx)
+      if (sequenceIdx == 1 && referenceIdx == 1) {
+        lastMove :: currentPath
+      } else {
+        lastMove match {
+          case AlignmentState.Match | AlignmentState.Mismatch => {
+            recFindBestPath(sequenceIdx - 1, referenceIdx - 1, lastMove :: currentPath)
+          }
+          case AlignmentState.Insertion => {
+            recFindBestPath(sequenceIdx - 1, referenceIdx, lastMove :: currentPath)
+          }
+          case AlignmentState.Deletion => {
+            recFindBestPath(sequenceIdx, referenceIdx - 1, lastMove :: currentPath)
+          }
+        }
+      }
+    }
+
+    val alignments = recFindBestPath(sequenceLength, referenceLength, List.empty)
+    ReadAlignment(alignments, argmin(alignmentScores.t(::, sequenceLength)))
+  }
+
+  /**
+   * Produces an alignment of an input sequence against a reference sequence
+   *
+   * @param sequence Input sequence to align
+   * @param reference Reference sequence to align against
+   * @param mismatchProbability Penalty to having a mismatch in the alignment
+   * @param openGapProbability Penalty to start an insertion or deletion in the alignment
+   * @param closeGapProbability Penalty to end in an insertion or deletion in the alignment
+   * @return An alignment path and score
+   */
+  def align(sequence: Seq[Byte],
+            reference: Seq[Byte],
+            mismatchProbability: Double = 1e-3,
+            openGapProbability: Double = 1e-5,
+            closeGapProbability: Double = 1e-2): ReadAlignment = {
+
+    val sequenceLength = sequence.length
+    val referenceLength = reference.length
+    val (alignmentStates, alignmentScores) = scoreAlignmentPaths(sequence,
+      reference,
+      mismatchProbability,
+      openGapProbability,
+      closeGapProbability)
+
+    findBestPath(alignmentStates, alignmentScores, sequenceLength, referenceLength)
+  }
+
+  def scoreAlignmentPaths(sequence: Seq[Byte],
+                          reference: Seq[Byte],
+                          mismatchProbability: Double = 1e-2,
+                          openGapProbability: Double = 1e-3,
+                          closeGapProbability: Double = 1e-2): (DenseMatrix[AlignmentState], DenseMatrix[Double]) = {
+
+    val logMismatchPenalty = -math.round(math.log(mismatchProbability))
+    val logOpenGapPenalty = -math.round(math.log(openGapProbability))
+    val logCloseGapPenalty = -math.round(math.log(closeGapProbability))
+
+    val logContinueGapPenalty = -math.round(math.log(1 - closeGapProbability)).toInt
+    val noGapPenalty = -math.round(math.log(1 - openGapProbability)).toInt
+
+    val sequenceLength = sequence.length
+    val referenceLength = reference.length
+
+    // alignmentScores is M x N matrix
+    // M is the length of the sequence and N is the length of the reference
+    val alignmentScores = DenseMatrix.fill[Double](sequenceLength + 1, referenceLength + 1) { Double.PositiveInfinity }
+    val alignmentStates = new DenseMatrix[AlignmentState](sequenceLength + 1, referenceLength + 1)
+
+    def transitionPenalty(nextState: AlignmentState, previousState: AlignmentState, isEndState: Boolean) = {
+
+      val openGap = nextState != previousState && AlignmentState.isGapAlignment(nextState)
+      val closeGap = nextState != previousState && AlignmentState.isGapAlignment(previousState)
+      val continueGap = nextState == previousState && AlignmentState.isGapAlignment(nextState)
+      val mismatch = nextState == AlignmentState.Mismatch
+
+      (if (openGap) logOpenGapPenalty else 0) +
+        (if (closeGap) logCloseGapPenalty else 0) +
+        (if (continueGap) logContinueGapPenalty else if (mismatch) noGapPenalty + logMismatchPenalty else noGapPenalty) +
+        (if (isEndState && AlignmentState.isGapAlignment(nextState)) logCloseGapPenalty else 0)
+    }
+
+    // empty case
+    alignmentScores(0, 0) = 0
+
+    for (referenceIdx <- 1 to referenceLength) {
+      for (sequenceIdx <- 1 to sequenceLength) {
+        // Given the change is position, is the transition a gap or match/mismatch
+        def classifyTransition(prevSeqPos: Int, prevRefPos: Int): AlignmentState = {
+          if (sequenceIdx == prevSeqPos) {
+            AlignmentState.Deletion
+          } else if (referenceIdx == prevRefPos) {
+            AlignmentState.Insertion
+          } else if (sequence(sequenceIdx - 1) != reference(referenceIdx - 1)) {
+            AlignmentState.Mismatch
+          } else {
+            AlignmentState.Match
+          }
+        }
+
+        val possiblePreviousStates = Seq(
+          (sequenceIdx - 1, referenceIdx),
+          (sequenceIdx, referenceIdx - 1),
+          (sequenceIdx - 1, referenceIdx - 1)
+        )
+
+        // Compute the transition costs based on the gap penalties
+        val stateScores: Seq[(AlignmentState, Double)] = possiblePreviousStates.map({
+          case (prevSeqPos, prevRefPos) => {
+            val nextState = classifyTransition(prevSeqPos, prevRefPos)
+            val previousState = alignmentStates(prevSeqPos, prevRefPos)
+            val previousScore = alignmentScores(prevSeqPos, prevRefPos)
+            val isEndState = (sequenceIdx == sequenceLength && referenceIdx == referenceLength)
+
+            val transitionCost = transitionPenalty(nextState, previousState, isEndState = isEndState)
+            (nextState, previousScore + transitionCost)
+          }
+        })
+
+        val optimalTransition = stateScores.minBy(_._2)
+        alignmentStates(sequenceIdx, referenceIdx) = optimalTransition._1
+        alignmentScores(sequenceIdx, referenceIdx) = optimalTransition._2
+      }
+    }
+
+    (alignmentStates, alignmentScores)
+  }
+}
--- a/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
+++ b/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
@@ -10,8 +10,8 @@ object AffineGapPenaltyAlignment {
   /**
    * Retraces the optimal alignment path from a matrix of scores and transitions
    *
-   * @param alignmentStates sequenceLength x referenceLength matrix of alignment score at each step
-   * @param alignmentScores sequenceLength x referenceLength matrix of best transition at each step
+   * @param alignmentStates (sequenceLength + 1) x (referenceLength + 1) matrix of best transition at each step
+   * @param alignmentScores (sequenceLength + 1) x (referenceLength + 1) matrix of alignment score at each step
    * @param sequenceLength Length of the input sequence to align
    * @param referenceLength Length of reference sequence to align against
    * @return An alignment path and score
@@ -31,15 +31,12 @@ object AffineGapPenaltyAlignment {
         lastMove :: currentPath
       } else {
         lastMove match {
-          case AlignmentState.Match | AlignmentState.Mismatch => {
+          case AlignmentState.Match | AlignmentState.Mismatch =>
             recFindBestPath(sequenceIdx - 1, referenceIdx - 1, lastMove :: currentPath)
-          }
-          case AlignmentState.Insertion => {
+          case AlignmentState.Insertion =>
             recFindBestPath(sequenceIdx - 1, referenceIdx, lastMove :: currentPath)
-          }
-          case AlignmentState.Deletion => {
+          case AlignmentState.Deletion =>
             recFindBestPath(sequenceIdx, referenceIdx - 1, lastMove :: currentPath)
-          }
         }
       }
     }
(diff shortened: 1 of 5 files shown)


src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
```diff
@@ -0,0 +1,157 @@
...
+import org.hammerlab.guacamole.alignment.AlignmentState.AlignmentState
+
+import scala.annotation.tailrec
+
+object AffineGapPenaltyAlignment {
+
+  /**
+   * Retraces the optimal alignment path from a matrix of scores and transitions
+   *
+   * @param alignmentStates sequenceLength x referenceLength matrix of alignment score at each step
```
ryan-williams: "alignment score" here is confusing given this parameter, and its data type, are "alignmentStates" and the next parameter is "alignmentScores" / Doubles.

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
```diff
@@ -0,0 +1,157 @@
...
+                   alignmentScores: DenseMatrix[Double],
+                   sequenceLength: Int,
+                   referenceLength: Int): ReadAlignment = {
+
+    @tailrec
+    def recFindBestPath(sequenceIdx: Int,
+                        referenceIdx: Int,
+                        currentPath: List[AlignmentState]): List[AlignmentState] = {
+
+      val lastMove = alignmentStates(sequenceIdx, referenceIdx)
```
ryan-williams: maybe a basic breeze question, but doesn't this look for an invalid row and column the first time this is called, at `(sequenceLength, referenceLength)`?

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
```diff
@@ -0,0 +1,157 @@
...
+   * @param sequence Input sequence to align
+   * @param reference Reference sequence to align against
+   * @param mismatchProbability Penalty to having a mismatch in the alignment
+   * @param openGapProbability Penalty to start an insertion or deletion in the alignment
+   * @param closeGapProbability Penalty to end in an insertion or deletion in the alignment
+   * @return An alignment path and score
+   */
+  def align(sequence: Seq[Byte],
+            reference: Seq[Byte],
+            mismatchProbability: Double = 1e-3,
```
ryan-williams: were these default values taken from / inspired by anything worth scaladoc'ing?

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
ryan-williams: ah just read on and saw that the sizes are actually `(sequenceLength + 1, referenceLength + 1)`; saying they're `sequenceLength x referenceLength` above seems confusing in that light

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
```diff
@@ -0,0 +1,157 @@
...
+    findBestPath(alignmentStates, alignmentScores, sequenceLength, referenceLength)
+  }
+
+  def scoreAlignmentPaths(sequence: Seq[Byte],
+                          reference: Seq[Byte],
+                          mismatchProbability: Double = 1e-2,
+                          openGapProbability: Double = 1e-3,
+                          closeGapProbability: Double = 1e-2): (DenseMatrix[AlignmentState], DenseMatrix[Double]) = {
+
+    val logMismatchPenalty = -math.round(math.log(mismatchProbability))
```
ryan-williams: why all the `math.round`ing, OOC?

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
```diff
@@ -0,0 +1,157 @@
...
+        (if (continueGap) logContinueGapPenalty else if (mismatch) noGapPenalty + logMismatchPenalty else noGapPenalty) +
+        (if (isEndState && AlignmentState.isGapAlignment(nextState)) logCloseGapPenalty else 0)
+    }
+
+    // empty case
+    alignmentScores(0, 0) = 0
+
+    for (referenceIdx <- 1 to referenceLength) {
+      for (sequenceIdx <- 1 to sequenceLength) {
+        // Given the change is position, is the transition a gap or match/mismatch
```
ryan-williams: `s/is/in/`

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
```diff
@@ -0,0 +1,157 @@
...
+    def recFindBestPath(sequenceIdx: Int,
+                        referenceIdx: Int,
+                        currentPath: List[AlignmentState]): List[AlignmentState] = {
+
+      val lastMove = alignmentStates(sequenceIdx, referenceIdx)
+      if (sequenceIdx == 1 && referenceIdx == 1) {
+        lastMove :: currentPath
+      } else {
+        lastMove match {
+          case AlignmentState.Match | AlignmentState.Mismatch => {
```
ryan-williams: jfyi, intellij says http://cl.ly/image/1h2X3w1T3r3h

src/main/scala/org/hammerlab/guacamole/alignment/ReadAlignment.scala
```diff
@@ -0,0 +1,62 @@
...
+          currentRun = 1
+        }
+        lastOperator = operators(i)
+        i += 1
+      }
+      rle.append(currentRun.toString + lastOperator)
+      rle.toString
+    }
+
+    runLengthEncode(alignments.map(cigarKey(_)))
```
ryan-williams: nit:

src/test/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignmentSuite.scala
```diff
@@ -0,0 +1,70 @@
...
+      .align(
+        "ATTCTCAAGTTTTAAGTGGTATTCTAATTATGGCAGTAATTAACTGAATAAAGAGATTCATCATGTGCAAAAACTAATCTTGTTTACTTAAAATTGAGAGT", 
+        "ATTCTCAAGTTTTAAGTGGTTTTCTAATTATGGCAGTAATAAACTGAATAAAGAGATTCATCATGTGCAAAAACTAATCTTGTTTACTTAAAATTGAGAGT")
+    alignment.toCigar should be("20=1X19=1X60=")
+  }
+
+  test("2 mismatch with deletion sequence") {
+    val alignment = AffineGapPenaltyAlignment
+      .align(
+        "ATTCTCAAGTTTTAAGTGGTATTCTAATTATGGCAGTAATTAACTGAATAAAGAGATTCATCATGTGCAAAAACTAATCTTGTTTACTTAAAATTGAGAGT",
```
ryan-williams: maybe a little clearer:

```
       //====================X===================X========================================DDD====================
        "ATTCTCAAGTTTTAAGTGGTATTCTAATTATGGCAGTAATTAACTGAATAAAGAGATTCATCATGTGCAAAAACTAATCTT"+"GTTTACTTAAAATTGAGAGT",
        "ATTCTCAAGTTTTAAGTGGTTTTCTAATTATGGCAGTAATAAACTGAATAAAGAGATTCATCATGTGCAAAAACTAATCTTCCCGTTTACTTAAAATTGAGAGT")
```

src/test/scala/org/hammerlab/guacamole/alignment/ReadAlignmentSuite.scala
```diff
@@ -0,0 +1,57 @@
...
+        AlignmentState.Match,
+        AlignmentState.Match,
+        AlignmentState.Match), 60)
+
+    alignment.toCigar should be("6=")
+  }
+
+  test("test cigar string: mixed match/insertion") {
+    val alignment = ReadAlignment(
+      Seq(AlignmentState.Match,
```
ryan-williams: I like the `Seq(` on its own line format you used above and below fwiw

ryan-williams:
This looks great @arahuja; I dropped a couple non-binding nits

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
arahuja: I think I flipped the comments on those

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
arahuja: Yup, will update the comment

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
arahuja: Not particularly, and they do need more experimentation.
Right now these result in reasonable alignments, but for example BWA sets these as 
mismatchProbability = e^-4 ~=0.018
gapOpen = e^-6 ~=0.0024
gapExtends= e^-1 ~=0.36

Plugging those in actually does give the same alignments as well, but seemed a bit more arbitrary.

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
arahuja: No reason anymore, originally was scoring everything as `int`

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
arahuja: Cool, cleaning up.

src/test/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignmentSuite.scala
arahuja: Great, thanks for this! adding to both test cases

src/main/scala/org/hammerlab/guacamole/alignment/ReadAlignment.scala
arahuja: Dropping the `_`

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
arahuja: Actually I set them to the BWA defaults for now

arahuja:
@ryan-williams thanks for the feedback, I think I responded to everything if you want to take another look

ryan-williams:
thx arun, lgtm, but Travis shows a test OOM. thoughts?

arahuja:
Hm, weird, seems to be coming from:
```
StructuralVariantCallerSuite:
- median stats
- read compatibility
insert stats: MedianStats(100.0,1.0)
- read filtering
An exception or error caused a run to abort. This may have been caused by a problematic custom reporter.
Exception in thread "ScalaTest-main" 
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "ScalaTest-main"
```
but works for me locally.  Also the push branch works, but no the PR ... might just force it to run again

ryan-williams:
@arahuja lmk if you'd like another review pass, Travis issues notwithstanding

arahuja:
@ryan-williams I don't think I've added anything since your last review, but if there will still things outstanding let me know!

arahuja:
Finally! Travis is happy! by adding ` -XX:MaxPermSize=512m` to scalatest `argLine` argument.

@ryan-williams let me know if you want to take another look, nothing changed other than that since the last time, but if you see anything you wanted to comment on lmk

ryan-williams:
looking

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
```diff
@@ -0,0 +1,159 @@
...
+      val mismatch = nextState == AlignmentState.Mismatch
+
+      (if (openGap) logOpenGapPenalty else 0) +
+        (if (closeGap) logCloseGapPenalty else 0) +
+        (if (continueGap) logContinueGapPenalty else if (mismatch) noGapPenalty + logMismatchPenalty else noGapPenalty) +
+        (if (isEndState && AlignmentState.isGapAlignment(nextState)) logCloseGapPenalty else 0)
+    }
+
+    // empty case
+    alignmentScores(0, 0) = 0
```
ryan-williams: What happens to the rest of the 0th row and 0th column? They get default scores of ∞, meaning we don't allow alignments that start with an insertion or deletion, right?

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
```diff
@@ -0,0 +1,159 @@
...
+
+    val sequenceLength = sequence.length
+    val referenceLength = reference.length
+    val (alignmentStates, alignmentScores) = scoreAlignmentPaths(sequence,
+      reference,
+      mismatchProbability,
+      openGapProbability,
+      closeGapProbability)
+
+    findBestPath(alignmentStates, alignmentScores, sequenceLength, referenceLength)
```
ryan-williams: one thing that just occurred to me: it's interesting that you're building this matrix and then traversing back through it to back out the best path; it could be built in such a way that each element stores the best path to it, saving this second traversal, right?

I assume that would be more memory-consumptive and therefore not worth it, just wanted to confirm that we're on the same page about that.

Also, very much as an aside: it seems like there's a pruned search that could be done here, as a possible optimization, using a "frontier" pqueue that keeps all next-to-be-expanded matrix positions in order of increasing score (starting with just `(0,0) → 0`… I'm sure there is literature about this, I will think about it some more.

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
```diff
@@ -0,0 +1,159 @@
...
+          (sequenceIdx - 1, referenceIdx),
+          (sequenceIdx, referenceIdx - 1),
+          (sequenceIdx - 1, referenceIdx - 1)
+        )
+
+        // Compute the transition costs based on the gap penalties
+        val stateScores: Seq[(AlignmentState, Double)] = possiblePreviousStates.map({
+          case (prevSeqPos, prevRefPos) => {
+            val nextState = classifyTransition(prevSeqPos, prevRefPos)
+            val previousState = alignmentStates(prevSeqPos, prevRefPos)
```
ryan-williams: When `(prevSeqPos, prevRefPos) == (0, 0)`, is `previousState == null`? Or does it default to the enum's "0", aka `Match`?

If the latter, are we relying on that behavior intentionally or accidentally? Should we make a note about it?

One, ahem, option is to make it an `Option[AlignmentState]`.

I'm also thinking about the version where each element keeps its best path.

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
```diff
@@ -0,0 +1,159 @@
...
+            val previousScore = alignmentScores(prevSeqPos, prevRefPos)
+            val isEndState = (sequenceIdx == sequenceLength && referenceIdx == referenceLength)
```
ryan-williams: similarly to another question I had, should we force every path to consume all passed-in reference bases?

What if our read ends nicely before the end of the reference; do we force it to take a deletion penalty?

Should "end state" be anything on the right column?

ryan-williams:
heh so it turns out I have a few questions! and I sent #353 on this with some ideas I'd like your thoughts on.

I'm not 100% sold on one-pass being the way to go, but that's independent from several other things I've touched on in #353 and the comments here.

re: the `-XX:MaxPermSize=512m`, any idea why that's necessary here?

arahuja:
> re: the -XX:MaxPermSize=512m, any idea why that's necessary here?

on Travis the `StructuralVariantCaller` tests were throwing OOM.

ryan-williams:
Right, sorry, what I meant is, are we doing particularly memory-hungry things in those tests?

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
arahuja: Yes, that was the idea here, starting with a match/mismatch and the proceeding.

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
arahuja: Yes, you are right here, it was relying on that behavior accidently, but you've fixed this in your PR

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
arahuja: Cool, yea aligning anywhere would be great, thanks for adding that!

arahuja:
@ryan-williams I think your PR is a big improvement overall, I posted some questions there, the only thing I'm not sure about is saving the paths as we go.  It also seems to me we should be able to drop some of the matrix as we proceed, since we only lookback into 3 positions - seems like you would only need the current and last row, but I need to think about it more.

ryan-williams:
ha, good call, that's basically the "pruned search" I was referring to, though I didn't make the full leap to that we don't need to store a whole matrix.

I don't want to derail this too much; that might be a thing one of us could knock out in a few hours but I'm about to go OOO for 2wks. 

I am fine with any of

1. going back to your two-pass impl (matrix-forward + backwards-trace to assemble best path) while keeping some of my other fixes,
2. using my full-matrix-full-paths one-pass approach,
3. waiting on an attempt at the one-pass "frontier pqueue" impl, where we only keep the to-be-expanded entries of the matrix, starting with (0,0), and iterate "expandin" the cheapest entry from that pqueue, so that the first path to hit (M+1,N+1) is the winner! that does sound like a cool algo but I'm not going to do it today so :)

Either way, we can also run a few versions on the cluster and see if the memory-stats in the UI tell us interesting things about the tradeoffs; we should probably do that either way "eventually"...

ryan-williams:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala, line 115 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/342#-JvL7Sfd0eM23BAQZkpm-r4-115)** ([raw file](https://github.com/hammerlab/guacamole/blob/7f02946f54fc40738a28b02e857d9a356ff4a05a/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala#L115)):</sup>
Cool; my impression from our chats has been that we would like the given seqbases to be able to align starting anywhere in the given refbases. lmk if you're not on board with that change from #353 

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala, line 142 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/342#-JvL7Sfd0eM23BAQZkpm-r4-142)** ([raw file](https://github.com/hammerlab/guacamole/blob/7f02946f54fc40738a28b02e857d9a356ff4a05a/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala#L142)):</sup>
OK, I'll leave this open (in reviewable) until we get a/that fix merged in here, I guess?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala, line 144 \[r4\]](https://reviewable.io:443/reviews/hammerlab/guacamole/342#-JvL7Sfd0eM23BAQZkpm-r4-144)** ([raw file](https://github.com/hammerlab/guacamole/blob/7f02946f54fc40738a28b02e857d9a356ff4a05a/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala#L144)):</sup>
Cool, same as above

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/342)

arahuja:
@ryan-williams I think the other fixes are key, if you can pull those out and we can merge them in and then move the single pass as a separate PR then we can benchmark both?  I've been playing with scalabench a bit so that might give us a good number.

I'm also ok with merging the whole single-pass version in, as I think option 3 will be the way to go and once that's in i think it'll be straightforward to adapt it to save less of the matrix.

ryan-williams:
OK, version with two passes plus other fixes makes sense.

What is your interest level in taking this and making that? I am just worried about promising stuff today; juggling a few things I want to get done before I head out :-\

arahuja:
Sure, will carve those pieces out!

PR #343, opened 2015-07-29
Find cliques within the SV Graph and output them

danvk:
DELLY's heuristic for finding a maximal clique is:

```
For each connected component:
  Collect all the edges within the component
  Sort the edges by weight (increasing)
  Seed the clique with the vertex that is the source of the lowest-weight edge.
      (Because of the way that the graph is constructed, this will have an earlier start position.)

  Repeat:
    Find the lowest-weight edge which connects a new node to the current clique.
    Add it to the clique if it's fully-connected and the resulting SV would still place all insert sizes in range.
```

Now that I have some tests in place, I'd like to refactor the clique-growing logic to be a bit cleaner.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/343)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -171,6 +171,7 @@ object StructuralVariant {
           } else {
             if (areReadsCompatible(read, nextRead, maxNormalInsertSize)) {
               // TODO: this is from DELLY but seems imprecise. It only considers the insert size, not position.
+              // If two reads are of similar length, but don't overlap much, they'll get a spuriously low weight.
               val weight = Math.abs((nextGapEnd - nextStart) - (gapEnd - start))
               graph += (read ~ nextRead) % weight
             }
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -8,7 +8,6 @@ import org.kohsuke.args4j.{ Option => Args4JOption }
 
 import scalax.collection.Graph
 import scalax.collection.edge._
-import scala.collection.mutable
 import scalax.collection.mutable.{ Graph => MutableGraph }
 import scalax.collection.GraphPredef._
 import scalax.collection.edge.Implicits._
@@ -16,12 +15,12 @@ import scalax.collection.edge.Implicits._
 /**
  * Structural Variant caller
  *
- * This currently looks for regions with abnormally large insert sizes. This typically indicates
- * that there's been a large deletion.
+ * This follows the DELLY algorithm:
+ * 1. Look for read pairs with abnormally large insert sizes.
+ * 2. Build a graph of pairs which could be explained by a single SV.
+ * 3. Find maximal cliques in this graph.
  *
- * It does this by finding the average insert size of all mate pairs which overlap each 25-base pair "block" of the
- * genome. If this average is significantly different than the average insert size genome-wide, then it's evidence that
- * there may be a structural variant.
+ * These cliques each represent a large deletion.
  *
  * The output is a text file of genomic ranges where there's some evidence of an SV.
  */
@@ -114,7 +113,7 @@ object StructuralVariant {
       val insertStats = medianStats(insertSizes.takeSample(false, 100000))
       println("insert stats: " + insertStats)
 
-      // Find blocks where the average insert size differs significantly from the genome-wide median.
+      // Find read pairs where the insert size differs significantly from the genome-wide median.
       // Following DELLY, we use a threshold of 5 median absolute deviations above the median.
       val maxNormalInsertSize = (insertStats.median + 5 * insertStats.mad).toInt
 
@@ -129,8 +128,10 @@ object StructuralVariant {
       )
     }
 
-    // Given two pairs of reads, is there a deletion which would make both of them have normal insert sizes?
-    def areReadsCompatible(read1: PairedMappedRead, read2: PairedMappedRead, maxNormalInsertSize: Long): Boolean = {
+    // Given two pairs of reads, is there a deletion which would make their insert sizes normal?
+    def areReadsCompatible(read1: PairedMappedRead,
+                           read2: PairedMappedRead,
+                           maxNormalInsertSize: Long): Boolean = {
       if (read1.minPos > read2.minPos) {
         areReadsCompatible(read2, read1, maxNormalInsertSize)
       } else {
@@ -149,12 +150,14 @@ object StructuralVariant {
       }
     }
 
-    // Given a list of reads with abnormally large inserts, construct a graph representation of the structural variants
-    // where:
+    // Given a list of reads with abnormally large inserts, construct a graph representation of the
+    // structural variants where:
     // Vertices = paired reads
     // Edges = two sets of paired reads could represent the same structural variant
+    // Weights = compatibility between reads (0=most compatible)
     // For more details, see the DELLY paper.
-    def buildVariantGraph(exceptionalReads: Iterable[PairedMappedRead], maxNormalInsertSize: Int): PairedReadGraph = {
+    def buildVariantGraph(exceptionalReads: Iterable[PairedMappedRead],
+                          maxNormalInsertSize: Int): PairedReadGraph = {
       val reads = exceptionalReads.toArray.sortBy(_.minPos)
       val graph = MutableGraph[PairedMappedRead, WUnDiEdge]()
 
@@ -170,8 +173,9 @@ object StructuralVariant {
             done = true
           } else {
             if (areReadsCompatible(read, nextRead, maxNormalInsertSize)) {
-              // TODO: this is from DELLY but seems imprecise. It only considers the insert size, not position.
-              // If two reads are of similar length, but don't overlap much, they'll get a spuriously low weight.
+              // TODO: this is from DELLY but seems imprecise. It only considers the insert size,
+              // not position. If two reads are of similar length, but don't overlap much, they'll
+              // get a spuriously low weight.
               val weight = Math.abs((nextGapEnd - nextStart) - (gapEnd - start))
               graph += (read ~ nextRead) % weight
             }
@@ -190,6 +194,7 @@ object StructuralVariant {
                         svEnd: Long,
                         maxNormalInsertSize: Int) {
 
+      // All connected components will have at least one node (by definition), so .head is safe.
       def span = GenomeRange(readPairs.head.read.referenceContig, svStart, svEnd)
 
       // If an edge has one node in the clique and one not, return the one that's not.
@@ -215,8 +220,8 @@ object StructuralVariant {
         val newSvStart = Math.max(svStart, nodeGapMin)
         val newSvEnd = Math.min(svEnd, nodeGapMax)
 
-        // The amount of wiggle room could shrink either because of the new read, or because it shrinks the
-        // the induced structural variant and pulls another read closer to being incompatible.
+        // The amount of wiggle room could shrink either because of the new read, or because it
+        // shrinks the induced SV and pulls another read closer to being incompatible.
         val wiggleNewRead = maxNormalInsertSize - (n.value.insertSize - (newSvEnd - newSvStart))
         val wiggleChange = wiggle + (newSvEnd - newSvStart) - (svEnd - svStart)
         val newWiggle = Math.min(wiggleNewRead, wiggleChange)
@@ -232,8 +237,9 @@ object StructuralVariant {
     object SVClique {
       def apply(pair: PairedMappedRead, maxNormalInsertSize: Int): SVClique = {
         val (_, svStart, svEnd, _) = pair.startsAndStops
-        // this is the "wiggle room": by how much can the SV shrink and still be compatible with all reads in the clique?
-        // It starts negative and then increases towards zero (at which point there's no wiggle room left).
+        // this is the "wiggle room": by how much can the SV shrink and still be compatible with all
+        // reads in the clique? It starts positive and then shrinks towards zero (at which point
+        // there's no wiggle room left).
         val wiggle = maxNormalInsertSize - ((pair.insertSize - (svEnd - svStart)))
         SVClique(Set(pair), wiggle, svStart, svEnd, maxNormalInsertSize)
       }
@@ -241,7 +247,8 @@ object StructuralVariant {
 
     def findOneClique(g: PairedReadGraph, maxNormalInsertSize: Int): SVClique = {
       // Seed the clique with the source of the lowest-weight edge.
-      // Because the graph is constructed by adding edges, all connected components will have at least one edge.
+      // Because the graph is constructed by adding edges, all connected components will have at
+      // least one edge.
       val edges = g.edges.toList.sortBy(_.weight)
       val bestEdge = edges.head
 
@@ -268,11 +275,13 @@ object StructuralVariant {
       val pairedMappedReads = readSet.mappedPairedReads
       val firstInPair = pairedMappedReads.filter(_.isFirstInPair).flatMap(PairedMappedRead(_))
 
-      val ExceptionalReadsReturnType(_, _, _, maxNormalInsertSize, exceptionalReads) = getExceptionalReads(firstInPair)
+      val ExceptionalReadsReturnType(_, _, _, maxNormalInsertSize, exceptionalReads) =
+          getExceptionalReads(firstInPair)
 
       val readsByContig = exceptionalReads.groupBy(_.read.referenceContig)
       val svGraph = readsByContig.mapValues(buildVariantGraph(_, maxNormalInsertSize))
-      val svs = svGraph.mapValues(findCliques(_, maxNormalInsertSize)).mapValues(_.map(_.span).toList)
+      val svs = svGraph.mapValues(findCliques(_, maxNormalInsertSize))
+                       .mapValues(_.map(_.span).toList)
 
       svs.coalesce(1).saveAsTextFile(args.output)
     }
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -8,6 +8,7 @@ import org.kohsuke.args4j.{ Option => Args4JOption }
 
 import scalax.collection.Graph
 import scalax.collection.edge._
+import scala.collection.mutable
 import scalax.collection.mutable.{ Graph => MutableGraph }
 import scalax.collection.GraphPredef._
 import scalax.collection.edge.Implicits._
@@ -158,18 +159,19 @@ object StructuralVariant {
       val graph = MutableGraph[PairedMappedRead, WUnDiEdge]()
 
       for { (read, i) <- reads.zipWithIndex } {
-        val start = read.minPos
+        val (start, _, end, _) = read.startsAndStops
         var j = i + 1
         var done = false
         // Loop through subsequent reads until it's impossible to find another compatible one.
         while (j < reads.length && !done) {
           val nextRead = reads(j)
-          val nextStart = nextRead.minPos
+          val (nextStart, _, nextEnd, _) = nextRead.startsAndStops
           if (Math.abs(nextStart + nextRead.readLength - start) > maxNormalInsertSize) {
             done = true
           } else {
             if (areReadsCompatible(read, nextRead, maxNormalInsertSize)) {
-              graph += (read ~ nextRead) % 1
+              val weight = Math.abs((nextEnd - nextStart) - (end - start))
+              graph += (read ~ nextRead) % Math.max(30000, weight)
             }
           }
           j += 1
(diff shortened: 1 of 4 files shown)


src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -180,6 +182,89 @@ object StructuralVariant {
       graph
     }
 
+    case class SVClique(
+      reads: Set[PairedMappedRead],
+      span: GenomeRange,
+      wiggle: Long
+    )
+
+    def findOneClique(g: PairedReadGraph, maxNormalInsertSize: Int): SVClique = {
```
danvk: This code is a bit cryptic & largely based on DELLY. I'll be experimenting with cleaning it up, but suggestions would be appreciated.

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -180,6 +182,89 @@ object StructuralVariant {
...
+
+          // The new node should be fully connected with the rest of the clique.
+          node filter { n =>
+            {
+              clique.subsetOf(n.neighbors map { _.value })
+            }
+          } filter { n =>
+            // And it should be compatible with the existing SV.
+            {
+              val (_, nodeGapMin, nodeGapMax, _) = n.value.startsAndStops
```
danvk: FYI, this is a line that IntelliJ doesn't like: https://youtrack.jetbrains.com/issue/IDEA-143285#

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -158,18 +159,19 @@ object StructuralVariant {
...
         while (j < reads.length && !done) {
           val nextRead = reads(j)
-          val nextStart = nextRead.minPos
+          val (nextStart, _, nextEnd, _) = nextRead.startsAndStops
           if (Math.abs(nextStart + nextRead.readLength - start) > maxNormalInsertSize) {
             done = true
           } else {
             if (areReadsCompatible(read, nextRead, maxNormalInsertSize)) {
-              graph += (read ~ nextRead) % 1
+              val weight = Math.abs((nextEnd - nextStart) - (end - start))
```
ryan-williams: what's going on here? difference of [distance from start of first read to start of second read] between the two pairs? "end" and "nextEnd" are not terribly clear; those are what I was calling "gap end"s right?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -158,18 +159,19 @@ object StructuralVariant {
...
+              graph += (read ~ nextRead) % Math.max(30000, weight)
```
ryan-williams: 30000 as a default minimum weight? plz explain?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -180,6 +182,99 @@ object StructuralVariant {
...
+    case class SVClique(
+      reads: Set[PairedMappedRead],
+      span: GenomeRange,
+      wiggle: Long
+    )
+
+    def findOneClique(g: PairedReadGraph, maxNormalInsertSize: Int): SVClique = {
+      // Seed the clique with the source of the lowest-weight edge.
+      val edges = g.edges.toList.sortBy(_.weight)
+      val bestEdge = edges.head
```
ryan-williams: do we know edges to be non-empty? "edge"-case, I know, but j/w

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -180,6 +182,99 @@ object StructuralVariant {
...
+
+      val initPair = bestEdge.nodes.minBy(_.minPos).value
+      var (pairMin, svStart, svEnd, _) = initPair.startsAndStops
+
+      // this is the "wiggle room": by how much can the SV shrink and still be compatible with all reads in the clique?
+      var wiggle = initPair.insertSize - maxNormalInsertSize - (svEnd - svStart)
```
ryan-williams: I'm confused by this calculation. What size is the SV shrinking from? Presumably `svEnd - svStart`? Either way, the read-pair currently spans `initPair.insertSize` and must still be at least `maxNormalInsertSize`, so the amount it (and therefore its gap) can shrink is `initPair.insertSize - maxNormalInsertSize`. Why subtracting `(svEnd - svStart)`?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -180,6 +182,99 @@ object StructuralVariant {
...
+
+      var cliqueGrow = true
+      while (cliqueGrow && clique.size < graphSize) {
+        cliqueGrow = false
+        // Try to find a node which:
+        // 1. isn't currently in the clique
+        // 2. is connected to every node in the clique
+        // 3. doesn't perturb the SV in a way that makes it incompatible with any reads in the clique
+        mapFirst(edges, (edge: PairedReadGraph#EdgeT) => {
+          val nodes = edge.nodes.toSeq
```
ryan-williams: could combine this line with the next no?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -180,6 +182,99 @@ object StructuralVariant {
...
+              } else {
+                false
+              }
+            }
+          }
+        })
+      }
+
+      val reads = clique.map(_.value).toSet
+      SVClique(reads=reads,
```
ryan-williams: spaces around `=` typically

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
ryan-williams: OK, I think I get it now: the SV starts as the whole gap, so the insert size is `initPair.insertSize - (svEnd - svStart)`, or equivalently: read length + mate length.

The question is how much that read pair can expand (shrinking the SV) to stay under `maxNormalInsertSize`.

I would think the answer is: `maxNormalInsertSize` minus current insert size (i.e. `initPair.insertSize - (svEnd - svStart)`).

That is the additive inverse of what you have here. Is this quantity negative?

ryan-williams: This and later calculations could possibly be clarified by giving `PairedMappedRead` a `.readsLength` method equal to `insertSize - gapSize`.

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -180,6 +182,99 @@ object StructuralVariant {
...
+            {
+              val (_, nodeGapMin, nodeGapMax, _) = n.value.startsAndStops
+              val newSvStart = Math.max(svStart, nodeGapMin)
+              val newSvEnd = Math.min(svEnd, nodeGapMax)
+
+              // The amount of wiggle room could shrink either because of the new read, or because it shrinks the
+              // the induced structural variant and pulls another read closer to being incompatible.
+              val wiggleNewRead = n.value.insertSize - maxNormalInsertSize - (newSvEnd - newSvStart)
+              val wiggleChange = wiggle + (svEnd - svStart) - (newSvEnd - newSvStart)
+              val newWiggle = Math.max(wiggleNewRead, wiggleChange)
```
ryan-williams: to be clear, you are `max`ing two negative numbers, which is sort of like shrinking their absolute values?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -180,6 +182,99 @@ object StructuralVariant {
...
+
+              if ((newSvStart < newSvEnd) && (newWiggle <= 0)) {
+                svStart = newSvStart
+                svEnd = newSvEnd
+                wiggle = newWiggle
+                cliqueGrow = true
```
ryan-williams: you could kill `cliqueGrow` if you just check `.isDefined` on the `mapFirst`'s returned `Option`, I think?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -180,6 +182,99 @@ object StructuralVariant {
...
+
+    def mapFirst[A, B](seq: Seq[A], f: (A) => Option[B]): Option[B] = {
+      seq.view.map(f).find(_.isDefined).flatten
+    }
+
+    def findCliques(g: PairedReadGraph, maxNormalInsertSize: Int): Traversable[SVClique] = {
+      for {
+        c <- g.componentTraverser()
+        if c.nodes.size >= 2
+      } yield findOneClique(c.toGraph, maxNormalInsertSize)
```
ryan-williams: this feels like it will miss the largest clique in the component a lot of the time. is the heuristic better than I'm giving it credit for?

src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
```diff
@@ -189,8 +284,9 @@ object StructuralVariant {
 
       val readsByContig = exceptionalReads.groupBy(_.read.referenceContig)
       val svGraph = readsByContig.mapValues(buildVariantGraph(_, maxNormalInsertSize))
+      val svs = svGraph.mapValues(findCliques(_, maxNormalInsertSize)).mapValues(x => x.map(_.span).toList)
```
ryan-williams: `s/x => x/_/`

ryan-williams:
this looks good @danvk. a few comments in there but I think I get it overall

danvk:
Thanks for the review, Ryan. It's too bad you had to miss the Lunch & Learn on the DELLY algorithm yesterday!

I addressed your comments & did some slight refactoring on `findOneClique`.

---

Review status: 0 of 4 files reviewed at latest revision, 11 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 173 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-173)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L173)):</sup>
Yep! I changed the variable names.

BTW, this is another example of DELLY being very loose. The `weight` here only considers how long the inserts are, not how much they overlap. I added a TODO to change that.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 174 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-174)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L174)):</sup>
It's a default maximum, hopefully that makes a little more sense :)

This comes [straight from DELLY][1]. It makes very little sense logically (why would you want to saturate weights?). My best guess is that it's to avoid integer overflows. I just dropped it.

[1]: https://github.com/tobiasrausch/delly/blob/7464cacfe1d420ac5bd7ed40a3093a80a93a6964/src/delly.cpp#L2074

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 194 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-194)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L194)):</sup>
Yes. We only add edges to the graph, so there can't be any isolated nodes. And hence every connected component has at least one edge.

Is it standard to `assert` things like this in Scala? In any case, I added a note.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 200 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-200)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L200)):</sup>
You're right, it is negative and grows towards zero, at which point there's no "wiggle room" left. I'm not sure why DELLY does it that way, but hence the `Math.abs` when it returns `wiggle`. I added a comment about this.

Adding a `.readsLength` method would simplify this expression, but then I'd lose the symmetry with expression on line 242 below, which couldn't be simplified.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 214 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-214)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L214)):</sup>
Done.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 244 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-244)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L244)):</sup>
Unless one of them is greater than zero, in which case we want to bail.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 250 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-250)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L250)):</sup>
Done. I had to move `newNode` out into a `var` and switch to a `do`/`while` loop, though. Were you thinking of something else? (I can't refer to `newNode` in the `while` clause without pulling it out into a `var`.)

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 262 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-262)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L262)):</sup>
Done.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 275 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-275)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L275)):</sup>
Sorting by weights should make it pretty effective—the most similar reads will be included in the clique when it's small (and easier to grow).

It does feel pretty unstable, though. For example, if there are two reads in very strong agreement and then a huge clique that's only loosely connected to them, it will miss the huge clique in favor of the two agreeing reads.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 287 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-287)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L287)):</sup>
Done. I have trouble telling when this is going to work and when expressions with two `_` will be treated as multi-parameter functions.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/343)

ryan-williams:
wha, Alex said he was doing NN L&L, Isaac posted on the wiki saying he did an Ibis L&L, and now you say you did a delly L&L... quite a day to miss

---

Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 175 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-174)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L175)):</sup>
nit, but I think it was accurately described as "a default minimum weight".

you were doing `math.max(30000, weight)` IIRC; that expression can't be less than 30000: the max/30000 meant that the expression's minimum possible value was 30000. I guess calling it "default" was misleading since it was not change-able.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 195 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-194)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L195)):</sup>
> We only add edges to the graph, so there can't be any isolated nodes

hm I don't see how that follows. If you have a graph with nodes but no edges, does the `componentTraverser` return one component per isolated node, or no components?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 256 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-250)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L256)):</sup>
I think reworking this loop more makes sense.

You should only have to do one pass through `edges`, greedily adding things to the `clique` or not as you go.

Seems well-suited to a `foldLeft`.

I just sent you https://github.com/hammerlab/guacamole/pull/346 with an alternative way of doing this, curious to hear your thoughts.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 281 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-275)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L281)):</sup>
Well, you're using "similar" and "in very strong agreement" to mean "having the same insert sizes" here; the bad cases will be ones where equally-sized pairs overlap but not by as large an amount as possible. Presumably, a lot of the time, they will not be identically-positioned pairs, so you will lose some wiggle room right out of the gate.

As you've probably picked up on, I am very sensitive to committing code that we know to contain these... subtleties... and not documenting them clearly while we have state on them.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 293 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-287)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L293)):</sup>
Makes sense. Two `_`s in one function --> multi-parameter. Here the `_`s are in different functions, one in each.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/343)

danvk:
Yeah, it was a good L&L!

---

Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 174 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-174)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L174)):</sup>
That is a great catch. It should have been `Math.min`.

What I don't get is why [this test][1] didn't catch the bug. Is it a problem with Scala Graph?

[1]: https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/test/scala/org/hammerlab/guacamole/commands/StructuralVariantCallerSuite.scala#L121

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 195 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-194)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L195)):</sup>
This is the only place where anything is added to the graph:

    graph += (read ~ nextRead) % weight

That adds `read` and `nextRead` to the graph as vertices (if they aren't already there) and adds an edge between them. I assume that Scala Graph would produce a connected component with a solitary vertex if such a thing were present in the graph. But because of the way we construct it, it's impossible. Every node in the graph has at least one edge ⇒ every connected component contains an edge.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 240 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-250)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L240)):</sup>
I left a few comments on your refactor and merged it. Thanks!

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 277 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-287)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L277)):</sup>
What doesn't make sense to me is why one of these works but not the other:

```
scala> List(1, 2, 3).map(Math.max(_, 2))
res3: List[Int] = List(2, 2, 3)

scala> List(1, 2, 3).map(Math.abs(_ - 2))
<console>:20: error: missing parameter type for expanded function ((x$1) => x$1.$minus(2))
              List(1, 2, 3).map(Math.abs(_ - 2))
```

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 281 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-275)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L281)):</sup>
Yeah, exactly. I added another line to the comment. Let me know if you think there's state we still haven't captured here.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/343)

ryan-williams:
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 176 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-174)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L176)):</sup>
wow, ok, the plot thickens! can you investigate?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 191 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-194)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L191)):</sup>
cool, mind adding a comment to the `.head` call below to this effect? Any time I see such a call I get nervous.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 236 \[r5\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvVYYYccx2MveqUwUSY)** ([raw file](https://github.com/hammerlab/guacamole/blob/8945fc6e4f15683a2f6a374b6b777f24e47459c2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L236)):</sup>
ah, sry, I changed this so that wiggle room is now positive. I think that's a good change, but this comment should now be updated.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 275 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-287)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L275)):</sup>
hm, interesting! yea also not obvious to me

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/343)

danvk:
Back to you, Ryan!

---

Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 174 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-174)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L174)):</sup>
I filed https://github.com/scala-graph/scala-graph/issues/46 and added an explicit check for the weights. As it turns out, the weight was really `0`, not `1`, because of the imprecision in DELLY's weight calculation that we've discussed elsewhere in this review.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 193 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvQx1F4HsASz1tyWEcW-r2-194)** ([raw file](https://github.com/hammerlab/guacamole/blob/f328faf197011444238d124f5b26668740d11e9e/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L193)):</sup>
Done.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 235 \[r5\]](https://reviewable.io:443/reviews/hammerlab/guacamole/343#-JvVYYYccx2MveqUwUSY)** ([raw file](https://github.com/hammerlab/guacamole/blob/8945fc6e4f15683a2f6a374b6b777f24e47459c2/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L235)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/343)

PR #346, opened 2015-07-30
refactor clique-finding

ryan-williams:
(PR on top of #343)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/346)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala
@@ -183,32 +183,19 @@ object StructuralVariant {
       graph
     }
 
-    case class SVClique(
-      reads: Set[PairedMappedRead],
-      span: GenomeRange,
-      wiggle: Long)
+    case class SVClique(readPairs: mutable.Set[PairedMappedRead],
+                        wiggle: Long,
+                        svStart: Long,
+                        svEnd: Long,
+                        maxNormalInsertSize: Int) {
 
-    def findOneClique(g: PairedReadGraph, maxNormalInsertSize: Int): SVClique = {
-      // Seed the clique with the source of the lowest-weight edge.
-      // Because the graph is constructed by adding edges, all connected components will have at least one edge.
-      val edges = g.edges.toList.sortBy(_.weight)
-      val bestEdge = edges.head
-
-      val initPair = bestEdge.nodes.minBy(_.minPos).value
-      var (pairMin, svStart, svEnd, _) = initPair.startsAndStops
-
-      // this is the "wiggle room": by how much can the SV shrink and still be compatible with all reads in the clique?
-      // It starts negative and then increases towards zero (at which point there's no wiggle room left).
-      var wiggle = initPair.insertSize - maxNormalInsertSize - (svEnd - svStart)
-
-      val clique = mutable.Set(initPair)
-      val graphSize = g.nodes.length
+      def span = GenomeRange(readPairs.head.read.referenceContig, svStart, svEnd)
 
       // If an edge has one node in the clique and one not, return the one that's not.
-      def findNodeConnectedToClique(edge: PairedReadGraph#EdgeT): Option[PairedReadGraph#NodeT] = {
+      def findEligibleNode(edge: PairedReadGraph#EdgeT): Option[PairedReadGraph#NodeT] = {
         val Seq(node1, node2) = edge.nodes.toSeq
-        val node1InClique = clique.contains(node1.value)
-        val node2InClique = clique.contains(node2.value)
+        val node1InClique = readPairs.contains(node1.value)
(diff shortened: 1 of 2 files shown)


danvk:
LGTM -- I'll merge this into my PR. TIL `foldLeft` === `reduce`, at least in the [JavaScript sense][1] of `reduce`.

Mapping and flatMapping over `Option`s still feels weird to me, but I guess that's Scala!

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 186 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/346#-JvVPM1SbqbTtrLMW_UO)** ([raw file](https://github.com/hammerlab/guacamole/blob/a81b4890f0d0c7467b5396f5a627581bd10c3cdb/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L186)):</sup>
This doesn't need to be `mutable`, does it?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 197 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/346#-JvVOPJ_F0E079vmLJae)** ([raw file](https://github.com/hammerlab/guacamole/blob/a81b4890f0d0c7467b5396f5a627581bd10c3cdb/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L197)):</sup>
TIL you don't need to use `this` to reference class members?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 252 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/346#-JvVQ5QrUQKd2Hdeub1m)** ([raw file](https://github.com/hammerlab/guacamole/blob/a81b4890f0d0c7467b5396f5a627581bd10c3cdb/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L252)):</sup>
Interesting that this works in Scala (in JS you'd have to do `clique.nodeIsConnected.bind(clique)`)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/346)

ryan-williams:
yea, scala's `reduce` omits the initial argument, returns the same type as the elems. It has only the "combiner" bits and not the "zeroValue" bits that we looked at in RDD's `combineByKey` hierarchy.

Interestingly, in RDD-land, `reduceByKey` is consistent with this, but `foldByKey` assumes that the `zeroValue` and `output` are of the same type as the elements, whereas `aggregateByKey` is more analogous to `foldLeft` here in that the initial value and output can be an arbitrary type.

---

Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 186 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/346#-JvVPM1SbqbTtrLMW_UO)** ([raw file](https://github.com/hammerlab/guacamole/blob/a81b4890f0d0c7467b5396f5a627581bd10c3cdb/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L186)):</sup>
good call, can you fix?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 197 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/346#-JvVOPJ_F0E079vmLJae)** ([raw file](https://github.com/hammerlab/guacamole/blob/a81b4890f0d0c7467b5396f5a627581bd10c3cdb/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L197)):</sup>
well, they're in scope from the `case class` definition above (would work for regular `class` too)

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala, line 252 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/346#-JvVQ5QrUQKd2Hdeub1m)** ([raw file](https://github.com/hammerlab/guacamole/blob/a81b4890f0d0c7467b5396f5a627581bd10c3cdb/src/main/scala/org/hammerlab/guacamole/commands/StructuralVariantCaller.scala#L252)):</sup>
hm yea, well it's the same as `.filter(n => clique.nodeIsConnected(n))` in scala, whereas it's not in JS

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/346)

PR #349, opened 2015-07-31
Switch to WkUnDiEdge

danvk:
The `k` variant of weighted edges includes them in equality tests. This doesn't make a lot of sense to me (why wouldn't that be the default behavior?) but it's how Scala Graph works.

The operator syntax for WkUnDiEdge is quite awkward:

    Graph(("a"~%#"b")(1))

So I switched to the named function. For the tests, where I explicitly construct many `Graph`s, I added a function to convert a `Graph` built using the natural syntax to the key-weighted variety.

See https://github.com/scala-graph/scala-graph/issues/46

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/349)


ryan-williams:
Lgtm. Weird decisions on the part of Scala graph

PR #350, opened 2015-07-31
Use MDTaggedRead in place of MappedRead in most places

### LINKED ISSUE #297: Remove requirement on MDTag for MappedRead
We may want to analysis mapped reads without the reference information, so we not care if they are MDTagged - should we remove this requirement? cc @ryan-williams @timodonnell

danvk:
I changed almost everything to use the new `MDTaggedRead` class, which inherits from `MappedRead`.

`MappedRead` now has `Option[String]` for its mdtag. The new class has a non-`Option` version.

To load MDTagged reads, you use `ReadSet.mdTaggedReads`.

One thing I don't like here is that `MDTaggedReadSerializer.read` is pure copy/paste. I was able to re-use the `write` method, but not the `read.

Fixes #297

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/350)


danvk:
Most of this change is find/replace `MappedRead`→`MDTaggedRead`. The substantive bits are the new `MDTaggedRead` class, the new methods on `ReadSet` and the serializer.

cc @timodonnell @arahuja

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -185,7 +185,7 @@ object Read extends Logging {
     val r = if (isMapped) {
       Option(record.getStringAttribute("MD")) match {
         case Some(mdTagString) =>
-          val result = MappedRead(
+          val result = new MDTaggedRead(
```
timodonnell: Since we now have the ability to represent mapped reads without MDtags, I think we should return such a read here if there is no mdtag string (on line 213)

timodonnell:
I don't see a strong rationale for making `MDTaggedRead` its own type. `MappedRead` and `PairedRead` lend themselves to having their own types, as I can easily imagine writing functions that only make sense one or the other. `MDtags` are not like that -- we only require them because we currently don't support loading the reference, and they give an easy way to get the reference sequence. When we eventually support loading the reference that's going to mean undoing most of these changes.

I know I mentioned this before and people seemed to disagree, so if the consensus is to go ahead with this, no worries, just wanted to give my $.02.

ryan-williams:
I hear those points, Tim. If Arun and/or Dan agree with you then I'm happy to be overruled; I feel positive about the spirit of this change (haven't had a chance to look over the code yet) but not exceedingly so. A couple thoughts:

* You've implied that we should have reads get their reference bases from a channel other than MDTags. I thought Dan's point somewhat refuting that yesterday made sense; I don't see a way to thread each read's ref seq to it that is cleaner or easier than using MDTags, which are already the relevant piece of ref seq denormalized on to the reads that have them. Would you do a join pass to get ref bases on to reads? What would that [read + ref bases] data structure be, another `*Read` subclass? Or would you broadcast the whole reference to tasks that want some ref bases?
* I'd be ok with just having a class with an `Option[MdTag]`, though if we're ultimately assuming we have MDTags in some/most code paths, then it's nice to have that encoded in the type system, and we've balked at the extra object allocation due to an Option on fields like this in the past.

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
danvk: Done

danvk:
Every code path except `StructuralVariant` wants reads with MDTags on them. I see the point about having `Option[MDTag]` and just a single type, but in practice `Option`s seem to be contagious. My code in `StructuralVariant` benefitted greatly from the addition of `PairedMappedRead`, which did a check for pairs once and then baked the non-optional mate pair into the type system.

I imagine that if we `Option`alized the `MDTag` property, then we'd see a contagion of option-y code throughout the pileup iterator and variant callers.

danvk:
@timodonnell if you'd prefer we change `MDTaggedRead`→`ReadWithReference`, I'd be fine with that.

timodonnell:
Gotcha, @ryan-williams . Two points:

The `option`-phobia was a brief period of insanity for me that I hope has passed :-) I don't think we have to worry about doing that.

I agree we don't want to be threading a ref sequence to each read. I'm thinking we'd add an optional parameter to pileupFlatMap and other functions that need reads + ref sequences. This paramter would give an RDD of the reference sequence (one item in the RDD per task, with a string giving the reference sequence for the loci assigned to that task). If you don't give that parameter, then you'll get an error if your reads don't have md tags. So the types of the paratmers for this function would be: `RDD[MappedRead], RDD[(Long, String)]`.

ryan-williams:
Yea we could have a ReadWithReference whose companion object includes a ctor that takes an MDTag and creates the read with just the ref bases. Then if another way of threading ref bases in came along it'd be easy to add.

Having trouble commenting on surviving lines on mobile but a few questions Dan:

* is it weird that MappedRead and MDTaggedRead both have a ctor param called mdTagString? Should the one be mdTagStringOpt?
* related: are all their ctor args private, since they're not `case class `es? Is that OK?
* are any of these find/replaced instances of MappedRead actually ok to not have an MDTag? Can/Should we use the more permissive superclass in any of the interfaces?

timodonnell:
`ReadWithReference` would be convenient but is inefficient. If we have a reference fasta, we don't want to duplicate the reference sequence for every read, or have to build up a compressed representation like an MDTag. We just need to make sure each task has one copy of the reference for the loci it's looking at. This can be done with a separate parameter to pileupFlatmap.

danvk:
@timodonnell if we went with the `ReadWithReference` trait approach, then the only method on it would be either `referenceBases` or `getReferenceBaseAtLocus`. If `pileupFlatmap` consumed `RDD[MappedRead]` and the FASTA reference, it should be able to construct Reads that implemented that trait efficiently.

timodonnell:
@danvk gotcha, that seems reasonable to me

danvk:
Just to make sure we're all on the same page, the path forward here is:

1. Create a new `trait`, `ReadWithReference`, which extends `MappedRead` and has `getReferenceBaseAtLocus` and `referenceBases` methods.
2. Make `MDTaggedRead` implement that trait.
3. Change most methods which currently return `MDTaggedRead` to return `ReadWithReference`.

ryan-williams:
I was suggesting that there be no MDTaggedRead, just an alternate
ReadWithReference ctor that takes an MDTag.

danvk:
@timodonnell and I chatted about this offline on Friday and he turned me around to his way of thinking.

The way forward will be:

1. (mostly) revert this change
2. Change `MDTag`→`Option[MDTag]` on `MappedRead`. Continue using `MappedRead` everywhere.
3. Change `pileupFlatmap` so that it `throw`s when reads don't have MDTags.
4. Change the callers that care about MDTags (i.e. all of them but `StructuralVariantCaller`) so that they either filter out reads w/o MDTags or complain about them being missing.

In the future, we'll have a separate version of `pileupFlatmap` which takes a secondary `RDD` of reference data, to support MDTag-less inputs.

This makes sense to me because:

1. The `Option` contagion is contained to `pileupFlatmap`.
2. @timodonnell persuaded me that it would be hard, expensive & unnecessary to have a pre-processing step which attaches reference data to every single read. We can just do that seamlessly in `pileupFlatmap`.

ryan-williams:
Interesting. A couple of thoughts:

> We can just do that seamlessly in `pileupFlatmap`.

Just to be clear, this will require a join of an RDD of [Pileups of MDTag-less reads] with an RDD of reference info, which we think is less "hard, expensive & unnecessary" than a join of an RDD of MDTag-less reads with an RDD of reference info (basically creating MDTags on the reads)? We often have 3B pileups (size of the first RDD in the former method) vs. 1B reads (size of the first RDD in the latter method), so that's not obvious to me.

And in that world, we have two `pileupFlatmap` versions, each taking the same read class with `Option[MdTag]`s, but in one case we `throw` if the options are `None` and in the other we assume they're `None` and do a join with reference info as a result?

Finally, I'm not really clear on what you're referring to by saying "`Option` contagion".

timodonnell:
@ryan-williams let's chat when we're all in the office. I seem to be not communicating my point well in text.

PR #353, opened 2015-08-21
PoC: one-pass affine alignment; other CR nits

ryan-williams:
* combine the two matrices into one
* store full paths instead of just last alignments
* treat 0th row and col correctly
* explicitly `Option` "previous state"
* allow alignment to start and end inside the provided reference bases

don't feel that strongly about most of this, but interested to hear thoughts @arahuja

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/353)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
+++ b/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
@@ -1,49 +1,11 @@
 package org.hammerlab.guacamole.alignment
 
-import breeze.linalg.{ DenseMatrix, argmin }
-import org.hammerlab.guacamole.alignment.AlignmentState.AlignmentState
-
-import scala.annotation.tailrec
+import breeze.linalg.DenseMatrix
+import org.hammerlab.guacamole.alignment.AlignmentState.{ AlignmentState, isGapAlignment }
 
 object AffineGapPenaltyAlignment {
 
-  /**
-   * Retraces the optimal alignment path from a matrix of scores and transitions
-   *
-   * @param alignmentStates (sequenceLength + 1) x (referenceLength + 1) matrix of best transition at each step
-   * @param alignmentScores (sequenceLength + 1) x (referenceLength + 1) matrix of alignment score at each step
-   * @param sequenceLength Length of the input sequence to align
-   * @param referenceLength Length of reference sequence to align against
-   * @return An alignment path and score
-   */
-  def findBestPath(alignmentStates: DenseMatrix[AlignmentState],
-                   alignmentScores: DenseMatrix[Double],
-                   sequenceLength: Int,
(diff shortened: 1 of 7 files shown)


arahuja:
@ryan-williams Had to a chance to look this over - looks good overall, but a couple questions.

I don' understand the last commit - what are the reference bases stored for?  Essentially we want something like an MDTag here right?

Also, I'm not sure how to evaluate the additional memory usage of storing the paths - any ideas?  

A test case on the starting/ending anywhere in the reference would be useful.

ryan-williams:
> I don' understand the last commit - what are the reference bases stored for? Essentially we want something like an MDTag here right?

If the best alignment to the provided refbases now starts and ends somewhere in the middle of those refbases, that seems like something that this code needs to communicate back out to its caller, so that was my way of doing that.

Suppose you call it with a 100bp read and 1000bp of reference, and you get back a path that is 100 `AlignmentState`s; where did it start and end? Seems important to know.

This could be communicated via an MDTag, but would we just pad the start and end with deletions? I'm assuming a contract where callers are passing some eligible reference region that has some slack on each side, and the best things for this code to return is 1) where in that region the read aligns best, and 2) what the alignment there is.

lmk if that makes sense.

> Also, I'm not sure how to evaluate the additional memory usage of storing the paths - any ideas?

I think we should just run it on some small/medium/large-sized datasets and see what the UI/Spree tells us?

I was just running into this with the ADAM MDTagging stuff though; we need some better ways to profile these jobs.

If we found one dataset we liked, we could also bisect various {driver,executor}-memory params until we run into problems, in addition to eyeballing the UI stats. Between those two I feel like we could get a reasonable sense of the tradeoffs.

The full-paths-matrix impl will probably show its additional mem usage pretty obviously; 100bp reads will mean 100x the mem usage...?

> A test case on the starting/ending anywhere in the reference would be useful.

Yea, definitely need test cases for this, sorry, I just threw it over the wall as soon as I could so that we could get talking about some of these issues.

Why don't I move this to a hammerlab branch so that we can both contribute to it?

ryan-williams:
I pushed HEAD here to the `affine` branch on this repo, which you can add to.

Seems like we'll have to make a new PR to merge that one in to #342 instead of this one.

You can also send changes to my fork to hit #353 directly.

lmk what you want to do

arahuja:
> This could be communicated via an MDTag, but would we just pad the start and end with deletions? I'm assuming a contract where callers are passing some eligible reference region that has some slack on each side, and the best things for this code to return is 1) where in that region the read aligns best, and 2) what the alignment there is.

Yea, that's the assumtption that we are aligning to an eligible and most likely longer region.  I think start/end and mdtag would suffice but we can add that in after.

I'm going to close this PR and pick it up from the `hammerlab/affine` branch.

ryan-williams:
Ah, start+end+MDTag seems reasonable.

switching to hammerlab/affine sgtm.

PR #355, opened 2015-08-24
Reference handling in pileup processing

arahuja:
This allows reference handling in `pileupFlatMap`

It does overlap with adam/#790 a bit, particularly with MDTag reads so we can rely on that functionality in ADAM mostly but this does offer it directly here as well.

Most importantly, I think this gives us a way to retrieve reference information while processing pileups which will be useful to retrieve variant context and reference statistics.

Right now this broadcasts each chromosome individually, but there is a more generic interface in `ReferenceGenome`.  Also `ReferenceBroadcast` can more finely chop the reference and broadcast those pieces.  This would result in only small fragments being shipped to the executors that need them.

There are 2 commits here:
the first gives the interface for loading the reference and sets the pileup base based on it
the second mdtags the reads as they are read in.

The second is still necessary for full functionality to retrieve the deleted bases, but this can be removed if we provide each pileup a longer stretch of the reference, but I've left that off for now.

Also, there were a few improvements we could make - i.e. we don't need to recompute the full reference sequence from the mdtag as we can access the deleted bases directly.

@ryan-williams interested to hear what you think

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/355)


arahuja:
Ah, I forgot to add some sample run information.  A WGS `somatic-standard` call takes ~12min.  With the pileup processing with reference information this take ~14 min and with mdtagging on read in on top of that takes ~16min

ryan-williams:
Looks pretty good to me, a few comments in there. Interested in continuing our discussion about what gets broadcast and how we can be more efficient about it.

In principle we should be able to send each task only the reference bases for its window; I think the way to do that will be to , after we've assigned loci to partitions, slice up the reference into chunks for each partition, and then `.zipPartitions`.

I'm skeptical that we can do the most efficient thing using broadcasts.

---

Review status: 0 of 22 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 57 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWQ1sEhfEPA0_y-R-_)** ([raw file](https://github.com/hammerlab/guacamole/blob/efb79c588d7b3943a28c460d21a7b3d3f2d4a1ff/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L57)):</sup>
echoing the conversation on ADAM#790: any interest in supporting `.2bit` too? I don't care that much here, j/w.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 74 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWQIpN2DMNlV8MiSgD)** ([raw file](https://github.com/hammerlab/guacamole/blob/efb79c588d7b3943a28c460d21a7b3d3f2d4a1ff/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L74)):</sup>
I remain somewhat partial to `Opt`-suffixing `Option`al `val`s like this, but don't feel that strongly.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/VariantSupport.scala, line 78 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWQcr5RHO9ErtqzvAy)** ([raw file](https://github.com/hammerlab/guacamole/blob/efb79c588d7b3943a28c460d21a7b3d3f2d4a1ff/src/main/scala/org/hammerlab/guacamole/commands/VariantSupport.scala#L78)):</sup>
could drop this and some of the others, yea?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/reference/ReferenceBroadcast.scala, line 46 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWUKQz322fEHtPcC-q)** ([raw file](https://github.com/hammerlab/guacamole/blob/efb79c588d7b3943a28c460d21a7b3d3f2d4a1ff/src/main/scala/org/hammerlab/guacamole/reference/ReferenceBroadcast.scala#L46)):</sup>
so, this is going to send all sequences to all executors, right?

---

<sup>**[src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala, line 65 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWXsfACxQBvocFOF9a)** ([raw file](https://github.com/hammerlab/guacamole/blob/02e9a7858b348bd47b6d2b31d362944d16cf8604/src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala#L65)):</sup>
what is the point of this? I don't see it being used anywhere of import

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/355)

arahuja:
Yea, I've been thinking about this as well, and I'm very interested to see the difference.  Slicing up the reference may be more work for the driver then shipping a few GB (I think it's <1GB compressed even, but need to check)

---

Review status: 0 of 22 files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 0 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWQ1sEhfEPA0_y-R-_)** ([raw file](https://github.com/hammerlab/guacamole/blob/efb79c588d7b3943a28c460d21a7b3d3f2d4a1ff/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L0)):</sup>
Yea, seems like it would be straightforward but I've never worked with the 2bit format.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala, line 0 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWQIpN2DMNlV8MiSgD)** ([raw file](https://github.com/hammerlab/guacamole/blob/efb79c588d7b3943a28c460d21a7b3d3f2d4a1ff/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala#L0)):</sup>
Sure I can add that

---

<sup>**[src/main/scala/org/hammerlab/guacamole/commands/VariantSupport.scala, line 78 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWQcr5RHO9ErtqzvAy)** ([raw file](https://github.com/hammerlab/guacamole/blob/efb79c588d7b3943a28c460d21a7b3d3f2d4a1ff/src/main/scala/org/hammerlab/guacamole/commands/VariantSupport.scala#L78)):</sup>
Was getting an odd error on "Can't find impl" when compiling which this resolved

---

<sup>**[src/main/scala/org/hammerlab/guacamole/reference/ReferenceBroadcast.scala, line 0 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWUKQz322fEHtPcC-q)** ([raw file](https://github.com/hammerlab/guacamole/blob/efb79c588d7b3943a28c460d21a7b3d3f2d4a1ff/src/main/scala/org/hammerlab/guacamole/reference/ReferenceBroadcast.scala#L0)):</sup>
Yes, but I don't think they will be deserialized unless the task actually needs it.  Otherwise, I think it is stored in the local block manager, but I have to dig into this more

---

<sup>**[src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala, line 0 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWXsfACxQBvocFOF9a)** ([raw file](https://github.com/hammerlab/guacamole/blob/02e9a7858b348bd47b6d2b31d362944d16cf8604/src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala#L0)):</sup>
Do you mean removing the `token = ` 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/355)

ryan-williams:
Cool, well we can merge this and bake off later. lgtm, I'll let you dig into the q's below now if you care to, otherwise merge away!

---

Review status: 0 of 22 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala, line 0 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWXsfACxQBvocFOF9a)** ([raw file](https://github.com/hammerlab/guacamole/blob/02e9a7858b348bd47b6d2b31d362944d16cf8604/src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala#L0)):</sup>
ah, sorry, I really meant what's the point of `token` in the first place?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/355)

arahuja:
Review status: 0 of 22 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala, line 65 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/355#-JxWXsfACxQBvocFOF9a)** ([raw file](https://github.com/hammerlab/guacamole/blob/02e9a7858b348bd47b6d2b31d362944d16cf8604/src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala#L65)):</sup>
I think it's used somewhere to differentiate two sets as it is added on to every read.  Similarly we could use `sampleName` to compare, but I think this was preferred for some reason that I'm not sure of.

If that's not true any more we can drop token.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/355)

PR #358, opened 2015-08-25
Updates to simple alignment to single-pass and matrix pruning

arahuja:
@ryan-williams I couldn't figure out the git surgery to split the first commit out so ended up just implementing it with only saving see 2 rows of the matrix.  There is only one new commit here compared to #353, aa6a356

This is a PR against `simple-alignment` but then we can merge all of these in.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/358)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
+++ b/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
@@ -1,49 +1,11 @@
 package org.hammerlab.guacamole.alignment
 
-import breeze.linalg.{ DenseMatrix, argmin }
-import org.hammerlab.guacamole.alignment.AlignmentState.AlignmentState
-
-import scala.annotation.tailrec
+import breeze.linalg.DenseMatrix
+import org.hammerlab.guacamole.alignment.AlignmentState.{ AlignmentState, isGapAlignment }
 
 object AffineGapPenaltyAlignment {
 
-  /**
-   * Retraces the optimal alignment path from a matrix of scores and transitions
-   *
-   * @param alignmentStates (sequenceLength + 1) x (referenceLength + 1) matrix of best transition at each step
-   * @param alignmentScores (sequenceLength + 1) x (referenceLength + 1) matrix of alignment score at each step
-   * @param sequenceLength Length of the input sequence to align
-   * @param referenceLength Length of reference sequence to align against
-   * @return An alignment path and score
-   */
-  def findBestPath(alignmentStates: DenseMatrix[AlignmentState],
-                   alignmentScores: DenseMatrix[Double],
-                   sequenceLength: Int,
-                   referenceLength: Int): ReadAlignment = {
-
-    @tailrec
-    def recFindBestPath(sequenceIdx: Int,
-                        referenceIdx: Int,
-                        currentPath: List[AlignmentState]): List[AlignmentState] = {
-
-      val lastMove = alignmentStates(sequenceIdx, referenceIdx)
-      if (sequenceIdx == 1 && referenceIdx == 1) {
-        lastMove :: currentPath
-      } else {
-        lastMove match {
-          case AlignmentState.Match | AlignmentState.Mismatch =>
-            recFindBestPath(sequenceIdx - 1, referenceIdx - 1, lastMove :: currentPath)
-          case AlignmentState.Insertion =>
-            recFindBestPath(sequenceIdx - 1, referenceIdx, lastMove :: currentPath)
-          case AlignmentState.Deletion =>
-            recFindBestPath(sequenceIdx, referenceIdx - 1, lastMove :: currentPath)
-        }
-      }
-    }
-
-    val alignments = recFindBestPath(sequenceLength, referenceLength, List.empty)
-    ReadAlignment(alignments, argmin(alignmentScores.t(::, sequenceLength)))
-  }
+  type Path = (Seq[AlignmentState], Double)
 
   /**
    * Produces an alignment of an input sequence against a reference sequence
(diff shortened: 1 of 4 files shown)


ryan-williams:
This is really cool, thanks for doing this. The two-row thing is cool; I was imagining a different impl: a sort of creeping advance from the (0,0) corner, using a pqueue, but that would be more complicated and is probably actually a slower big-O now that I think about it; I guess if you really thought most alignments would go right down the diagonal it could be faster in practice, because you'd only compute ~N entries of the matrix instead of N^2, but whatevs, I think this is better.

This looks good to me. Feel free to merge or follow up on the couple points below.

---

Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala, line 76 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/358#-JxY4Q3taGRzjBro-73H)** ([raw file](https://github.com/hammerlab/guacamole/blob/aa6a3567eae52e4b827e952ed38f9631f43cf569/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala#L76)):</sup>
just noticed this logic might a little funky; if we transition from insertion to deletion, should we add a close-gap penalty? we're doing that here IIUC. I could see the right thing being to do it either way, j/w

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala, line 107 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/358#-JxY4nR2H-ph6yJW-dWH)** ([raw file](https://github.com/hammerlab/guacamole/blob/aa6a3567eae52e4b827e952ed38f9631f43cf569/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala#L107)):</sup>
I guess we don't need the `sI >= 0` check anymore, but maybe it's simpler to leave it in tbh

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala, line 129 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/358#-JxY5P9Yigbr2qyuvTIX)** ([raw file](https://github.com/hammerlab/guacamole/blob/aa6a3567eae52e4b827e952ed38f9631f43cf569/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala#L129)):</sup>
ah, meant to note this in #353; I think building the path this way is O(n^2), if we care.

could build it in reverse via `prevPath :: nextState` and then `.reverse` it at the end.

up to you, just wanted to note what a questionable thing i did on my first pass :)

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/ReadAlignment.scala, line 20 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/358#-JxY5ssZKRe6drjA6Hgu)** ([raw file](https://github.com/hammerlab/guacamole/blob/aa6a3567eae52e4b827e952ed38f9631f43cf569/src/main/scala/org/hammerlab/guacamole/alignment/ReadAlignment.scala#L20)):</sup>
another thing I thought about while doint #353 was that `Path` and `ReadAlignment` are really close to being the same thing.

something kept me from unifying them, and I don't really think it's worth it, just wanted to note that

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/358)

arahuja:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala, line 76 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/358#-JxY4Q3taGRzjBro-73H)** ([raw file](https://github.com/hammerlab/guacamole/blob/aa6a3567eae52e4b827e952ed38f9631f43cf569/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala#L76)):</sup>
Yea I think we should, otherwise a lot mismatches can be replaced with an insertion/deletion combination

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala, line 129 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/358#-JxY5P9Yigbr2qyuvTIX)** ([raw file](https://github.com/hammerlab/guacamole/blob/aa6a3567eae52e4b827e952ed38f9631f43cf569/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala#L129)):</sup>
Cool, I was wondering that too - but I looked up the collections and this should be amortized constant on a vector/array? Is that wrong?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/358)

ryan-williams:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala, line 129 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/358#-JxY5P9Yigbr2qyuvTIX)** ([raw file](https://github.com/hammerlab/guacamole/blob/aa6a3567eae52e4b827e952ed38f9631f43cf569/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala#L129)):</sup>
ah, OK, interesting point. I'd believe that about vector/array, but we're initializing all of these to `Nil`, which IIUC means this will be `List`s, which IIRC have linear-time append. I'll dig more to confirm this but lmk if that makes sense or not

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/358)

ryan-williams:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala, line 129 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/358#-JxY5P9Yigbr2qyuvTIX)** ([raw file](https://github.com/hammerlab/guacamole/blob/aa6a3567eae52e4b827e952ed38f9631f43cf569/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala#L129)):</sup>
So, this strongly suggests that we should append to `Vector`s, and that `List` and `Array` append are quadratic https://gist.github.com/ryan-williams/d8a6e6afa716ff08f187

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/358)

arahuja:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala, line 129 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/358#-JxY5P9Yigbr2qyuvTIX)** ([raw file](https://github.com/hammerlab/guacamole/blob/aa6a3567eae52e4b827e952ed38f9631f43cf569/src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala#L129)):</sup>
Yeesh, will fix then

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/358)

PR #360, opened 2015-09-14
Kmer Graph processing for assembly caller

arahuja:
Along with #342 this is the second large piece for #124.  This builds a DeBrujin graph from a set of reads and finds path between any two nodes.  The graph can be reduced to only branching path and all non-branching paths are merged using `mergeNodes()`

There is a lot that may change as we actually use this, but wanted to get this larger piece in.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/360)


ryan-williams:
Reviewed 4 of 7 files at r1.
Review status: 4 of 7 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/test/resources/assemble-reads-set3-chr2-73613071.sam, line 3 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzC1BlvjcE4mYPsqxnq)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/test/resources/assemble-reads-set3-chr2-73613071.sam#L3)):</sup>
seems like only this contig is used in the reads, maybe remove the others?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

arahuja:
Review status: 4 of 7 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/test/resources/assemble-reads-set3-chr2-73613071.sam, line 3 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzC1BlvjcE4mYPsqxnq)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/test/resources/assemble-reads-set3-chr2-73613071.sam#L3)):</sup>
Yea, good idea!

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

ryan-williams:
This looks cool, a couple of questions below. Also: `s/DeBrujin/DeBruijn/g` :)

---

Review status: 4 of 7 files reviewed at latest revision, 10 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 33 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzM7xec2A_DTiX3rbCW)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L33)):</sup>
Can you explain the `(Int, Kmer)` more in this comment?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 83 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMVpbKAVZ9zzJITGR2)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L83)):</sup>
In general, I prefer doing computations like this before the object (DBG in this case) is instantiated, so that it can be immutable.

A ctor-like helper in the DBG companion object could do this computation and then instantiate a DBG with the merged kmer Map.

* Do we really think we'll want DBG's where we don't merge the nodes? The totally conservative/lossless merge process you've implemented here seems like something we'd basically always want to apply.
* Do we want some flexibility in the merging, i.e. sometimes merge where we throw away side branches with few occurrences and merge over them? I imagine us instantiating DBG's at different levels of pruning less-supported side-branches, and that to me feels more like it'd make sense to just instantiate the DBG based on those parameters and have it compute the induced merged graph at instantiation time.
* There could also be a method on DBG's to apply a further, stricter merge, that would return a new immutable DBG object, etc.

lmk your thoughts on this.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 86 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMFaRcwZFUUKFrK9b7)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L86)):</sup>
`allNodes.nonEmpty`

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 104 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMFiPVzNM4VXCDqqWp)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L104)):</sup>
IJ doesn't like shadowing `node` here

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 112 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMGJH0U9QFinQVtaXP)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L112)):</sup>
this assumes we've not seen this node before, right? is that safe?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 121 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMGarrnjhKFMvSRuFD)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L121)):</sup>
I can't parse "Searches from a node to merge node"

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 148 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMI1cHTPJsnjW9m5R5)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L148)):</sup>
do you not need to check for loops in the `prevFunc` call as well?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 192 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMJRWB5Zk8VPVWy5_k)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L192)):</sup>
won't dropping some prefix of a merged node result in a "kmer" that is not found in `kmerCounts` (cf. call to `kmerScore` below)? Seems like there's not a test of this scenario, so I'll let you weigh in before I dig further.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 248 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMQo1hTo00JfI7o91G)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L248)):</sup>
`parents(_).isEmpty`

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

arahuja:
Review status: 4 of 7 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 33 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzM7xec2A_DTiX3rbCW)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L33)):</sup>
Yes, I'll add that.  If it was unclear, it's a map from the substring to the string it came from and the position in that string.  I've sure there is a more memory-efficient way to save this as well

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 83 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMVpbKAVZ9zzJITGR2)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L83)):</sup>
> In general, I prefer doing computations like this before the object (DBG in this case) is instantiated, so that it can be immutable.

Yes, agreed.  This is somewhat an artifact of the development/debugging process where first I implemented the standard way and then added the merging and wanted to switch back and forth for testing

> Do we really think we'll want DBG's where we don't merge the nodes?
I don't think so, but I wanted he option, again mainly for debugging now.

> Do we want some flexibility in the merging, i.e. sometimes merge where we throw away side branches with few occurrences and merge over them? 
Yes I think so

> There could also be a method on DBG's to apply a further, stricter merge, that would return a new immutable DBG object
Could you explain this one a bit?  I'm not sure I know what I look like

In summary, to me this is a yes to the intent but a bit of let's keep things simple until we enumerate more uses/harden this a bit?

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 86 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMFaRcwZFUUKFrK9b7)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L86)):</sup>
Yup, still need to add those intellj pointers

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 104 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMFiPVzNM4VXCDqqWp)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L104)):</sup>
will rename

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 112 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMGJH0U9QFinQVtaXP)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L112)):</sup>
Oh good catch, if we have a very large region we may rebuild the same merged sequence twice

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 148 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMI1cHTPJsnjW9m5R5)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L148)):</sup>
No, because we are pulling elements off of next.  Should pull off the next non-loop element and check that the last one had a single parent.  If the last one had itself as a parent we don't want to merge forward (although we could).

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 192 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMJRWB5Zk8VPVWy5_k)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L192)):</sup>
Yea you are right - need to think about this more, possible we can just double count them in `kmerCounts` then...

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

arahuja:
Thanks for the review! Working on the minor fixes now and some require some more thought

---

Review status: 4 of 7 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

ryan-williams:
Review status: 4 of 7 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 83 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMVpbKAVZ9zzJITGR2)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L83)):</sup>
I guess the thing I was imagining would be the same as calling `.pruneKmers(n).mergeNodes`, but in a world where each of those methods returned a `DBG`. This also removes the necessity of  e.g. `removeKmer` method.

I think in this world the DBG takes a (also immutable) `mergeIndex` as a ctor argument, and the companion object exposes a ctor that doesn't require that `mergeIndex`, builds one itself, and instantiates a `DBG` with it.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 86 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMFaRcwZFUUKFrK9b7)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L86)):</sup>
weirdly, I don't think I ever added them, they were just enabled for me. Looks like this is where you can enable/disable them:



---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala, line 192 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMJRWB5Zk8VPVWy5_k)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBrujinGraph.scala#L192)):</sup>
OK, I'll think about it as well; it feels like we're about to reinvent suffix trees :)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

hammer:
@arahuja are you waiting to merge this branch for a reason?

ryan-williams:
I paged out some state on this but I think there was a correctness concern around [here](https://reviewable.io/reviews/hammerlab/guacamole/360#-JzMJRWB5Zk8VPVWy5_k) that we needed to come back to


---

Review status: 4 of 7 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

arahuja:
@ryan sorry for the delay, finally revisted this and updated based on your comments.  For now I removed the path scoring piece all together to move this along.  If you have time and want to look at fresh that would be very helpful!

ryan-williams:
np @arahuja , this is a great overhaul, thanks. I've left some comments below.

I'm also still interested in going immutable, but that can be an open-ended follow-on aspirational task


---

Review status: 4 of 7 files reviewed at latest revision, 21 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala, line 33 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzM7xec2A_DTiX3rbCW)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala#L33)):</sup>
Some light tweaks to this:

Map from k-mers to the sequence they were merged in to.
The value is the merged sequence and the index in that sequence at which the "key" k-mer occurs.


---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala, line 112 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMGJH0U9QFinQVtaXP)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala#L112)):</sup>
Second-guessing myself now: for the same merged sequence to happen twice, its constituent k-mers would have to have appeared at least twice each, meaning they wouldn't have been merged, no? Anyway I like playing it safe here, just curious as I think through this again.


---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala, line 10 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LNynTrP1E6QsS3udC)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala#L10)):</sup>
redundant with the one defined in the companion object?


---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala, line 35 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LNqImQ7N8blX-7Elk)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala#L35)):</sup>
This should probably be a `mutable.Map[Kmer, (Int, Sequence)]`. Actually making it `mutable.Map[Kmer, (Sequence, Int)]` might be clearer / make the comment above easier to write/interpret.


---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala, line 56 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1L90FQZMriudDp8NWJ)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala#L56)):</sup>
nit: I think this should technically be `Kmer => SubKmer`?


---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala, line 210 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LGSxJ9NW8eAtHRoIR)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala#L210)):</sup>
might as well cache `mergeIndex.get(sink)` in a `val` outside this loop?


---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala, line 298 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LI__7S2buL0V-gbov)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala#L298)):</sup>
unused?


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 93 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LQAWVfCr6QmE6SpdE)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L93)):</sup>
might as well test the children here, for consistency?


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 113 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LQRRHeCJ6slgUIpsh)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L113)):</sup>
`s/merageable/mergeable/g`


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 116 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LQoZWhUGM4G_LrBoC)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L116)):</sup>
maybe move the `.reverse` into the impl of `mergeForward`?


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 134 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LSLdbrA5Eb8KNZyEM)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L134)):</sup>
thanks for all these test cases; want to make this "with bubble at end" here and later, for consistency with other cases?


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 206 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LSmBn6nQm9jf0iPJ2)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L206)):</sup>
 :)


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 239 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LT8-aFPl94QvARIxM)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L239)):</sup>
ideally you could just assert equality here once, on a full `Map` showing all that you expect, along with each kmer's counts; if it's not obvious that scalatest does that easily/sanely then don't worry about it though.


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 276 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LV6gIWxkkMYvTAdMa)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L276)):</sup>
might as well reverse the paths in the impl, no?


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 304 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LWRLEIoyMRQPtFoTX)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L304)):</sup>
remove?


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 307 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LWV4VR9OTkITh8C9E)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L307)):</sup>
`_.sequence`, if you want


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 318 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LWZtB7zma0UAhlH8z)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L318)):</sup>
great test case!

does the file have a snp somewhere where you can test that >=2 paths are correctly identified?


---

<sup>**[src/test/scala/org/hammerlab/guacamole/commands/VAFHistogramSuite.scala, line 11 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LWgb9auSfghiUX7-X)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/commands/VAFHistogramSuite.scala#L11)):</sup>
is this change in this PR? or did it slip in from somewhere else?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

arahuja:
Review status: 4 of 7 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala, line 112 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-JzMGJH0U9QFinQVtaXP)** ([raw file](https://github.com/hammerlab/guacamole/blob/b158a0686fda4fd3005b607d78ac0228ffd43ad1/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala#L112)):</sup>
Yea, I guess this makes a case for the more immutable graph.  One corner case of two of the same merged paths emerging is if the graph contained different kmerSizes ever, which is implied here, but not enforced.

So I'll leave this for now.


---

<sup>**[src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala, line 10 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LNynTrP1E6QsS3udC)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala#L10)):</sup>
It is, but a bit clunky to refer to, I changed that definition to `type Kmer = DeBruijnGraph#Kmer`


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 116 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LQoZWhUGM4G_LrBoC)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L116)):</sup>
Yea, good idea


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 239 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LT8-aFPl94QvARIxM)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L239)):</sup>
Yea, that seems reasonable


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 318 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LWZtB7zma0UAhlH8z)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L318)):</sup>
Not sure, but I'm going add a more complete caller soon so I will add test cases of that variety


---

<sup>**[src/test/scala/org/hammerlab/guacamole/commands/VAFHistogramSuite.scala, line 11 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LWgb9auSfghiUX7-X)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/commands/VAFHistogramSuite.scala#L11)):</sup>
Yea it actually was as I pulled this out of a full caller I was working on.  Not really related to the PR, but simple thing I wanted to get out of the way.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

ryan-williams:
looking good, a couple nits and one previous comment that was seemingly missed, sorry this has dragged on


---

Review status: 4 of 7 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 189 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1dRkGacTv_lfUFPAKV)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L189)):</sup>
`s/Merageable/Mergeable/g`


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 206 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LSmBn6nQm9jf0iPJ2)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L206)):</sup>
sorry if I wasn't clear here, but this test case appears to just be a copy/paste of the "find forward unique path; with bubble in middle" case above, though it should have some changes, i.e. merging backward instead of forward and starting from the end rather than the beginning, unless I'm misreading.


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 224 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1dRwv7r8GNEko1Cxl0)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L224)):</sup>
nit: unnecessary newline?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

arahuja:
Review status: 4 of 7 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 189 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1dRkGacTv_lfUFPAKV)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L189)):</sup>
This also seems updated in the latest commit?


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 206 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LSmBn6nQm9jf0iPJ2)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L206)):</sup>
Hm I think I did update this though? I can't find it in reviewable, but the latest commit updates this? 
https://github.com/hammerlab/guacamole/pull/360/files#diff-49ad55aa4ed9dc2f36a9d2fe54bb7420R190


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

ryan-williams:
ok, LGTM!


---

Review status: 4 of 7 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 189 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1dRkGacTv_lfUFPAKV)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L189)):</sup>
yea you're right, sry


---

<sup>**[src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala, line 206 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/360#-K1LSmBn6nQm9jf0iPJ2)** ([raw file](https://github.com/hammerlab/guacamole/blob/900e7870f877979f050ca2f1f586ee6750abc0bd/src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala#L206)):</sup>
Weird, yea I guess reviewable wasn't showing me the most recent commit, I'll have to keep an eye on that going forward, thx


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/360)

PR #362, opened 2015-10-13
Assorted infra tweaks

timodonnell:
Some small tweaks I have accumulated in another branch and wanted to get into master:

* Don't emit alt=N variants in somatic standard
* Base complementing in `Bases`
* `LociSet.parse` accepts strings like `chr1,chr2` instead of requiring start and ends
* Slightly better progress message in Distributed loci partitioning

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/LociSetSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/LociSetSuite.scala
@@ -18,8 +18,6 @@
 
 package org.hammerlab.guacamole
 
-import org.apache.commons.io.FileUtils
-import org.apache.hadoop.fs.FileUtil
 import org.apache.spark.rdd.RDD
 import org.hammerlab.guacamole.util.{ TestUtil, GuacFunSuite }
 import org.hammerlab.guacamole.reads.Read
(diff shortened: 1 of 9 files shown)


src/test/scala/org/hammerlab/guacamole/LociSetSuite.scala
```diff
@@ -162,6 +160,13 @@ class LociSetSuite extends GuacFunSuite with Matchers {
     set1.union(set2).toString should equal("chr1:40-43")
   }
 
+  sparkTest("loci set parsing with contig lengths") {
+    LociSet.parse(
+      "chr1,chr2,17,chr2:3-5,chr20:10-20",
+      Some(Map("chr1" -> 10, "chr2" -> 20, "17" -> 12, "chr20" -> 5000))).toString should equal(
+        "17:0-12,chr1:0-10,chr2:0-20,chr20:10-20")
```
ryan-williams: I assume I'm misunderstanding what's happening here, but should the output not be `chr2:3-5` instead of `chr2:0-20`?

ryan-williams:
lgtm, one question

src/test/scala/org/hammerlab/guacamole/LociSetSuite.scala
timodonnell: The first `chr2` in the list is just a bare chromosome, which means "include the whole chromosome". The second `chr2` in the list is thereofore irrelevant -- we're taking the union of `chr2:0-20` and `chr2:3-5` and get back `chr2:0-20`. Does that make sense?

ryan-williams: ah yea, makes sense

PR #363, opened 2015-10-31
Upgrade to ADAM 0.18

arahuja:
This fixes issues with MdTag parsing in previous versions of ADAM for RNA data.
- Some of the `PileupSuite` tests had to be updated to account for this
   - One had an incorrect CIGAR string
   - One was testing `P` operators, which was not supported in ADAM.  I removed this test for now, but added in the functionality in ADAM https://github.com/bigdatagenomics/adam/pull/875

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/363)


timodonnell:
Thanks for taking care of this long-standing annoyance @arahuja . We should add back the failing test once your ADAM fix goes in (perhaps comment it out here?) LGTM

arahuja:
@timodonnell I'll open an issue with the relavant commit with the test so we can add it back. I do want to investigate first the usage of CIGAR `P`.  We also have an odd comment on in it in `PileupElement` https://github.com/hammerlab/guacamole/blob/master/src/main/scala/org/hammerlab/guacamole/pileup/PileupElement.scala#L132-L133

PR #364, opened 2015-10-31
Default to spark parallelism setting

### LINKED ISSUE #344: the --parallelism should default to spark.default.parallelism
Or perhaps we should get rid of this option and just use  `spark.default.parallelism` instead

arahuja:
Closes #344

This sets the spark setting as the default.  Though it may be confusing to have both though, so I'd be ok with removing this option as well?

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/364)


timodonnell:
LGTM, I think the way you did it here is a reasonable way to go, since it lets us set the two options independently if the need arises

PR #367, opened 2015-11-26
Load local bams efficiently

### LINKED ISSUE #366: make it possible to read a subset of a BAM efficiently (i.e. using the index) when running locally
This would be helpful for local testing and probably some production uses too

timodonnell:
This PR adds support for loading BAMs directly using samtools (htsjdk) in addition to the HadoopBAM implementation we have always supported. Closes #366.

Loading BAMs using samtools has two advantages over HadoopBAM: we can perform some filtering immediately as the reads are loaded, and we can use the BAM index when a subset of loci are being analyzed. On a cluster with data stored in HDFS, HadoopBAM will be faster since it will load in parallel. When the data is stored on a local file system, samtools can be much faster.

We've added an argument "--bam-reader-api" which specifies what api to use: "samtools", "hadoopbam", or "best". The default is "best", which will use samtools when the file is local and otherwise hadoop bam.

This PR required a surprising amount of refactoring, since we needed to be able to parse the loci specified on the commandline before we know what contigs are available (which requires reading the bam).

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/367)


src/main/scala/org/hammerlab/guacamole/reads/PairedRead.scala
```diff
@@ -18,29 +18,11 @@ case class PairedRead[+T <: Read](read: T,
...
-      token,
-      requireMDTagsOnMappedReads,
-      referenceGenome
-    )
-
-    val mateAlignment = MateAlignmentProperties(record)
-    PairedRead(read, isFirstInPair = record.getFirstOfPairFlag, mateAlignment)
-  }
-}
-
```
arahuja: I'd be in favor of leaving this here. I think it more sense to split more of these out from `Read` then adding more back in there.

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -85,22 +83,72 @@ object Read extends Logging {
...
+       * This is implemented as a static function instead of a method in InputFilters because the overlapsLoci
+       * attribute cannot be serialized.
+       */
+      var result = reads
+      if (filters.overlapsLoci.nonEmpty) {
+        val loci = filters.overlapsLoci.get.result(contigLengths(sequenceDictionary))
+        val broadcastLoci = reads.sparkContext.broadcast(loci)
+        result = result.filter(
+          read => read.isMapped && read.asMappedRead.get.overlapsLociSet(broadcastLoci.value))
+      }
```
arahuja: Is there a reason to the the `asMappedRead` and `get` here as opposed to matching on the class?

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -173,9 +220,6 @@ object Read extends Logging {
                     referenceGenome: Option[ReferenceGenome]): Read = {
 
     val isMapped = (
-      // NOTE(ryan): this flag should maybe be the main determinant of the mapped-ness of this SAM record. SAM spec
-      // (http://samtools.github.io/hts-specs/SAMv1.pdf) says: "Bit 0x4 is the only reliable place to tell whether the
-      // read is unmapped."
```
arahuja: Is this comment not valid anymore?

src/main/scala/org/hammerlab/guacamole/reads/PairedRead.scala
timodonnell: I see your point and thought about that too, but my thought was that it was nice to have all the SAM/BAM/ADAM-specific stuff in `Read` so the remaining classes are independent of where the data came from

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
timodonnell: Yeah, it's awkward because we can have plain `MappedRead` instances but also `PairedRead[MappedRead]`, so can't easily match on the type. The `asMappedRead` method deals with this issue.

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
timodonnell: The comment is correct, but to be extremely nit picky it's not directly relevant to the code. The SAM spec is saying that if that flag says the read is unmapped, we can't assume anything about the values like alignment start, cigar, etc. -- they might well have non-empty values, and we shouldn't use those values to decide that the read is actually mapped even though that flag says it's unmapped. If the flag says the read is mapped, however, it's totally reasonable for us to also require that those other values are meaningful / non-empty, which is what we do here. (Arguably we could throw an error if not, instead of calling the read unmapped.)

timodonnell:
Thanks for the review, @arahuja . I had some clarifications on a few things; once I hear back on those I'll update the PR with fixes.

src/main/scala/org/hammerlab/guacamole/reads/PairedRead.scala
arahuja: What do you mean for "remaining classes"? The other `Read` classes? Right now `Read.scala` is a monster so I think, keeping the constructors in their respective files makes sense?

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
arahuja: :+1:

src/main/scala/org/hammerlab/guacamole/reads/PairedRead.scala
timodonnell: Yeah, I'm thinking of the `MappedRead`, `PairedRead`, `UnmappedRead`, and `PairedMappedRead` classes (the latter is only used I think in the SV caller, and is kind of different from the others, possibly needs to be reconsidered if necessary). Currently all of these classes except `PairedRead` have no BAM/SAM/ADAM-specific code in them; this change moves that code out of `PairedRead`  to make it consistent with the others. I agree that `Read.scala` is a monster and we may want to refactor that, perhaps to split the BAM/SAM/ADAM low-level stuff from the high-level `Read` trait, but I think that's better done in a separate PR.

timodonnell:
Updated this PR to rebase onto current master and address feedback:
- Changed LociSet.Builder end parameter to be Option[Long]
- Added `—recompute-md-tags` argument to all callers
- Make it an error to try to load non-local files with samtools

Issue #368, opened 2015-11-30
Fix reference loading for masked regions


arahuja:
When loading the FASTA file, the masked regions of the reference are lower-case with different byte values. We can either ignore the masking or leave this encoding.

arahuja:
Fixed by #369

PR #369, opened 2015-12-03
Fix loading of reference by unmasking bases

arahuja:
Fix for #368

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/369)


timodonnell:
LGTM!

PR #371, opened 2016-01-19
Fix germline-standard error with no alleles

arahuja:
This fixes the empty maxBy error in germline-standard.

This error was caused after we added the allele file in `Likelihood.scala` as `pileup.distinctAlleles.filter(allele => allele.altBases.forall((Bases.isStandardBase _)))`

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/371)


timodonnell:
:+1:

PR #372, opened 2016-01-25
Germline variant caller that does a local reassembly

arahuja:
The basic summary of this caller is:
- Find areas where > some % of the reads show a variant
- Rexamine those areas with N base window around it
- Place the reads in the N-base window into a DeBruijn graph and find paths between the start and end of the reference sequence that covers that region
- Align those paths to the reference sequence to discover variants

This code update `DeBruijnGraph` code to backtrack on dead-end paths as well as some helper functions to `ReadAlignment`

Right now there is a single test based on the Illumina NA12878 platinum extract that @timodonnell created.

There are more updates to come (in future PRs) but this is the first full working version.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/372)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
+++ b/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
@@ -100,7 +100,7 @@ class DeBruijnGraph(val kmerSize: Int,
         })
 
         // Get full node from non-branch path
-        val mergedNode = DeBruijnGraph.mergeKmers(fullMergeablePath)
+        val mergedNode = DeBruijnGraph.mergeKmers(fullMergeablePath, kmerSize)
 
         // Save each node that was merged
         fullMergeablePath.zipWithIndex.foreach({
@@ -293,9 +293,9 @@ object DeBruijnGraph {
     graph
   }
 
-  def mergeKmers(kmers: Seq[Kmer]): Sequence = {
-    val head = kmers.headOption.map(h => h.take(h.length - 1)).getOrElse(Seq.empty)
-    val rest = kmers.map(_.last)
+  def mergeKmers(kmers: Seq[Kmer], kmerSize: Int): Sequence = {
+    val head = kmers.headOption.getOrElse(Seq.empty)
+    val rest = kmers.tail.flatMap(kmer => kmer.takeRight(kmer.length - kmerSize + 1))
     head ++ rest
   }
 
--- a/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
+++ b/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
@@ -180,8 +180,8 @@ class DeBruijnGraph(val kmerSize: Int,
                        maxPaths: Int = 10,
                        avoidLoops: Boolean = true): List[(Path)] = {
 
-    assume(source.length == kmerSize)
-    assume(sink.length == kmerSize)
+    assume(source.length == kmerSize, s"Source kmer ${Bases.basesToString(source)} has a size != $kmerSize")
+    assume(sink.length == kmerSize, s"Sink kmer ${Bases.basesToString(sink)} has a size != $kmerSize")
 
     var paths = List.empty[(Path)]
 
--- a/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
+++ b/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
@@ -96,11 +96,11 @@ class DeBruijnGraph(val kmerSize: Int,
         // Remove everything in the merged path from the prefix/suffix tables
         fullMergeablePath.foreach(k => {
           allNodes.remove(k)
-          removeKmer(k)
+          if (kmerCounts.contains(k)) removeKmer(k)
         })
 
         // Get full node from non-branch path
-        val mergedNode = DeBruijnGraph.mergeKmers(fullMergeablePath, kmerSize)
+        val mergedNode = DeBruijnGraph.mergeOverlappingSequences(fullMergeablePath, kmerSize)
 
         // Save each node that was merged
         fullMergeablePath.zipWithIndex.foreach({
@@ -202,7 +202,11 @@ class DeBruijnGraph(val kmerSize: Int,
     val nodeContainingSink = mergeIndex.get(sink)
 
     // Track if we branch in the path if we need to return to this point
-    var lastBranchIndex = 0
+    // Keep track of the branch and number of children
+    val lastBranchIndex = mutable.Stack[Int]()
+    val lastBranchChildren = mutable.Stack[Int]()
+    var numBacktracks = 0
+
     // explore branches until we find the sink
     // or accumulate the maximum number of appropriate length paths
     while (frontier.nonEmpty && paths.size < maxPaths) {
@@ -221,15 +225,15 @@ class DeBruijnGraph(val kmerSize: Int,
 
         // Track if this is a branching node
         if (filteredNextNodes.size > 1) {
-          lastBranchIndex = currentPath.length
+          lastBranchIndex.push(currentPath.length)
+          lastBranchChildren.push(filteredNextNodes.size)
         }
 
         // Keep searching down tree
         frontier.pushAll(filteredNextNodes)
 
-      } else {
-        // Found sink or too long
-        if (foundSink && currentPath.size + 1 >= minPathLength) {
+      } else { // Found sink or search was cut short
+        if (foundSink && currentPath.size >= minPathLength) { // Found sink with valid length
 
           // Trim merged node if the sink is inside of it
           if (foundMergedSink) {
@@ -241,8 +245,19 @@ class DeBruijnGraph(val kmerSize: Int,
           paths = (currentPath.reverse) :: paths
         } // else degenerate path, too long or too short
 
-        // Backtrack the current path to the last node with siblings
-        currentPath = currentPath.drop(currentPath.length - lastBranchIndex)
+        if (lastBranchIndex.nonEmpty) {
+          // Backtrack the current path to the last node with siblings or start
+          val backtrackIndex = lastBranchIndex.top
+          currentPath = currentPath.drop(currentPath.length - backtrackIndex)
+          numBacktracks += 1
+
+          // If we've exhausted all paths from this backtrack point, next time we will back track to the previous point
+          if (numBacktracks == lastBranchChildren.top - 1) {
+            lastBranchIndex.pop()
+            lastBranchChildren.pop()
+            numBacktracks = 0
+          }
+        }
       }
     }
     paths
@@ -304,9 +319,18 @@ object DeBruijnGraph {
     graph
   }
 
-  def mergeKmers(kmers: Seq[Kmer], kmerSize: Int): Sequence = {
-    val head = kmers.headOption.getOrElse(Seq.empty)
-    val rest = kmers.tail.flatMap(kmer => kmer.takeRight(kmer.length - kmerSize + 1))
+  /**
+   * Merge sequences where we expect consecutive entries to overlap by `overlapSize` bases.
+   *
+   * @param sequences Set of sequences that we would like to combine into a single sequence
+   * @param overlapSize The amount of the sequences we expect to overlap (The number of bases the last sequence overlaps
+   *                    with the next. For a standard kmer graph, this is the length of the kmer length - 1
+   *
+   * @return A single merged sequence
+   */
+  def mergeOverlappingSequences(sequences: Seq[Sequence], overlapSize: Int): Sequence = {
+    val head = sequences.headOption.getOrElse(Seq.empty)
+    val rest = sequences.tail.flatMap(sequence => sequence.takeRight(sequence.length - overlapSize + 1))
     head ++ rest
   }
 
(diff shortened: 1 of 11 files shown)


timodonnell:
Very cool. I didn't look at this in detail but my only suggestion is to add the basic summary you gave in this PR as a comment somewhere (maybe in GermlineAssemblyCaller.scala). 

Unit tests for the caller on specific cases would be good to have in the final version, but I like your idea to merge in this caller piecemeal so we avoid diverging even more on our separate branches. LGTM

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -213,13 +215,20 @@ class DeBruijnGraph(val kmerSize: Int,
...
+
         // Keep searching down tree
-        val nextNodes = children(next)
-        frontier ++= (if (avoidLoops) nextNodes.filterNot(visited.contains) else nextNodes)
+        frontier.pushAll(filteredNextNodes)
 
       } else {
-        //found sink or too long
+        // Found sink or too long
         if (foundSink && currentPath.size + 1 >= minPathLength) {
```
ryan-williams: I'm confused by the `currentPath.size + 1 >= minPathLength` check here; don't you want to skip if the path length is greater than `maxPathLength`?

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -213,13 +215,20 @@ class DeBruijnGraph(val kmerSize: Int,
...
       lazy val foundMergedSink = nodeContainingSink.exists(_._1 == next)
       val foundSink = (next == sink || foundMergedSink)
-      if (!foundSink && currentPath.size < maxPathLength) {
+      val nextNodes = children(next)
+      val filteredNextNodes = (if (avoidLoops) nextNodes.filterNot(visited.contains) else nextNodes)
+      if (!foundSink && filteredNextNodes.nonEmpty && currentPath.size < maxPathLength) {
+
+        // Track if this is a branching node
+        if (filteredNextNodes.size > 1) {
+          lastBranchIndex = currentPath.length
```
ryan-williams: can there be nested branches? should this keep a stack of branch indices?

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -293,9 +304,9 @@ object DeBruijnGraph {
     graph
   }
 
-  def mergeKmers(kmers: Seq[Kmer]): Sequence = {
-    val head = kmers.headOption.map(h => h.take(h.length - 1)).getOrElse(Seq.empty)
-    val rest = kmers.map(_.last)
+  def mergeKmers(kmers: Seq[Kmer], kmerSize: Int): Sequence = {
+    val head = kmers.headOption.getOrElse(Seq.empty)
+    val rest = kmers.tail.flatMap(kmer => kmer.takeRight(kmer.length - kmerSize + 1))
```
ryan-williams: so I guess you've adapted this to work on merged "k-mers"?

Calling merged-kmers (that are longer than k) "kmers" makes this a little confusing; can we try to distinguish the var names or types, or comment it at all?

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -0,0 +1,271 @@
...
+        currentWindow.referenceName,
+        referenceStart,
+        referenceEnd
+      )
+
+      val referenceKmerSource = currentReference.take(kmerSize)
+      val referenceKmerSink = currentReference.takeRight(kmerSize)
+
+      // If the current window size doesn't cover the kmer size we won't be able to find the reference start/end
+      if (referenceKmerSource.length != kmerSize || referenceKmerSink.length != kmerSize)
```
ryan-williams: is this the same as saying `currentReference.size < kmerSize`?

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -0,0 +1,271 @@
...
+        return (graph, Iterator.empty)
+
+      val paths = currentGraph.depthFirstSearch(
+        referenceKmerSource,
+        referenceKmerSink
+      )
+
+      // Take `ploidy` paths
+      val topPaths = paths
+        .take(expectedPloidy)
```
ryan-williams: won't this take the two longest paths found? is that intentional?

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -0,0 +1,271 @@
...
+        val allele = Allele(
+          referenceBases,
+          alternateBases
+        )
+
+        val depth = reads.length
+        val mappingQualities = DenseVector(reads.map(_.alignmentQuality.toFloat).toArray)
+        val baseQualities =  DenseVector(reads.flatMap(_.baseQualities).map(_.toFloat).toArray)
+        CalledAllele(
+          reads.head.sampleName,
```
ryan-williams: minor nit but maybe store these in `sampleName` and `referenceContig` `val`s outside of this method / the loops that call it?

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -0,0 +1,271 @@
...
+    /**
+     *
+     * @param graph An existing DeBruijn graph of the reads
+     * @param currentWindow Window of reads that overlaps the current loci
+     * @param kmerSize Length kmers in the DeBruijn graph
+     * @param minOccurrence Minimum times a kmer must appear to be in the DeBruijn graph
+     * @param expectedPloidy Expected ploidy, or expected number of valid paths through the graph
+     * @return A set of variants in the window
+     */
+    def discoverHaplotypes(graph: Option[DeBruijnGraph],
```
ryan-williams: `graph` is never modified in any way and is simply returned, afaict; is there a reason to include it as a param and return value?

ryan-williams:
This is awesome @arahuja. I've left some comments/questions, lmk what you think.

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
arahuja: Yea, I think that makes sense. I think I will change this to 'mergeOverlaping` or something of that sort and specify an overlap size?

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
arahuja: Yea, that also works, seems more concise will change

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
arahuja: Yea, not even necessarily the two longest but just any two really. Need to come up with some sort of weighting/scoring strategy. I think the better solution will be to take some larger number (6-8) and align them and takes the ones with the best alignments.

ryan-williams: right, DFS, so random two paths.

aligning all and choosing best two seems like the correct output.. but it seems like we could easily end up with exponentially many paths easily; hard-stopping at `maxNumPaths` isn't a great solution. I wonder what the right way to do this is.

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
ryan-williams: sure I think that would be clearer, maybe a comment about it too?

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
arahuja: This was part of a plan to have the graph updated each time rather than rebuilt each time, but this was not implemented. I can simplify this more now and remove it until it is actually implemented.

ryan-williams: gotcha, sorry if I missed that somewhere. I'm fine leaving it in but maybe a TODO comment somewhere seems warranted

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
arahuja: Maybe the nesting is confusing here, but what should happen is, we should end up here if we either found our goal or we cut-off the search. This check that if we did find the sink/goal, the path was long enough.  Will clarify some comments, but let me know if that seems wrong

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  object Caller extends SparkCommand[Arguments] {
+    override val name = "germline-assembly"
+    override val description = "call germline variants by assembling the surrounding region of reads"
+
+    /**
+     * Extract alignments that are mismatches, insertions or deletions
+     * @param alignment
+     * @return Filtered set of alignments that are mismatches, insertions or deletions
```
timodonnell: nit: it seems to be returning a Seq not Set, and also pairs, might be worth clarifying what the second element of the pair is

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -0,0 +1,271 @@
...
+     */
+    def getVariantAlignments(alignment: ReadAlignment): Seq[(AlignmentState, Int)] = {
+      alignment
+        .alignments
+        .zipWithIndex
+        .filter(tup => AlignmentState.isGapAlignment(tup._1) || tup._1 == AlignmentState.Mismatch)
+    }
+
+    /**
+     *
```
timodonnell: An overview sentence or two on what this does would be helpful

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
arahuja: You are right! A stack would fix this, along with checking how often we back-tracked.

arahuja:
Thanks @timodonnell and @ryan-williams for all the feedback! Updated with comments where necessary as well as some of the various simplifications. @ryan-williams thanks for pointing out the issue with the multiple splits in the graph that was not handled. I added the stack as suggested and a test case where we need to backtrack twice to find the right path.

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -213,14 +219,21 @@ class DeBruijnGraph(val kmerSize: Int,
...
         // Keep searching down tree
-        val nextNodes = children(next)
-        frontier ++= (if (avoidLoops) nextNodes.filterNot(visited.contains) else nextNodes)
+        frontier.pushAll(filteredNextNodes)
 
-      } else {
-        //found sink or too long
-        if (foundSink && currentPath.size + 1 >= minPathLength) {
+      } else { // Found sink or search was cut short
+        if (foundSink && currentPath.size + 1 >= minPathLength) { // Found sink with valid length
```
ryan-williams: thanks for the added comment; forgot to ask before: why `+ 1` here?

ryan-williams:
Thanks @arahuja; the big open question seems to be about scoring paths vs. returning two random ones.. do you want to tackle that here or later? I'm fine either way.

There are a few tests I could see adding also, e.g. a case where a valid path is found as the third path from a branch-point. Up to you though.

arahuja:
> the big open question seems to be about scoring paths vs. returning two random ones.. do you want to tackle that here or later? I'm fine either way.

Yea for this, I'd like to get some more experience with this on real data - basically how many valid paths are usually found. (My suspusicion is it will be small unless we are in a repetitive region, but then we should find a way to identify to those and do something else)

>a case where a valid path is found as the third path from a branch-point. Up to you though.
Not sure I get this test case exactly, but I think we can add these as we find some interesting cases. If you have some specific examples that might fail, please add them or let me know though.

ryan-williams:
> basically how many valid paths are usually found. (My suspusicion is it will be small…

It seems to me that any single-bp seq-error at the location of a het SNV will give a 2/3 chance of not picking the right 2 (of 3) paths, with the current approach; is that consistent with your understanding?

arahuja:
> It seems to me that any single-bp seq-error at the location of a het SNV will give a 2/3 chance of not picking the right 2 (of 3) paths, with the current approach; is that consistent with your understanding?

Yes, but `minOccurence` should prune out most of the simpler cases.

ryan-williams:
Fair enough; do we have any way of learning "how many valid paths are usually found"? Should we log a warning or something if it's >2? Or how will we know?

arahuja:
> Should we log a warning or something if it's >2? Or how will we know?

I was thinking we would just find some interesting variants and start testing on them to get a sense for performance. With that we can develop a reasonable scoring strategy.

ryan-williams:
I think we're agreed on these two things:

* if there are >2 paths, this code will do the wrong thing with >50% probability.
* in a germline-calling scenario this should ~never happen.

Given this, I think we should throw an exception. Occasionally outputting variants we know are most likely wrong, and having to debug this later starting from such calls, sounds like a bad time.

arahuja:
>  I think we should throw an exception.  
I don't want to throw an exception as this would halt execution on a dataset where it did happen and make it difficult to examine the rest of the results.

I've added a parameter to score them if there are less them `maxNumberPathsToScore` and to otherwise output no variants and log a warning.

ryan-williams:
thanks, lgtm

PR #373, opened 2016-01-26
faster unmasking (i.e. lowercasing) of reference sequence in Bases.scala

timodonnell:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/373)


ryan-williams:
lgtm

PR #377, opened 2016-01-28
Pass reference genome to load function in germline callers

### LINKED ISSUE #376: Allow specifying of a reference fasta and recomputing mdtags
Although there is a command-line option to allow the recomputation of mdTags, we never pass along the reference genome for this to happen in `germline-standard` or `germline-threshold`

See #375

arahuja:
Fixes #376

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/377)


ryan-williams:
lgtm!

PR #378, opened 2016-01-29
Make Structural Variant caller work for adam.

MartijnAB:
The boolean in line 106 of StructuralVariantCaller.scala is always false, because of the different types of line 519 and 557 in Read.scala. So I changed that. To work with Delly you need to know the insert size, that is not loaded in 470 Read.scala (see https://github.com/bigdatagenomics/bdg-formats/issues/37 and https://github.com/bigdatagenomics/adam/issues/729). The “inferredInsertSize” is not always None for Dally 559 in Read.scala. So I changed that to make it Null when it is None or zero (43 MateAlignmentProperties.scala).

This makes the boolean on line 106 of StructuralVariantCaller.scala true and ensures “getExceptionalReads” has an insert size to work with. So that Structural Variant caller, no more empty files returns.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/378)


ryan-williams:
Thanks for the fixes @MartijnAB!

PR #380, opened 2016-02-03
Update aligner to ensure left-aligned variants

arahuja:
This ensures that alignments (specifically ambiguous insertions or deletions) are in the left-most position as opposed as they currently are (in their right-most position).

This is done by just performing the alignment from end to beginning instead of beginning to end. Since we are choosing optimal transitions at each turn we delay the gap as long as possible. Previously this meant pushing it later in the sequence, now it is pushed earlier into the sequence.

This makes variant comparison easier as most tools push these to the left.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/380)

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignmentSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignmentSuite.scala
@@ -13,7 +13,7 @@ class AffineGapPenaltyAlignmentSuite extends FunSuite with Matchers {
       openGapProbability = 1e-3,
       closeGapProbability = 1e-2
     )
-    alignments(4)._3.toInt should be(0)
+    alignments(0)._3.toInt should be(0)
   }
 
   test("score alignment: single mismatch") {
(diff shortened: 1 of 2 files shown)


src/test/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignmentSuite.scala
```diff
@@ -13,7 +13,7 @@ class AffineGapPenaltyAlignmentSuite extends FunSuite with Matchers {
       openGapProbability = 1e-3,
       closeGapProbability = 1e-2
     )
-    alignments(4)._3.toInt should be(0)
+    alignments(0)._3.toInt should be(0)
```
ryan-williams: why did this go from the 5th alignment to the 1st?

ryan-williams:
lgtm modulo one thing about the tests i'm not grokking

src/test/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignmentSuite.scala
arahuja: The alignments score matrix is in reverse now, so the full alignment is no longer in the last position, but in the first.

arahuja:
thanks @ryan-williams!

PR #382, opened 2016-02-09
Expanded tests for germline-assembly caller

arahuja:
This expands the test suite for the germline assembly caller, adding specific variants to find of different varieties.

The test suite is also altered to create a single spark context for the full suite (as opposed to a spark context per test). This lets us load the reads/reference once rather than in each test.

This also fixes an issue with merging nodes in the DeBruijn graph (nodes were not being properly removed), a test case has been added.

Additionally, makes two changes to the germline caller itself:
- Instead of search from a reference from start of the first read, it uses the start with the window. The first read may extend earlier than the start of the window, and is likely the only read that stretches that far back. Therefore, only paths supported by that _single_ read would have been found.
- Instead of first doing a pass to find variant pileups, this checks this in the windows as it is processing. It find's variable windows, and only performs the assembly on those regions.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/382)


src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -212,9 +223,16 @@ class DeBruijnGraph(val kmerSize: Int,
     while (frontier.nonEmpty && paths.size < maxPaths) {
       val next = frontier.pop()
 
+      if (debugPrint) {
+        if (currentPath.length == 0) {
```
ryan-williams: `.isEmpty`

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
```diff
@@ -37,12 +37,12 @@ object AffineGapPenaltyAlignment {
         closeGapProbability
       )
 
-    val ((refStartIdx, path, score), refEndIdx) =
+    val ((refEndIdx, path, score), refStartIdx) =
```
ryan-williams: was this mix-up causing broken behavior?

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -241,13 +260,18 @@ class DeBruijnGraph(val kmerSize: Int,
...
           // Found legitimate path to sink, save path
-          paths = (currentPath.reverse) :: paths
+          paths = currentPath.reverse :: paths
         } // else degenerate path, too long or too short
 
         if (lastBranchIndex.nonEmpty) {
           // Backtrack the current path to the last node with siblings or start
           val backtrackIndex = lastBranchIndex.top
+          // Remove the nodes from the last branch from the visited set
+          currentPath.slice(0, currentPath.length - backtrackIndex).map(visited.remove)
```
ryan-williams: nit: `s/map/foreach/`?

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -104,12 +104,18 @@ object GermlineAssemblyCaller {
       val referenceKmerSink = currentReference.takeRight(kmerSize)
 
       // If the current window size doesn't cover the kmer size we won't be able to find the reference start/end
-      if (currentReference.size < kmerSize)
+      if (currentReference.size < kmerSize || referenceKmerSource == referenceKmerSink)
```
ryan-williams: hm, I'm not sold on this added clause. is there nothing useful we can do in regions where the reference as a k-repeat? I can imagine a reasonable-sized window with identical `k`-mers on each end where we'd still want to call variants inside it..

src/main/scala/org/hammerlab/guacamole/alignment/ReadAlignment.scala
```diff
@@ -32,7 +32,8 @@ object AlignmentState extends Enumeration {
  * @param alignmentScore Score of the alignment
```
ryan-williams: add the new params? they're a closed interval?

src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala
```diff
@@ -229,6 +229,32 @@ class DeBruijnGraphSuite extends GuacFunSuite with Matchers {
...
+      Seq(longRead),
+      kmerSize,
+      minOccurrence = 1,
+      mergeNodes = true
+    )
+
+    longRead
+      .sliding(kmerSize)
+      .zipWithIndex.map(
+        {
```
ryan-williams: nit: IJ doesn't like the parends around the braces here

ryan-williams: also wants `s/map/foreach/`

src/main/scala/org/hammerlab/guacamole/alignment/AffineGapPenaltyAlignment.scala
arahuja: Somewhat, they weren't used anywhere, but when I added some tests and wanted to use them I noticed.

src/test/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCallerSuite.scala
```diff
@@ -1,34 +1,203 @@
...
+  args.parallelism = 2
 
-    GermlineAssemblyCaller.Caller.run(args, sc)
+  var sc: SparkContext = _
+  var reference: ReferenceBroadcast = _
+  var readSet: ReadSet = _
 
-    val variants = new ADAMContext(sc).loadVariants(args.variantOutput).collect()
+  before {
+    sc = Common.createSparkContext()
```
ryan-williams: I was struggling with this recently; surprised that `sparkTest`, `sparkBefore` etc. each set up and tear down a SparkContext! Will have to use this pattern more often

src/main/scala/org/hammerlab/guacamole/alignment/ReadAlignment.scala
```diff
@@ -32,7 +32,8 @@ object AlignmentState extends Enumeration {
...
  */
 case class ReadAlignment(alignments: Seq[AlignmentState],
-                         refBases: Seq[Byte],
```
arahuja: Sorry, didn't mean you delete your comment here @ryan-williams , meant to delete my empty response

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
arahuja: Yes, definitely possible, but as this is setup, I don't see a good way to handle it for now. Possible solutions may be just increasing the kmer size and looking again, but just punting on it for now.

src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala
arahuja: > nit: IJ doesn't like the parends around the braces here

weird, that looks weird to me, but I'm ok with it

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
ryan-williams: OK, punting wfm. I see that there is a more general inability to handle regions with repeating k-mers atm. I wonder if that happens frequently? Adding a TODO and/or a warning about this case might make sense.

Basically any time we skip a node in the DBG DFS due to it already being in `visited`, we are discarding a possible valid path with a repeat segment in it, right?

ryan-williams:
lgtm % one comment about whether we should leave any breadcrumbs in the repeat-kmer case

arahuja:
@ryan-williams thanks for reviewing! Added a warning on the invalid start and end kmers, should be useful for diagnosing when/how often it occurs.

PR #393, opened 2016-03-09
Updating germline assembly to determine anchor from reads

arahuja:
This PR has 3 key updates:
1. Updates the source/sink searching to first determine the possible start/end kmers (instead of just relying on the reference) and then attempting to find all paths between all pairs of those
2. Instead of filtering to regions when any of the reads exceed the variability threshold, this waits for those reads to enter the center. This fixes the issue when only the end kmer contains the variant and we can't find a proper path to the reference kmer at that point.
3. Refactors a few aspects into functions. This is so these can be reused in the somatic version

---
This change is [Review on Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/393)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
@@ -86,13 +86,12 @@ object GermlineAssemblyCaller {
     def getConsensusKmer(reads: Seq[MappedRead],
                          startLocus: Int,
                          endLocus: Int,
-                         kmerSize: Int,
                          minOccurrence: Int): Iterable[Array[Byte]] = {
 
       // Filter to reads that entirely cover the region
       // Exclude reads that have any non-M Cigars (these don't have a 1 to 1 base mapping to the region)
       val overlapping = reads
-        .filter(r => !r.cigarElements.exists(el => el.getOperator != CigarOperator.M))
+        .filter(!_.cigarElements.exists(_.getOperator != CigarOperator.M))
         .filter(r => r.overlapsLocus(startLocus) && r.overlapsLocus(endLocus - 1))
 
       // Extract the sequences from each region
@@ -101,7 +100,6 @@ object GermlineAssemblyCaller {
 
       // Filter to sequences that appear at least `minOccurrence` times
       sequences
-        .filter(_.size == kmerSize)
         .groupBy(identity)
         .map(kv => (kv._1, kv._2.length))
         .filter(_._2 >= minOccurrence)
@@ -303,18 +301,18 @@ object GermlineAssemblyCaller {
           (graph, windows) => {
             val window = windows.head
 
-            // Find the reads the overlap the center locus
-            val centerReads =
+            // Find the reads the overlap the center locus/ current locus
+            val currentLocusReads =
               window
                 .currentRegions()
                 .filter(_.overlapsLocus(window.currentLocus))
             val variableReads =
-              centerReads
+              currentLocusReads
                 .count(read => read.cigar.numCigarElements() > 1 || read.mdTagOpt.get.countOfMismatches > 0)
 
-            // Computer the number reads with variant bases from the reads overlapping the center
-            val centerVAF = variableReads.toFloat / centerReads.length
-            if (centerVAF > minAreaVaf) {
+            // Compute the number reads with variant bases from the reads overlapping the currentLocus
+            val currentLocusVAF = variableReads.toFloat / currentLocusReads.length
+            if (currentLocusVAF > minAreaVaf) {
               val result = discoverHaplotypes(
                 None,
                 window,
@@ -353,7 +351,7 @@ object GermlineAssemblyCaller {
                                kmerSize: Int,
                                minOccurrence: Int,
                                maxPaths: Int,
-                               debugPrint: Boolean = false): Seq[DeBruijnGraph#Path] = {
+                               debugPrint: Boolean = false) = {
       val referenceKmerSource = referenceSequence.take(kmerSize)
       val referenceKmerSink = referenceSequence.takeRight(kmerSize)
 
@@ -361,7 +359,6 @@ object GermlineAssemblyCaller {
         reads,
         referenceStart,
         referenceStart + kmerSize,
-        kmerSize = kmerSize,
         minOccurrence = minOccurrence
       ) ++ Seq(referenceKmerSource)).toSet
 
@@ -369,7 +366,6 @@ object GermlineAssemblyCaller {
         reads,
         referenceEnd - kmerSize,
         referenceEnd,
-        kmerSize = kmerSize,
         minOccurrence = minOccurrence
       ) ++ Seq(referenceKmerSink)).toSet
 
@@ -381,8 +377,8 @@ object GermlineAssemblyCaller {
       )
 
       for {
-        source <- sources;
-        sink <- sinks;
+        source <- sources
+        sink <- sinks
         path <- currentGraph.depthFirstSearch(
           source,
           sink,
--- a/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
@@ -59,6 +59,57 @@ object GermlineAssemblyCaller {
     override val description = "call germline variants by assembling the surrounding region of reads"
 
     /**
+     * Extract a subsequence from a read sequence
+     *
+     * NOTE: This assumes that the read entirely covers the reference region and does not check for insertions
+     * or deletion
+     *
+     * @param read Read to extract subsequence from
+     * @param startLocus Start (inclusive) locus on the reference
+     * @param endLocus End (exclusive) locus on the reference
+     * @return Subsequence overlapping [startLocus, endLocus)
+     */
+    def getSequenceFromRead(read: MappedRead, startLocus: Int, endLocus: Int) = {
+      read.sequence.slice(startLocus - read.unclippedStart.toInt, endLocus - read.unclippedStart.toInt)
+    }
+
+    /**
+     * For a given set of reads identify all kmers that appear in the specified reference region
+     *
+     * @param reads  Set of reads to extract sequence from
+     * @param startLocus Start (inclusive) locus on the reference
+     * @param endLocus End (exclusive) locus on the reference
+     * @param kmerSize Length of the subsequence
+     * @param minOccurrence Mininum number of times a subsequence needs to appear to be included
+     * @return List of subsequences overlapping [startLocus, endLocus) that appear at least `minOccurrence` time
+     */
+    def getConsensusKmer(reads: Seq[MappedRead],
+                         startLocus: Int,
+                         endLocus: Int,
+                         kmerSize: Int,
+                         minOccurrence: Int) = {
+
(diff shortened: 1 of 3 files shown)


src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -59,6 +59,57 @@ object GermlineAssemblyCaller {
...
+     *
+     * NOTE: This assumes that the read entirely covers the reference region and does not check for insertions
+     * or deletion
+     *
+     * @param read Read to extract subsequence from
+     * @param startLocus Start (inclusive) locus on the reference
+     * @param endLocus End (exclusive) locus on the reference
+     * @return Subsequence overlapping [startLocus, endLocus)
+     */
+    def getSequenceFromRead(read: MappedRead, startLocus: Int, endLocus: Int) = {
```
timodonnell: No need to actually use this, but just FYI this may be useful here https://github.com/hammerlab/guacamole/blob/master/src/main/scala/org/hammerlab/guacamole/commands/jointcaller/ReadSubsequence.scala#L63

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -59,6 +59,57 @@ object GermlineAssemblyCaller {
...
+    def getConsensusKmer(reads: Seq[MappedRead],
+                         startLocus: Int,
+                         endLocus: Int,
+                         kmerSize: Int,
+                         minOccurrence: Int) = {
+
+      // Filter to reads that entirely cover the region
+      // Exclude reads that have any non-M Cigars (these don't have a 1 to 1 base mapping to the region)
+      val overlapping = reads
+        .filter(r => r.cigarElements.count(el => el.getOperator != CigarOperator.M) == 0)
```
timodonnell: nit: `!cigarElements.exists` may be clearer than a `count(...) == 0`

timodonnell: also is it intentional to discard reads that have non-M cigars only at sites outside the region of interest?

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -59,6 +59,57 @@ object GermlineAssemblyCaller {
...
+        .filter(r => r.overlapsLocus(startLocus) && r.overlapsLocus(endLocus - 1))
+
+      // Extract the sequences from each region
+      val sequences = overlapping
+        .map(r => (r, getSequenceFromRead(r, startLocus, endLocus)))
```
timodonnell: why make this into pairs if only the second element in the pair is used below?

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -59,6 +59,57 @@ object GermlineAssemblyCaller {
...
+     * @param endLocus End (exclusive) locus on the reference
+     * @param kmerSize Length of the subsequence
+     * @param minOccurrence Mininum number of times a subsequence needs to appear to be included
+     * @return List of subsequences overlapping [startLocus, endLocus) that appear at least `minOccurrence` time
+     */
+    def getConsensusKmer(reads: Seq[MappedRead],
+                         startLocus: Int,
+                         endLocus: Int,
+                         kmerSize: Int,
+                         minOccurrence: Int) = {
```
timodonnell: might be good to have the return type here

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -59,6 +59,57 @@ object GermlineAssemblyCaller {
...
+        .map(r => (r, getSequenceFromRead(r, startLocus, endLocus)))
+
+      // Filter to sequences that appear at least `minOccurrence` times
+      sequences
+        .map(_._2)
+        .filter(_.size == kmerSize)
+        .groupBy(identity)
+        .map(kv => (kv._1, kv._2.length))
+        .filter(_._2 >= minOccurrence)
+        .map(_._1.toArray).toList
```
timodonnell: why `toList`? is that actually more efficient?

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -301,5 +333,65 @@ object GermlineAssemblyCaller {
...
+     * @return List of paths that traverse the region
+     */
+    def discoverPathsFromReads(reads: Seq[MappedRead],
+                               referenceStart: Int,
+                               referenceEnd: Int,
+                               referenceSequence: Array[Byte],
+                               kmerSize: Int,
+                               minOccurrence: Int,
+                               maxPaths: Int,
+                               debugPrint: Boolean = false) = {
```
timodonnell: return type might be good to have

timodonnell:
This looks good to me, made a few small comments

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
arahuja: Updated to the `exists`. Yea, for now I'm filtering out anything non-M since otherwise it's difficult to figure out where that reference base is in the read sequence without processing the full cigar. i.e. if their was a insertion/deletion earlier in the sequence the position would need to be shifted

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
arahuja: Good point, I can drop that

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
arahuja: Yea it could be an iterable, just effects how the additional item is appended below, updated that

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
arahuja: Added

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
arahuja: good to know, I just need the sequence for now though.

ryan-williams:
looks good overall, couple optional nits and one question about the params to `getConsensusKmer`.

also, thoughts on adding a test of a source and/or sink that differs from the reference?

---

Review status: 0 of 3 files reviewed at latest revision, 10 unresolved discussions, some commit checks broke.

---

*[src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala, line 95 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/393#-KCRJk6uO9g_jezDFcJ--r1-95:-KCaXIL-4fsl4zFQEIlf:-525421262) ([raw file](https://github.com/hammerlab/guacamole/blob/12b0a6f64fc93ecbf83aedff1eccef2e7832347d/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala#L95)):*
nit, feel free to disregard, but you could shorten this with a couple `_`s:
```
.filter(!_.cigarElements.exists(_.getOperator != CigarOperator.M))
```

---

*[src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala, line 104 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/393#-KCaVy2xuEuoXVHpz1pP:-KCaVy2xuEuoXVHpz1pQ:-1421306649) ([raw file](https://github.com/hammerlab/guacamole/blob/07dd84923d10bdf789c9aed75a01e3a96821e062/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala#L104)):*
I'm confused by `kmerSize` vs. `endLocus - startLocus`; won't the size of most/all of these be the latter?

update: I see that the calls both ensure that `endLocus - startLocus == kmerSize`; I'd vote to change the API to this function if that's a precondition.

---

*[src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala, line 306 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/393#-KCaXtg06SVsDTEN4A94:-KCaXtg06SVsDTEN4A95:-981277231) ([raw file](https://github.com/hammerlab/guacamole/blob/07dd84923d10bdf789c9aed75a01e3a96821e062/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala#L306)):*
j/w: is "center" locus synonymous with "current" locus here? cf. `window.currentLocus` below

---

*[src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala, line 315 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/393#-KCaXkDeaAVAvDzJyE6x:-KCaXkDeaAVAvDzJyE6y:-627854012) ([raw file](https://github.com/hammerlab/guacamole/blob/07dd84923d10bdf789c9aed75a01e3a96821e062/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala#L315)):*
`s/Computer/Compute/` :)

---

*[src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala, line 384 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/393#-KCaYDvmUSlt5wkx52XI:-KCaYDvne3P8z0MeAQI_:-293560435) ([raw file](https://github.com/hammerlab/guacamole/blob/07dd84923d10bdf789c9aed75a01e3a96821e062/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala#L384)):*
nit: intentional semicolons? IJ flags them as unnecessary but i'm fine to disable that if you like them

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/393#-:-KCaYxfjHQc287-Hilza:422154205)*

arahuja:
Review status: 0 of 3 files reviewed at latest revision, 10 unresolved discussions.

---

*[src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala, line 306 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/393#-KCaXtg06SVsDTEN4A94:-KCfWLx1wtAyEYA62cAd:2064357574) ([raw file](https://github.com/hammerlab/guacamole/blob/07dd84923d10bdf789c9aed75a01e3a96821e062/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala#L306)):*
Yes it is, I can update that naming

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/393)*

arahuja:
I'm going merge this in to keep things moving, I think what's left is @ryan-williams 's comment on adding a specific test. I will need to find one in the checked in datasets, but this did improve sensitivity a lot

thanks for reviewing @timodonnell and @ryan-williams!

PR #396, opened 2016-03-21
Bump ADAM version to 0.19

timodonnell:
Small fixes due to changes in ADAM, as well as some very minor unrelated fixes

I did this originally in the hope that it would fix another issue (turned out to be https://github.com/bigdatagenomics/adam/issues/742, which is still open) but I thought we might as well use the latest version

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/396)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
   <name>guacamole: variant caller</name>
 
   <properties>
-    <adam.version>0.18.1</adam.version>
+    <adam.version>0.19.0</adam.version>
     <bdg-formats.version>0.6.1</bdg-formats.version>
     <java.version>1.8</java.version>
     <scala.version>2.10.3</scala.version>
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,7 @@
 
   <properties>
     <adam.version>0.19.0</adam.version>
-    <bdg-formats.version>0.6.1</bdg-formats.version>
+    <bdg-formats.version>0.7.0</bdg-formats.version>
     <java.version>1.8</java.version>
     <scala.version>2.10.3</scala.version>
     <scala.version.prefix>2.10</scala.version.prefix>
--- a/src/main/scala/org/hammerlab/guacamole/reads/Read.scala
+++ b/src/main/scala/org/hammerlab/guacamole/reads/Read.scala
@@ -478,7 +478,8 @@ object Read extends Logging {
 
     val adamContext = new ADAMContext(sc)
 
-    val adamRecords: RDD[AlignmentRecord] = adamContext.loadAlignments(filename, projection = None, stringency = ValidationStringency.LENIENT)
+    val adamRecords: RDD[AlignmentRecord] = adamContext.loadAlignments(
+      filename, projection = None, stringency = ValidationStringency.LENIENT).rdd
     val sequenceDictionary = new ADAMSpecificRecordSequenceDictionaryRDDAggregator(adamRecords).adamGetSequenceDictionary()
 
     val allReads: RDD[Read] = adamRecords.map(fromADAMRecord(_, token, referenceGenome))
@@ -546,7 +547,7 @@ object Read extends Logging {
       )
       else
         None
-      PairedRead(read, isFirstInPair = alignmentRecord.getReadNum == 1, mateAlignment)
+      PairedRead(read, isFirstInPair = alignmentRecord.getReadInFragment == 1, mateAlignment)
     } else {
       read
     }
--- a/src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
@@ -18,8 +18,9 @@
 
 package org.hammerlab.guacamole.reads
 
+import org.apache.parquet.hadoop.metadata.CompressionCodecName
 import org.apache.spark.rdd.RDD
-import org.bdgenomics.adam.rdd.ADAMContext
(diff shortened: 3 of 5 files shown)


src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -478,7 +478,8 @@ object Read extends Logging {
 
     val adamContext = new ADAMContext(sc)
 
-    val adamRecords: RDD[AlignmentRecord] = adamContext.loadAlignments(filename, projection = None, stringency = ValidationStringency.LENIENT)
+    val adamRecords: RDD[AlignmentRecord] = adamContext.loadAlignments(
+      filename, projection = None, stringency = ValidationStringency.LENIENT).rdd
```
arahuja: Interesting change, what happens without `.rdd`?

src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
```diff
@@ -85,6 +85,9 @@ class ReadSetSuite extends GuacFunSuite with Matchers {
     readSet.reads.count should be(23)
   }
 
+  // This test is disabled for now since loading ADAM files is not a common use case in Guacamole, and the ADAM API
+  // is in flux.
+  /*
```
arahuja: Is this test failing now?

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
timodonnell: Compile error. The loadAlignments [method](https://github.com/bigdatagenomics/adam/blob/2996ee644307c312fd9e8e75c560627a2c21b300/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L851) seems to now return an `AlignmentRecordRDD` not an actual RDD

src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
timodonnell: Yeah, it's not finding a `.seqdict` file that seems to have been introduced. I looked briefly and couldn't immediately see how to create such a file so I thought I'd punt on it and disable the test for now. ( @fnothaft or @laserson , do you happen to know the fix for this off hand? thanks!)

```
--> [0.76 sec. later]: Using ADAM to read: TestUtil.73707d27-ce24-45f7-8368-7f24c0b38473.adam
- load read from ADAM *** FAILED ***
  java.io.FileNotFoundException: File TestUtil.73707d27-ce24-45f7-8368-7f24c0b38473.adam.seqdict does not exist
  at org.apache.hadoop.fs.RawLocalFileSystem.deprecatedGetFileStatus(RawLocalFileSystem.java:511)
  at org.apache.hadoop.fs.RawLocalFileSystem.getFileLinkStatusInternal(RawLocalFileSystem.java:722)
  at org.apache.hadoop.fs.RawLocalFileSystem.getFileStatus(RawLocalFileSystem.java:501)
  at org.apache.hadoop.fs.FilterFileSystem.getFileStatus(FilterFileSystem.java:398)
  at org.apache.hadoop.fs.ChecksumFileSystem$ChecksumFSInputChecker.<init>(ChecksumFileSystem.java:137)
  at org.apache.hadoop.fs.ChecksumFileSystem.open(ChecksumFileSystem.java:339)
  at org.apache.hadoop.fs.FileSystem.open(FileSystem.java:765)
  at org.bdgenomics.adam.rdd.ADAMContext.loadAvro(ADAMContext.scala:401)
  at org.bdgenomics.adam.rdd.ADAMContext.loadParquetAlignments(ADAMContext.scala:477)
  at org.bdgenomics.adam.rdd.ADAMContext$$anonfun$loadAlignments$1.apply(ADAMContext.scala:881)

Comment (arahuja) [inline src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala]: It does seem like `saveAsParquet` should create them

```
  def saveAsParquet(args: ADAMSaveAnyArgs,
                    sd: SequenceDictionary,
                    rgd: RecordGroupDictionary) = {
....
    // write the sequence dictionary and record group dictionary to disk
    saveAvro("%s.seqdict".format(args.outputPath),
      rdd.context,
      Contig.SCHEMA$,
      contigs)
    saveAvro("%s.rgdict".format(args.outputPath),
      rdd.context,
      RecordGroupMetadata.SCHEMA$,
      rgMetadata)
```

File: pom.xml [hunk 3]
```diff
@@ -16,7 +16,7 @@
   <name>guacamole: variant caller</name>
 
   <properties>
-    <adam.version>0.18.1</adam.version>
+    <adam.version>0.19.0</adam.version>
     <bdg-formats.version>0.6.1</bdg-formats.version>
```
arahuja: Also does this not require an update to bdg-formats version?

timodonnell: Oh good call -- I wonder if that's related to my saveAsParquet issue

timodonnell: I tried bumping up to bdg-formats 0.7.0 but this problem persists

timodonnell:
Just pushed a fix for the test. Switching from `adamParquetSave` to `saveAsParquet` fixed the problem. (Not sure what the exact semantics are of these various save methods.) Also bumped bdg formats to 0.7.0, although that does not appear to have any obvious impact on Guacamole. If the travis tests succeed this should be good to merge.

arahuja:
Cool, @timodonnell LGTM. I guess I was looking at `saveAsParquet` too, I didn't notice that the code was using the other. Was also confused why the stacktrace didn't match the function, but I guess that's why.

PR #400, opened 2016-03-25
Update README.md

arahuja:
Update Hadoop example to match local calling example

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/400)

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,9 +88,9 @@ spark-submit \
 	--verbose \
 	/path/to/target/guacamole-with-dependencies-<x.y.z>.jar \
 	somatic-joint \
-    		src/test/resources/synth1.normal.100k-200k.withmd.bam \
-    		src/test/resources/synth1.tumor.100k-200k.withmd.bam \
-    		--reference-fasta /path/to/your/b37.fasta \
+    		/hdfs/path/to/normal.bam \
+    		/hdfs/path/to/tumor.bam \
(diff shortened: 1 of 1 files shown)


timodonnell:
LGTM

README.md
```diff
@@ -87,9 +87,11 @@ spark-submit \
 	--class org.hammerlab.guacamole.Guacamole \
 	--verbose \
 	/path/to/target/guacamole-with-dependencies-<x.y.z>.jar \
-	germline-threshold \
-        --reads hdfs:///path/to/reads.bam \
-        --out hdfs:///path/to/result.vcf
+	somatic-joint \
+    		src/test/resources/synth1.normal.100k-200k.withmd.bam \
```
ryan-williams: this might be somewhat misleading because these are valid relative paths in the repo but will only resolve if you have files at these paths in your HDFS deployment, right?

arahuja: Yea, was thinking that as well, should I just put placeholders to HDFS bams?

ryan-williams: I'd put a path or path placeholder that doesn't suggest it's reading from the test `.bam`s in this repo, and maybe add a comment here/above explaining that in `--master yarn` mode paths resolve to HDFS by default.

When to use `file://`, `hdfs://`, or neither proves consistently tricky for newcomers so it would be nice to have a note about that somewhere, but this doesn't have to be the place for it or time to add it.

PR #405, opened 2016-03-28
Fix reference sequence in SomaticStandardCallerSuite. Closes #403

### LINKED ISSUE #403: After upgrade to adam 0.19, many tests use mostly mismatching read alignments.
While working on my mutect-like algorithm, and porting tests over from the `SomaticStandardCallerSuite`, I noticed that the [multi-base deletion](https://github.com/hammerlab/guacamole/blob/master/src/test/scala/org/hammerlab/guacamole/commands/SomaticStandardCallerSuite.scala#L178 ) test was not working, while it seemed to work fine for the `SomaticStandardCaller`. After digging in I noticed that the "sum of mismatching base quality score filter" was removing all of the reads from the tumor sample. On digging deeper into the test, I noticed that the test was set up to put the normal/tumor read on a reference sequence that didn't match the normal read sequence. I was able to get that test passing by adding another contig that matched the normal read sequence, and use that instead.

I am a little concerned that this might be a more global problem. Do any other methods in guacamole make use of filters based on mismatching bases? If so any `negative` tests need to be examined carefully to make sure this recent transition away from user defined MD tags is not causing any other unexpected issues.

In the case of the SomaticStandardCaller I think that bogus alignment is no problem. Basically it is an alignment where all bases either mismatch, or based on the cigar are marked as deleted, so I think this test is still doing something useful in that case. 

As a user though it threw me off, and I am not sure that my other passing negative tests in MuTect are because a bogus alignment is getting filtered in my base caller function, or if it is actually negative. I'll have to be careful :)

timodonnell:
Tentative trivial fix for #403 

By adding intentional mistakes into `SomaticStandardCaller` I was able to get the `testing complex region negative variants on syn1` negative test to fail, so I have reasonable confidence that these tests are in fact testing something. Most of the tests involve both positive and negative assertions too, which also inspires some confidence.

Minor note: I wasn't able to get the `"difficult negative variants"` test to fail by changing the`SomaticStandardCaller.findVariantsAtLocus` since the tumor VAF appears to be 0, causing the`SomaticGenotypeFilter` filtering to remove it.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/405)

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/commands/SomaticStandardCallerSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/commands/SomaticStandardCallerSuite.scala
@@ -34,7 +34,8 @@ class SomaticStandardCallerSuite extends GuacFunSuite with Matchers with TableDr
   def simpleReference = TestUtil.makeReference(sc, Seq(
     ("chr1", 0, "TCGATCGACG"),
     ("chr2", 0, "TCGAAGCTTCG"),
-    ("chr3", 10, "TCGAATCGATCGATCGA")))
+    ("chr3", 10, "TCGAATCGATCGATCGA"),
+    ("chr4", 0, "TCGAAGCTTCGAAGCT")))
 
   def loadPileup(filename: String, contig: String, locus: Long = 0): Pileup = {
     val contigReference = grch37Reference.getContig(contig)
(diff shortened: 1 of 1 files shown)


ryan-williams:
:lgtm:

haven't tested these myself but seems reasonable

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/guacamole/405#-:-KDyNpWOFiVo155ImvU9:-545518946)*

PR #407, opened 2016-03-31
Change Seq to Vector and other small fixes

timodonnell:
* Change Seq to Vector in a number of places. For vast majority of cases, Vector performs better. Where possible, we use Vector for concreteness but in some places, especially deserialization, it's desirable to be able to pass Arrays around without copying them to a Vector. In these places, we use IndexedSeq.
 * Also change per-sample Seq's to be of type PerSample in a few places
 * Add isClipped method to Alignment
 * Small comment updates

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/407)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/Bases.scala
+++ b/src/main/scala/org/hammerlab/guacamole/Bases.scala
@@ -84,7 +84,7 @@ object Bases {
   }
 
   /** Convert a string (e.g. "AAAGGC") to a byte array. */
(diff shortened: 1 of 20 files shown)


timodonnell:
The impetus for this was watching some cluster runs (hitting 'thread dump') and often seeing tasks that are spending time trying to index into a Seq that is List-backed

ryan-williams:
lgtm

PR #409, opened 2016-04-01
Remove scalariform, print full stack traces on test failures

ryan-williams:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/409)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,6 +173,7 @@
           <junitxml>.</junitxml>
           <filereports>GuacamoleTestSuite.txt</filereports>
           <argLine>-Xmx3072m</argLine>
+          <stdout>F</stdout>
         </configuration>
         <executions>
           <execution>
--- a/pom.xml
+++ b/pom.xml
@@ -185,25 +185,6 @@
         </executions>
       </plugin>
       <plugin>
-        <groupId>org.scalariform</groupId>
-        <artifactId>scalariform-maven-plugin</artifactId>
-        <version>0.1.4</version>
-        <executions>
-          <execution>
-            <phase>process-sources</phase>
-            <goals>
-              <goal>format</goal>
-            </goals>
-            <configuration>
-              <alignParameters>true</alignParameters>
-              <alignSingleLineCaseStatements>true</alignSingleLineCaseStatements>
-              <doubleIndentClassDeclaration>true</doubleIndentClassDeclaration>
-              <preserveDanglingCloseParenthesis>true</preserveDanglingCloseParenthesis>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-release-plugin</artifactId>
         <version>${maven.release.plugin.version}</version>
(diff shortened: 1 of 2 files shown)


ryan-williams:
Actually while working on some of the other PRs I'm filing right now, Scalariform is screwing up my whitespace all over the place, I'm back on wanting to get rid of it :-\

arahuja:
Yea, totally fine with getting rid of it - some kind of lint check might be nice though if there are alternatives

ryan-williams:
Agreed. I'll look around tomorrow.

timodonnell:
Yeah, agree that scaliform is annoying when it disagrees with intellij, but also agree that having some kind of linting enforcing style consistency is really nice. I wonder if there's a way to get it to agree with intellij better, or use some different linter that agrees better?

ryan-williams:
I am theoretically in favor of linting but in practice our codebase does not obey un-disable-able scalariform rules in many places, resulting in a minefield of noise-changes that happen when you compile code. I find it to be a big pain and think we should remove it now and then think about a replacement.

[Here's what it just did to one of my branches](https://github.com/ryan-williams/guacamole/commit/77442c8f2731094f9b96290572508f32fe2a0c9b). I don't want any of those changes but afaik I can't make it stop doing that. The changes boil down to:
* imports: scalariform likes spaces inside each `{` and `}`. I actually slightly prefer this, but IntelliJ's "optimize imports" (⌘⇧O for me, idk if that's the default) can't be told to do it, and that functionality is *way* more important to me than the spacing nit.
* comment spacing: [one space before line comments](https://github.com/ryan-williams/guacamole/commit/77442c8f2731094f9b96290572508f32fe2a0c9b#diff-e5001cba3abbdfa83fc2a4489aa510ecR148), [removes intentional vertical alignment](https://github.com/ryan-williams/guacamole/commit/77442c8f2731094f9b96290572508f32fe2a0c9b#diff-23c169ad54dd125057edd0773d6a335eL41).
* scaladocs: [disagrees with Intellij](https://github.com/ryan-williams/guacamole/commit/77442c8f2731094f9b96290572508f32fe2a0c9b#diff-a5925f655fe562f03e4c20652341418bL163)

I'll look into e.g. [scalastyle](http://www.scalastyle.org/maven.html) next week, but I don't actually believe we're getting any benefit from this today and I'd like to just get rid of it as it's probably cost me tens of minutes in the last few days.

timodonnell:
One kind of draconian approach would be do have travis fail if scalariform
changes anything when it is built. Then everyone just has to run
scalariform before doing a PR (probably good to run `mvn package` before
doing a PR anyway, and that's more convenient now with Ryan's test
speedups), so we no longer have the situation of running scalariform
affecting lots of files your PR had nothing to do with. I wouldn't be
opposed to this solution, but if consensus is better to just drop it then
that's fine with me too

arahuja:
@timodonnell I think the problem with that is when intellij and scalariform keep fighting each other

ryan-williams:
Yea, intellij vs. scalariform is bad. Scalariform also does things that intellij doesn't care about but that I do, and it can't be told to stop, afaik; some of my above examples are in each of these categories.

ryan-williams:
[Here's](https://github.com/ryan-williams/guacamole/commit/118e080e57e285bfe9f4fa1a951b0dbd051463a0) another commit showing scalariform mangling my whitespace

arahuja:
I'd say let's merge this in the meantime? Right now 11 files are re-formatted by scalariform on master, so every `mvn package` will wreck your branch.

ryan-williams:
Yea, that's my bad for getting ahead of this disabling and throwing out scalariform changes in the other PRs I merged in. Sorry about that.

arahuja:
thanks @ryan-williams!

PR #410, opened 2016-04-01
Several LociSet improvements

ryan-williams:
see commit msgs for details

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/410)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/LociSet.scala
+++ b/src/main/scala/org/hammerlab/guacamole/LociSet.scala
@@ -160,22 +160,33 @@ object LociSet {
     }
 
     /**
-     * Parse a loci expression and add it to the builder. Example expressions:
-     *  "all" (all sites on all contigs)
-     *  "none" (no loci, used as a default in some places)
-     *  "chr1,chr3" (all sites on contigs chr1 and chr3)
-     *  "chr1:10000-20000,chr2" (sites x where 10000 <= x < 20000 on chr1, all sites on chr2)
-     */
+      * Parse a loci expression and add it to the builder. Example expressions:
+      *
+      *  "all": all sites on all contigs.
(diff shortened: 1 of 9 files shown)


src/main/scala/org/hammerlab/guacamole/LociSet.scala
```diff
@@ -159,22 +160,33 @@ object LociSet {
...
-     *  "chr1:10000-20000,chr2" (sites x where 10000 <= x < 20000 on chr1, all sites on chr2)
-     */
+      * Parse a loci expression and add it to the builder. Example expressions:
+      *
+      *  "all": all sites on all contigs.
+      *  "none": no loci, used as a default in some places.
+      *  "chr1,chr3": all sites on contigs chr1 and chr3.
+      *  "chr1:10000-20000,chr2": sites x where 10000 <= x < 20000 on chr1, all sites on chr2.
+      *  "chr1:10000": just chr1, position 10000; equivalent to "chr1:10000-10001".
+      *  "chr1:10000-": chr1, from position 10000 to the end of chr1; logically "chr1:10000-∞".
```
timodonnell: nit: it does not include positions after the end of the contig, so it may be confusing to say it's `logically "chr1:10000-∞".` also I'd prefer to avoid special characters in the code (infinity symbol)

timodonnell:
Cool, LGTM module minor nit comment and special characters

arahuja:
LGTM

src/main/scala/org/hammerlab/guacamole/LociSet.scala
ryan-williams: Done.

I disagree that having unicode in comments is a problem but agree that the second clause here probably just makes things less clear.

PR #411, opened 2016-04-01
Two Spark API optimizations

ryan-williams:
Two Spark API nits:
- changed a `reduce` over `Map`s in ReadSet to be a `reduceByKey` over `Long`s; I've observed merging `Map`s in a reduce* to be super slow in Pageant, in the past. It also feels cleaner to use the simpler combiner.
- switched a `keyBy(…).countByKey` to a `map(…).reduceByKey(_+_)`, which should produce fewer intermediate objects

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/411)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/ReadSet.scala
+++ b/src/main/scala/org/hammerlab/guacamole/ReadSet.scala
@@ -72,10 +72,11 @@ case class ReadSet(
       sequenceDictionary.get.records.foreach(record => builder += ((record.name.toString, record.length)))
       builder.result
     } else {
-      mappedReads.map(read => Map(read.referenceContig -> read.end)).reduce((map1, map2) => {
-        val keys = map1.keySet.union(map2.keySet).toSeq
-        keys.map(key => key -> math.max(map1.getOrElse(key, 0L), map2.getOrElse(key, 0L))).toMap
-      }).toMap
+      mappedReads
+        .map(read => read.referenceContig → read.end)
+        .reduceByKey(math.max)
+        .collectAsMap()
+        .toMap
     }
   }
 }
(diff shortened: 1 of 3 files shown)


timodonnell:
I'm not sure about the advantage for having `PerSample` be a `Seq` instead of an `IndexedSeq`. In the joint caller attempting to run on ~20 BAMs I was observing tasks spending time indexing into a `List`-backed `PerSample` (since we do lots of random access into the group of pileups, e.g. look at the normal pileup, now the tumor rna, etc.). Having it be an `IndexedSeq` seems like a reasonable way to make sure we don't accidentally make that slow.

Modulo that point LGMT . Thanks @ryan-williams

arahuja:
LGTM

> switched a keyBy(…).countByKey to a map(…).countByValue, which should produce fewer intermediate objects

Would love to know about this - how come?

ryan-williams:
@timodonnell thx for explanation about `PerSample`, I've switched it back to the way it was.

@arahuja I think you're right, I messed up what I was trying to do here. I just pushed an improved version. My thoughts:

Previously, we had a `keyBy`, which [first did a `map`](https://github.com/apache/spark/blob/v1.6.0/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L1478):

```
before: vaf
after:  (vaf, bin)
```

then we had countByKey, which [does a `mapValues`](https://github.com/apache/spark/blob/v1.6.0/core/src/main/scala/org/apache/spark/rdd/PairRDDFunctions.scala#L381):

```
before: (vaf, bin)
after:  (vaf, 1L)
```

then a `reduceByKey`.

I'd changed it to do a `.map`:

```
before: vaf
after:  bin
```

then a `countByValue`, which added a `null` value:

```
before: bin
after:  (bin, null)
```

followed by a countByKey, which maps the values to 1:

```
before: (bin, null)
after:  (bin, 1L)
```

before reducing by key, again.

What I meant to do, and have pushed to this PR now, is `map` directly to `(bin, 1L)` and then `reduceByKey`.

So the difference is between

```
vaf → (bin, vaf) → (bin, 1L) → reduceByKey
```

and

```
vaf → (bin, 1L) → reduceByKey
```

In reality, this would just get collapsed within a stage, and the longer path probably JIT'd to a point where any perf difference might be negligible, but it seems like the latter way must be at least as good, and it feels cleaner to me.

Thanks for probing, and lmk if that makes sense!

PR #401, opened 2016-03-26
Remove unused reference argument in read loading

arahuja:
After #398 the various `loadRead..` functions still took the reference as an argument but it was unused. I removed that parameter - unless you had other plans for it @timodonnell?

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/401)

PR #412, opened 2016-04-01
fix VariantSupportSuite; tests were not being run

ryan-williams:
iterator.size was being called before iterating over an iterator to check its elements; after size() the iterator is always empty, so the subsequent checks were never being run.

The "expected" data in several test cases was wrong, and I couldn't figure out what they might have originally been intended to point at in the BAM used in this file. I picked loci that exhibited similar behavior to what the test cases seemed to be intending to test, from nearby loci where possible, and fixed the cases up.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/412)

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/commands/VariantSupportSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/commands/VariantSupportSuite.scala
@@ -17,7 +17,7 @@ class VariantSupportSuite extends GuacFunSuite with Matchers with TableDrivenPro
   def testAlleleCounts(window: SlidingWindow[MappedRead],
                        variantAlleleLoci: (String, Int, Seq[(String, String, Int)])*) = {
     for {
-      (contig, locus, alleleCounts) ← variantAlleleLoci
+      (contig, locus, alleleCounts) <- variantAlleleLoci
     } {
       withClue(s"$contig:$locus") {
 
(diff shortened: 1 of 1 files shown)


timodonnell:
Cool, thanks for debugging this @ryan-williams !

LGTM modulo use of arrow characters

src/test/scala/org/hammerlab/guacamole/commands/VariantSupportSuite.scala
```diff
@@ -14,99 +15,111 @@ class VariantSupportSuite extends GuacFunSuite with Matchers with TableDrivenPro
...
+  def nonDuplicateGatkReads(loci: String) =
+    TestUtil.loadReads(
+      sc,
+      "gatk_mini_bundle_extract.bam",
+      Read.InputFilters(
+        mapped = true,
+        nonDuplicate = true,
+        overlapsLoci = Some(LociSet.parse(loci))
+      ),
+      reference = grch37Reference
```
arahuja: I thin this will break now on merge?

arahuja:
Great catch!

src/test/scala/org/hammerlab/guacamole/commands/VariantSupportSuite.scala
ryan-williams: #401 didn't remove this `reference`; #415 does but it's not merged yet, so I'll leave it here and resolve conflicts as they arise

PR #413, opened 2016-04-01
restrict GermlineAssembly test loci parsing

ryan-williams:
This speeds up this test (and the total time to run all Guac tests) by a lot.

This PR takes this suite from several minutes down to 45s.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/413)

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCallerSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCallerSuite.scala
@@ -7,9 +7,9 @@ import org.hammerlab.guacamole.reads.Read
 import org.hammerlab.guacamole.reference.ReferenceBroadcast
 import org.hammerlab.guacamole.util.TestUtil
 import org.hammerlab.guacamole.variants.CalledAllele
-import org.scalatest.{ BeforeAndAfter, FunSuite, Matchers }
+import org.scalatest.{BeforeAndAfterAll, FunSuite, Matchers}
 
-class GermlineAssemblyCallerSuite extends FunSuite with Matchers with BeforeAndAfter {
+class GermlineAssemblyCallerSuite extends FunSuite with Matchers with BeforeAndAfterAll {
 
   val args = new Arguments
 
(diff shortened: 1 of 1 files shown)


timodonnell:
Leaving this one for @arahuja

src/test/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCallerSuite.scala
```diff
@@ -21,188 +21,138 @@ class GermlineAssemblyCallerSuite extends FunSuite with Matchers with BeforeAndA
...
-    variant.start should be(1303426)
-    Bases.basesToString(variant.allele.refBases) should be("ACT")
-    Bases.basesToString(variant.allele.altBases) should be("A")
-  }
-
-  test("test assembly caller: illumina platinum tests; homozygous insertion") {
-
-    val loci = "chr1:1321298-1321299"
-    val variants = discoverGenotypesAtLoci(loci)
+  def testFn(name: String,
```
arahuja: nit: since this specific to this Suite, it's clear what it's doing but maybe 'test/assertVariant` or something?

src/test/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCallerSuite.scala
```diff
@@ -21,188 +21,138 @@ class GermlineAssemblyCallerSuite extends FunSuite with Matchers with BeforeAndA
   var reference: ReferenceBroadcast = _
   var readSet: ReadSet = _
 
-  before {
+  override def beforeAll() {
```
arahuja: q: I had thought this was what the before was doing, but perhaps I misinterpreted, OOC what's the difference here? and the effect of the override?

arahuja:
Oh sorry, I remembered why I left out the overlaps loci from the read loading. It would throw away reads that are in the window, but do not overlap the locus. I wanted to have all the reads in the window loaded as well, but call variants/assemble at specific locus

We can adjust `testFn` to expand the loading window based on `snvWindowRange`?

src/test/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCallerSuite.scala
ryan-williams: `before` and `after` (defined in [`BeforeAndAfter`](https://github.com/scalatest/scalatest/blob/release-2.2.1-for-scala-2.11-and-2.10/src/main/scala/org/scalatest/BeforeAndAfter.scala#L21)) run for each "test".

`beforeAll` and `afterAll` (cf. [`BeforeAndAfterAll`](https://github.com/scalatest/scalatest/blob/release-2.2.1-for-scala-2.11-and-2.10/src/main/scala/org/scalatest/BeforeAndAfterAll.scala#L19)) run for each *suite*.

The definitions are different for whatever reason so I can't just say

```
beforeAll {
  …
}
```

but have to use the `override def beforeAll() { … }` you see here.

ryan-williams:
> We can adjust testFn to expand the loading window based on snvWindowRange?

good call, done

src/test/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCallerSuite.scala
ryan-williams: good call, renamed to `verifyVariantsAtLocus`

arahuja:
LGTM, I've got some more tests to add with some updates, so I'll merge this is in! Thanks! This will be very useful

arahuja:
@ryan-williams seems like I can't run individual tests anymore in intellij after this change - are you able to do that?

ryan-williams:
yea, the scalatest integration looks for calls to `test(…) {…}` when you use ⌃⇧F9 to debug the current test case, so after this that shortcut will pick up the whole suite.

If you want to run a specific test, you'll have to make a test configuration for it: 
* open configurations (⌥⇧⌘C)
* "+" > "ScalaTest" > make a "Test name" configuration with the desired suite and test, e.g.:

  

* "OK"
* ⇧F9/⇧F10 to debug/run current configuration.

Losing ⌃⇧F9 / ⌃⇧F10 is annoying, but the configuration steps above are a one-time cost for each test case that you want to be able to run individually, which I find is typically O(1) given O(n) test cases, because after a few individual ones work you typically want to run the whole suite.

If this is too great a cost we can work the `verifyVariantsAtLocus` test-case shorthand I made here to be more like:

```
test(…) { verifyVariantsAtLocus(…) }
```

for each case. Let me know if you want me to send a PR doing that.

arahuja:
> If this is too great a cost we can work the verifyVariantsAtLocus test-case shorthand I made here to be more like ...

Yea, for me, I often make a test case when debugging any specific issue, and these might be changing as I'm investigating an issue so being able to right-click run/debug is really useful. I can change it to the 
`test(…) { verifyVariantsAtLocus(…) }` syntax

ryan-williams:
Yea, the change would be so minor to re-enable it that I think it's worth doing. Happy to do it if you want, lmk.

PR #414, opened 2016-04-01
add optional loci filtering to TestUtil.loadPileup

ryan-williams:
brings ReadSubsequenceSuite from ~2mins down to about 1min

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/414)

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/util/TestUtil.scala
+++ b/src/test/scala/org/hammerlab/guacamole/util/TestUtil.scala
@@ -26,10 +26,11 @@ import com.twitter.chill.{ IKryoRegistrar, KryoInstantiator, KryoPool }
 import org.apache.commons.io.FileUtils
 import org.apache.spark.SparkContext
 import org.hammerlab.guacamole.pileup.Pileup
+import org.hammerlab.guacamole.reads.Read.InputFilters
 import org.hammerlab.guacamole.reads._
 import org.hammerlab.guacamole.reference.ReferenceBroadcast.MapBackedReferenceSequence
 import org.hammerlab.guacamole.reference.{ ContigSequence, ReferenceBroadcast }
-import org.hammerlab.guacamole.{ Bases, GuacamoleKryoRegistrator, ReadSet }
+import org.hammerlab.guacamole.{ Bases, GuacamoleKryoRegistrator, LociSet, ReadSet }
 import org.scalatest._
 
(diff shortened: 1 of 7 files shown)


src/test/scala/org/hammerlab/guacamole/util/TestUtil.scala
```diff
@@ -231,11 +232,30 @@ object TestUtil extends Matchers {
       Pileup(normalReads, contig, locus, reference.getContig(contig)))
   }
 
-  def loadPileup(sc: SparkContext, filename: String, reference: ReferenceBroadcast, locus: Long = 0, contig: Option[String] = None): Pileup = {
-    val records = TestUtil.loadReads(sc, filename, reference = reference).mappedReads
+  def loadPileup(sc: SparkContext,
+                 filename: String,
+                 reference: ReferenceBroadcast,
+                 locus: Long = 0,
+                 contigOpt: Option[String] = None): Pileup = {
```
timodonnell: nit: `Opt` is an abbreviation (I feel kind of awful being this nit picky but abbreviation creep is so easy to happen and we need to be steadfast)

timodonnell:
LGTM, good idea @ryan-williams !

src/test/scala/org/hammerlab/guacamole/util/TestUtil.scala
ryan-williams: I like the `Opt` suffix on `Option`s. Would you prefer I name this `contigOption` or just `contig` The former feels unnecessarily verbose to me, and the latter is awkward when you want to do anything with the Option:

```
contig.map(contig => …)
```

I'll make it `contigOption` if you prefer.

ryan-williams: Also `Option`s and the naming patterns here are so basic that I don't think this is a slippery slope to other inscrutable/abbreviated variable names, if that is your concern.

timodonnell: Yeah I might be being overly pedantic here. I've used the convention
`maybeContig` elsewhere, but really anything here you think is best works
for me

ryan-williams: cool, renamed to `maybeContig`

PR #415, opened 2016-04-01
remove unused reference parameters

ryan-williams:
turns out there were many of these laying around

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/415)


ryan-williams:
heh I just remembered this overlaps with #401. I think this one does more though. I'm happy to rebase this on that though if we merge it first.

timodonnell:
Delaying review of this one for the reason you noted

ryan-williams:
I updated this one so it's ready for review.

timodonnell:
LGTM

PR #416, opened 2016-04-01
use sameElements check in ReadSetSuite

ryan-williams:
this test was doing some convoluted things to emit a custom error msg when the expected condition didn't hold.

i reworked it to use `withClue` and `assert` in more idiomatic ways, with the help of a `LazyMessage` object whose `toString` (which is expensive to compute) is only evaluated if the test fails.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/416)

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
@@ -108,7 +108,7 @@ class ReadSetSuite extends GuacFunSuite with Matchers {
     }
     new AlignmentRecordRDDFunctions(adamRecords.rdd).saveAsParquet(args, adamRecords.sequences, adamRecords.recordGroups)
 
-    val (allReads, _) = Read.loadReadRDDAndSequenceDictionaryFromADAM(adamOut, sc, token = 0, reference = chr22Fasta)
+    val (allReads, _) = Read.loadReadRDDAndSequenceDictionaryFromADAM(adamOut, sc, token = 0)
     allReads.count() should be(8)
     val collectedReads = allReads.collect()
 
@@ -116,8 +116,7 @@ class ReadSetSuite extends GuacFunSuite with Matchers {
       adamOut,
       sc,
       1,
-      Read.InputFilters(mapped = true, nonDuplicate = true),
-      reference = chr22Fasta)
+      Read.InputFilters(mapped = true, nonDuplicate = true))
     filteredReads.count() should be(3)
     filteredReads.collect().forall(_.token == 1) should be(true)
   }
--- a/src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
@@ -20,36 +20,75 @@ package org.hammerlab.guacamole.reads
 
 import org.apache.parquet.hadoop.metadata.CompressionCodecName
 import org.bdgenomics.adam.rdd.read.AlignmentRecordRDDFunctions
-import org.bdgenomics.adam.rdd.{ ADAMSaveAnyArgs, ADAMContext }
-import org.hammerlab.guacamole.reference.ReferenceBroadcast
-import org.hammerlab.guacamole.{ LociSet }
+import org.bdgenomics.adam.rdd.{ADAMContext, ADAMSaveAnyArgs}
+import org.hammerlab.guacamole.LociSet
 import org.hammerlab.guacamole.reads.Read.InputFilters
-import org.hammerlab.guacamole.util.{ TestUtil, GuacFunSuite }
+import org.hammerlab.guacamole.reference.ReferenceBroadcast
+import org.hammerlab.guacamole.util.{GuacFunSuite, TestUtil}
 import org.scalatest.Matchers
 
 class ReadSetSuite extends GuacFunSuite with Matchers {
 
   def chr22Fasta = ReferenceBroadcast.readFasta(TestUtil.testDataPath("chr22.fa.gz"), sc)
 
+  case class LazyMessage(msg: () => String) {
+    override def toString: String = msg()
+  }
+
   sparkTest("using different bam reading APIs on sam/bam files should give identical results") {
     def check(paths: Seq[String], filter: InputFilters): Unit = {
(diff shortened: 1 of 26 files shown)


src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
```diff
@@ -20,36 +20,75 @@ package org.hammerlab.guacamole.reads
...
+          TestUtil.loadReads(
+            sc,
+            paths.head,
+            filter,
+            config = configs.head,
+            reference = chr22Fasta
+          ).reads.collect
+
+        for {
+          config ← configs
```
timodonnell: Is that a funky ascii arrow character or is github just being weird?

timodonnell:
Thanks @ryan-williams . To me, the new version with the lazy message is actually a bit harder to follow than the original version, which is ugly but straightforward, but that may be because I am not very idiomatic in scala. If @arahuja 's vote is that the new version is better, then I'd be ok to have this merged

timodonnell:
Also sorry for the annoying nit, but I'd prefer to not use the arrow symbols in our codebase (assuming this is not just a github rendering thing)

src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
ryan-williams: that is a non-ascii arrow; I recently set Intellij to automatically collapse `->`, `<-`, and `=>` to `→`, `←`, and `⇒`, which I like a fair amount, but if you/others hate it I can stop. 

It makes the code a lot cleaner, to me; these are logically one symbol, so displaying them with two otherwise-unrelated punctuation marks feels worse now that I've tried it this way.

ryan-williams:
I had a hard time figuring out what the original version was doing, and felt like the `assert(false, "shouldn't get here")` was a red flag for how it was arranged.

There are 3 logical changes here:

1. whitespace added between some of the variable declarations
2. existing double-`.foreach` is now a `for` comprehension that skips the tautological first case where a {path,config} is compared to itself.
  * this is the most expensive test in the codebase atm, taking about half the total test time in my local runs, so optimizing to skip the first case feels worthwhile, and doing so with `for` syntax seems much cleaner than the alternative.
3. rearranging the `if (a != b) { <print errors>; assert(false) }` to use scalatest `assert(a == b, <error>)`, using a wrapper for `<error>` that lazily evaluates.

It sounds like the third item is what you are not in favor of?

timodonnell:
Woah I had no idea this was such a performance improvement, that makes me more in favor of it. I think my preference would be to avoid change number 3, as you mentioned, and just do the boring straightforward thing but I see your point and trust your judgement

src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
timodonnell: Yeah, gotcha. Just an asthetic / personal preference thing, but I feel weirdly strongly that we should avoid special characters in the codebase

ryan-williams:
> Woah I had no idea this was such a performance improvement

With all my PRs locally, tests finish in 1-2mins, so this test being half of that isn't necessarily the worst thing.

I'm also not observing the speedup I theorized I'd see thanks to that optimization, when running locally right now, so take this all with a grain of salt.

> my preference would be to avoid change number 3

I definitely prefer using an `assert(a == b)` to the `if (a != b) { <print error>; assert(false) }` but I'll let @arahuja or @e5c weigh in on how they feel about it. It's not a big deal either way.

FWIW / general PSA: I find most PRs, including this one, impossible to grok in Github; it might be cleaner to look at it in Intellij if you've not already.

arahuja:
I don't think I understand the test you are debating - the reads should always be the same? and then if not we print the differences? In the original, I think the asserts are confusing since we should already not get there, so I think asserting on the reads being the same makes sense.

What does `LazyMessage` add? Isn't the string being constructed there anyways? Or does that restrict it to only happen on failure?

ryan-williams:
> the reads should always be the same?

Correct.

> What does LazyMessage add? Isn't the string being constructed there anyways? Or does that restrict it to only happen on failure?

Yea, it turns out that constructing the `missing` and `extra` sets is really time-consuming (10's of seconds, IIRC?), so this is a way of doing it only in the failure case, while still using a standard `assert` flow.

src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
ryan-williams: I think you should reconsider. scalac/zinc, Intellij, ([even scalariform!](https://github.com/scala-ide/scalariform#rewritearrowsymbols)), all support this, and I think it's a clear improvement.

You never mean "dash, greater than", "less than, dash", or "equals, greater than", and scala doesn't understand them that way; you mean "right arrow", "left arrow", and "double right arrow".

I thought it was a weird idea but after using it the old way seems broken. No need to ASCII-art our code where the entire toolchain and language idioms have freed us from having to. You don't have to do anything different, either, intellij rewrites it for you.

You can also skip the setting, but I'd like to keep it on. lmk what you think.

timodonnell: These symbols seem like a complexity cost without much benefit to me. They
will look odd to someone new to scala (even someone not so new, like me),
render weirdly on github (the => character looks like a -> to me on my
screen), may be a hassle for non-intellij manipulation of the codebase like
grep, and will likely result in inconsistency where in some parts of the
code we use the symbols and in other parts of the code we do not. They also
evoke the horrors of APL. There are lots of other non-single-character
syntax tokens like 'val', so I don't see a problem with having '->' spelled
out as two characters.

ryan-williams: > They will look odd to someone new to scala (even someone not so new, like me)

Seems to me they would look about as weird to someone new to scala as their ascii-art equivalents; they look weird to you because you're not new to Scala and are used to the old way.

> There are lots of other non-single-character syntax tokens like 'val', so I don't see a problem

If there were a symbol that the entire toolchain was supporting instead of `val` I would be interested in using it / allowing it in the codebase.

In any case I removed them here and force-pushed.

PR #417, opened 2016-04-01
decouple Reference sequences from Spark-broadcasts

ryan-williams:
only allowing instantiation of (and access to) References via Spark Broadcast objects was resulting in abstraction creep.

lots of code touching Spark internals that didn't need to be, and lots of tests being heavier-weight spark tests.

Here I've:
- flattened `ContigSequence` into simply a type alias for `Array[Byte]`
- replaced `ReferenceBroadcast` with `ReferenceGenome`; the latter just wraps a `Map[String, ContigSequence]`.
- changed the {pileup,window}Flatmap* interfaces to take a reference and hide the broadcasting; the multipleRDDs version now takes a UDF that expects pileups and a reference.
- pulled SparkContexts out of many tests and parts of the code that shouldn't have had one

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/417)


src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
```diff
@@ -312,19 +321,24 @@ object DistributedUtil extends Logging {
     skipEmpty: Boolean,
     function: (Pileup, Pileup) => Iterator[T],
     reference: ReferenceGenome): RDD[T] = {
+
+    val referenceBC = reads1.context.broadcast(reference)
```
timodonnell: nit: use of abbreviation

timodonnell: Does this re-broadcast the reference genome every time we call `pileupFlatMapTwoRDDs`? I think the old implementation would avoid that?

src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
```diff
@@ -336,8 +350,11 @@ object DistributedUtil extends Logging {
     readsRDDs: PerSample[RDD[MappedRead]],
     lociPartitions: LociMap[Long],
     skipEmpty: Boolean,
-    function: PerSample[Pileup] => Iterator[T],
+    function: (PerSample[Pileup], ContigSequence) => Iterator[T],
     reference: ReferenceGenome): RDD[T] = {
+
+    val referenceBC = readsRDDs.head.context.broadcast(reference)
```
timodonnell: Since it's a recurring pattern, I won't mention it after this time, but I think we probably don't want to re-broadcast the reference every time we call a pileup function (unless Spark has an optimization for this that I don't know about)

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -79,14 +79,13 @@ object GermlineAssemblyCaller {
...
      * @param endLocus End (exclusive) locus on the reference
-     * @param kmerSize Length of the subsequence
      * @param minOccurrence Mininum number of times a subsequence needs to appear to be included
      * @return List of subsequences overlapping [startLocus, endLocus) that appear at least `minOccurrence` time
      */
     def getConsensusKmer(reads: Seq[MappedRead],
                          startLocus: Int,
                          endLocus: Int,
-                         minOccurrence: Int): Iterable[Array[Byte]] = {
+                         minOccurrence: Int): Iterable[ContigSequence] = {
```
timodonnell: I'm not sure it makes sense to change `Array[Byte]` types to `ContigSequence` unless they are truly the sequence of a contig, i.e. they are indexed by genomic locus. Looking at this it would appear to me this function is returning an entire contig's sequence

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -171,8 +172,8 @@ object GermlineAssemblyCaller {
 
       // Build a variant using the current offset and read evidence
       def buildVariant(referenceOffset: Int,
-                       referenceBases: Array[Byte],
```
timodonnell: Also a recurring pattern, so I won't mention it more, but unless the sequence is actually indexed by genomic locus I think it adds confusion to call it a `ContigSequence`

src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
```diff
@@ -336,8 +350,11 @@ object DistributedUtil extends Logging {
     readsRDDs: PerSample[RDD[MappedRead]],
     lociPartitions: LociMap[Long],
     skipEmpty: Boolean,
-    function: PerSample[Pileup] => Iterator[T],
+    function: (PerSample[Pileup], ContigSequence) => Iterator[T],
```
timodonnell: Since each `Pileup` has a `ContigSequence` attribute, the user defined function already has access to the contig sequence, so I'm not sure it makes sense to add this argument

src/main/scala/org/hammerlab/guacamole/reference/MapBackedReferenceSequence.scala
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * A ContigSequence implementation that uses a Map to store only a subset of bases. This is what you get if you load
+ * a "partial fasta".
+ *
+ * @param wrapped
+ */
+case class MapBackedReferenceSequence(length: Int, wrapped: Map[Int, Byte]) extends ContigSequence {
+  def apply(index: Int): Byte = wrapped.getOrElse(index, Bases.N)
+
+  override def iterator: Iterator[Byte] = {
```
timodonnell: I think this may be a confusing method to have implemented since it will iterate only through the bases that are in the map, right? So the user would have no idea what site each base corresponds to. I think either need to start at 0 or just throw an error here

src/main/scala/org/hammerlab/guacamole/reference/ReferenceGenome.scala
```diff
@@ -1,52 +1,147 @@
...
-  def buildMdTag(readSequence: String, referenceContig: String, referenceStart: Int, cigar: Cigar): String = {
-    val referenceSequence =
-      Bases.basesToString(
-        getReferenceSequence(referenceContig, referenceStart, referenceStart + cigar.getReferenceLength)
-      )
-    MdTag(readSequence, referenceSequence, cigar, referenceStart).toString
+  def apply(fastaPath: String, partialFasta: Boolean = false): ReferenceGenome = {
+    cache.getOrElseUpdate(
+      fastaPath,
+      if (partialFasta)
```
timodonnell: I don't know scala well enough to be totally sure this second argument will be lazily evaluated. I'm assuming it will be but just want to point it out in case there's some special syntax that might need to be used to get that behavior

src/main/scala/org/hammerlab/guacamole/reference/package.scala
```diff
@@ -0,0 +1,5 @@
+package org.hammerlab.guacamole
+
+package object reference {
+  type ContigSequence = Seq[Byte]
```
timodonnell: I think this is reasonable, but another option would be to just make it a trait and spell out that it implements `apply`, `slice`, and `length`. To me the latter would be a little clearer but I'm not super opposed to this solution (although if you do stick with this implementation maybe make it an `IndexedSeq`? I think it wouldn't matter as the code is now but I wouldn't want it to ever be possible to end up with a reference genome as a `List`)

timodonnell:
Thanks @ryan-williams . I made a few comments and I think there are some things to sort out here before merging but I like the overall idea of cleaning up this pretty messy part of the code

src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
ryan-williams: > Does this re-broadcast the reference genome every time we call pileupFlatMapTwoRDDs?

I guess it would, but we currently call this function (or its counterparts) once in each variant caller, and that doesn't seem super likely to change afaict?

This is currently called once in the codebase, by SomaticStandardCaller, and twice in DistributedUtilSuite, so this doesn't seem like a problem to me.

We could make this a wrapper for a version of this method that takes a `Broadcast[ReferenceGenome]` instead of a `ReferenceGenome`; let me know if you prefer that. I think it cleaner to not force that on callers if we don't have to, personally.

src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
ryan-williams: The pileup function we're in here is `pileupFlatMapMultipleRDDs`, which is called once in the codebase, by `SomaticJointCaller`, which calls it once per Spark app run (it's also called three times in DistributedUtilSuite).

Per [my comment above](https://github.com/hammerlab/guacamole/pull/417/files#r58210739), I don't agree that this is a problem, but if you are concerned with optimizing those tests or for future callers that might make two pileupFlatmap passes over a dataset in one run, we can discuss making this a wrapper for a version that takes a Broadcast as a parameter.

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
ryan-williams: OK, I was using `ContigSequence` to denote any *sub*sequence of a contig, just as a way to add some semantics to the otherwise somewhat opaque `Iterable[Array[Byte]]` here, but I understand that prior to this PR `ContigSequence` meant whole contigs, so I can revert that tagging here and the few other similar places.

src/main/scala/org/hammerlab/guacamole/reference/ReferenceGenome.scala
ryan-williams: Good question! Scala saves the day:

```
scala> import scala.collection.mutable
import scala.collection.mutable

scala> val cache = mutable.HashMap.empty[String, Int]
cache: scala.collection.mutable.HashMap[String,Int] = Map()

scala> cache.getOrElseUpdate("foo", { println("yay"); 4 })
yay
res0: Int = 4

scala> cache.getOrElseUpdate("foo", { println("yay"); 4 })
res1: Int = 4
```

[Relevant code pointer](https://github.com/scala/scala/blob/v2.10.4/src/library/scala/collection/mutable/MapLike.scala#L186).

src/main/scala/org/hammerlab/guacamole/reference/package.scala
ryan-williams: I like the idea of making it an alias for `IndexedSeq`! Wrapping it in another class (or trait, and then subclassing that) feels like added overhead for no benefit.

src/main/scala/org/hammerlab/guacamole/reference/MapBackedReferenceSequence.scala
ryan-williams: good point, throwing an error instead

src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
ryan-williams: (I expanded the abbreviation)

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
ryan-williams: Now that `ContigSequence` is an alias for `IndexedSeq`, I am more in favor of having it (or an alias like it) here instead of `Array[Byte]`, as a way of enforcing that genomic sequences that we index in to are efficiently indexable.

Do you want me to rename it to just `Sequence`? `GenomicSequence`? Something else? Leave it as `ContigSequence`?

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
ryan-williams: Yea I think this folds in to the `ContigSequence` discussion [above](https://github.com/hammerlab/guacamole/pull/417#discussion_r58218719).

src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
ryan-williams: good call! that simplifies things, fixed

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
timodonnell: Yeah, I think we should keep `ContigSequence` to have only the narrow use as an entire contig indexed by genomic locus. Sequences of bytes that are not entire contigs should be something else. I think keeping them as `Array[Byte]` or `IndexedSeq[Byte]` is the most clear but would not be strongly opposed to introducing a new type (maybe `Bases`?) for them

src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
timodonnell: Yeah, even though we don't currently map over pileups more than once in any callers, I think that is a case that will come up eventually and we should not make it require broadcasting the reference genome each time. Since these pileupFlatmap functions are already taking RDD's and thus are already using Spark infra, maybe best to just have them take a`Broadcast[ReferenceGenome]`?

ryan-williams: I don't like removing the Broadcast-less interface, adding marginal complexity to all callers in service of an imagined future use case.

I can add overloaded versions, but that'll be a lot of DistributedUtil boilerplate. I'll try it and see how it looks though.

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
ryan-williams: I made `ContigSequence` and `ContigSubsequence` and used them in respective appropriate places.

src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala
ryan-williams: I added overloaded versions of the pileupFlatmap functions that take a `Broadcast[ReferenceGenome]` directly.

ryan-williams:
Last [comment](https://github.com/hammerlab/guacamole/pull/417#discussion_r58543442) and [commit](https://github.com/hammerlab/guacamole/pull/417/commits/b02a31c05772cc8ebc94ae23cbf8ded982525f06) address the only outstanding issue afaict @timodonnell.

arahuja:
@ryan-williams @timodonnell Sorry, I didn't look at this earlier, but it's important to note that this may also be a performance change. One of the reasons for the previous implementations was tasks that only needed a single contig did not need to deserialize the entire genome, but only a single contig. This may not be much, but now a task will have to request the full genome. One of the thoughts previously was we could further subdivide the genome to make the access more directed.

ryan-williams:
That's an interesting point; I didn't realize that the per-contig `Broadcast`s were intentional. This approach (especially it's extrapolation to finer chunks) feels like a weird pattern: pushing what is logically partitioning logic into Broadcast's lazy-fetch machinery. I'll have to think about the implications of that.

arahuja:
> feels like a weird pattern: pushing what is logically partitioning logic into Broadcast's lazy-fetch machinery. I'll have to think about the implications of that.

Yes, I could see that - I'm not opposed to this change, just wanted to point it out at least. I think the ideal solution is to distribute exactly the piece of reference needed for each task. If we had that, there is no need for the broadcast at all, as the 100-200 bytes needed for that reference subsequence could be shipped with the task. The disadvantage is that tasks overlap in the what reference region they process, duplicating the data sent, hence the broadcast. But since we know that we should be able to find a middle ground and seems like broadcasting chunks would do that.

ryan-williams:
> The disadvantage is that tasks overlap in the what reference region they process, duplicating the data sent, hence the broadcast.

It would only result in [duplication that a super-fine-grained Broadcast would save us from] when [tasks that overlap each other] happen to be scheduled on the same executor, but that's not something we should really bank on or optimize for afaict.

In any case, the per-contig broadcast today results in far more duplication of reference material transmitted than a proper reference-partitioning would, since each task gets the full reference contig for the region it is processing.

I'll look in to "distributing exactly the piece of reference needed for each task", that sounds ultimately like the right solution here.

ryan-williams:
Closing this temporarily until I have a chance to experiment further with reference-plumbing strategies.

PR #419, opened 2016-04-01
make integration tests a standalone app

ryan-williams:
if resurrected at some later date, they should probably be wrapped in
a main() and run from login shells where a human will read their
output.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/419)


timodonnell:
Can we leave these marked as ignore but not delete them? I run them occasionally to assess accuracy

ryan-williams:
I've restored them and made them an app with a main(). I think that's preferable to leaving them in as `@Ignore`d tests (which e.g. show up when running `mvn test`). They're not really unit or integration tests, as written, but a standalone app that a human can run and read some output about, so I think this makes more sense for now.

The `if (true)` and `if (false)` statements should become cmdline args, perhaps; kind of weird to have those in trunk, but that's for another PR.

timodonnell:
Cool, LGTM. Thanks @ryan-williams !

PR #421, opened 2016-04-03
use contigLengths instead of readset as param

ryan-williams:
feels weird to have to pass a whole `ReadSet` to parse some loci, and we were only using it for the contig lengths

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/421)


timodonnell:
LGTM

PR #422, opened 2016-04-03
remove unused read "token" field on reads

ryan-williams:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/422)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -184,25 +184,6 @@
         </executions>
       </plugin>
       <plugin>
-        <groupId>org.scalariform</groupId>
-        <artifactId>scalariform-maven-plugin</artifactId>
-        <version>0.1.4</version>
-        <executions>
-          <execution>
-            <phase>process-sources</phase>
-            <goals>
-              <goal>format</goal>
-            </goals>
-            <configuration>
-              <alignParameters>true</alignParameters>
-              <alignSingleLineCaseStatements>true</alignSingleLineCaseStatements>
-              <doubleIndentClassDeclaration>true</doubleIndentClassDeclaration>
-              <preserveDanglingCloseParenthesis>true</preserveDanglingCloseParenthesis>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
         <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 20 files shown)


timodonnell:
Yeah, I guess the original need for the `token` field has been obviated by having support for flatmapping over multiple pileups at once (instead of having to take a union of them and then splitting the pileup by token).

LGTM

ryan-williams:
Yea, we've basically [duplicated this field outside the read and pushed it through the SplitIterator machinery](https://github.com/hammerlab/guacamole/blob/b254425236df5d52fb0f451613d50967b8ec843a/src/main/scala/org/hammerlab/guacamole/DistributedUtil.scala#L620)… that said, I could an argument for putting this info in the Read, in which case I would just vote for renaming it to e.g. `sampleID`.

Any thoughts @arahuja / @e5c ?

arahuja:
I think two things work here 1) dropping token/sampleID 2) Dropping sampleName and replacing with sampleID and a broadcast variable tracked somewhere of sampleID->sampleName.

Keeping the sampleName on every read is not very useful, it's only useful on read in and writing outputs, rarely in a per-read operation.

ryan-williams:
Oof I didn't even think about the sample name per read!

Agreed that a sample ID -> name broadcast makes sense.

Let me see about folding that in here too.

arahuja:
> Agreed that a sample ID -> name broadcast makes sense.

Thinking about this (and the other reference broadcast PR), it's tiny data, so these could also be arguments to any task that needs them, and we can thread them through `pileupFlatMap`? But it'd be nice to not have them in every read.

ryan-williams:
Per offline chat, I think plan of record is to just drop `token` here since it's actually never used, and we can debate whether to do anything about `sampleName` later.

arahuja:
@ryan-williams Looks like this merged? But the PR is still open?

ryan-williams:
Very confusing; it does seem to be merged:



but this is open and says it has merge conflicts.

I think I was trying to merge this PR the other day while github was having downtime; something must have gone wrong. Anyway I'll close this since it does seem to be in HEAD already.

PR #423, opened 2016-04-04
remove unused PairedRead, PairedMappedRead

ryan-williams:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/423)


timodonnell:
I like the idea of getting rid of these classes, but I think then we need to add the mate properties in the `Read` and `MappedRead` classes. E.g. it should be possible to see if a given `MappedRead` has its mate mapped, and if so to what position, etc. Even if no callers currently use this I will likely need it for phasing in the joint caller soon

ryan-williams:
Maybe we should just leave them in then.

Issue #329, opened 2015-07-16
Move back to multi-module project


arahuja:
I wanted to discuss moving Guacamole back to a multi-module project. There are two main reasons I'd like to do this
1. Separate out Spark and non-Spark components, it'd be nice to have the commands and `DistributedUtil` separated from the other more general code, that code can be used in other project then as well.
2. Setting up some Avro/serialization structures.  It may be nice to have easy Avro/Parquet serialization for some of our outputs as opposed to CSVs.

Interested to hear thoughts before though cc @timodonnell @ryan-williams @danvk 

Also, while doing this, maybe it's worth going to an SBT build, @ryan-williams would you recommend this?

ryan-williams:
otomh I don't have a strong desire to switch to SBT and stop supporting Maven. supporting both also seems reasonable but I also don't have a strong desire to make that happen either. willing to hear from ppl who might feel strongly about either.

arahuja:
OK, I have no strong feelings on SBT either, other than it was difficult to find a coveralls tool to work with maven for #244. Mostly, I was just thinking if we were taking this on might be nice to do both at the same time.

timodonnell:
If you think this is a good idea, Arun, I'm not opposed

PR #427, opened 2016-04-06
split DistributedUtil into new `distributed` package

ryan-williams:
* pileup flatmap functions → `PileupFlatMapUtils`
* window flatmap functions → `WindowFlatMapUtils`
* loci partitioning functions → `LociPartitioningUtils`
* other miscellaneous classes into their own files

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/427)


src/main/scala/org/hammerlab/guacamole/dist/WindowFlatMapUtils.scala
```diff
@@ -0,0 +1,240 @@
...
+    * @param function function to flatMap: (task number, loci, iterators of regions that overlap a window around these
+    *                 loci (one region-iterator per sample)) -> T
+    * @tparam T type of value returned by function
+    * @return flatMap results, RDD[T]
+    */
+  private[dist] def windowTaskFlatMapMultipleRDDs[M <: HasReferenceRegion: ClassTag, T: ClassTag](
+    regionRDDs: PerSample[RDD[M]],
+    lociPartitions: LociMap[Long],
+    halfWindowSize: Long,
+    // TODO(ryan): factor this function type out (as a PartialFunction?)
```
arahuja: Worth removing this TODO - is it still a todo? Or make an issue for it instead?

arahuja:
👍  Like this change!

src/main/scala/org/hammerlab/guacamole/dist/WindowFlatMapUtils.scala
ryan-williams: good idea! done

timodonnell:
Yeah, this looks good to me @ryan-williams . My only thought is that ‘dist’ package name may be somewhat unclear (my first thought is it has something to do with package management), but it is tough to come up with a clear name. Maybe `distributed_utils` or `distributed_primitives`? Then could also drop the ‘Utils’ from the object names perhaps.

ryan-williams:
Good point, I'd support changing the package name to `utils` instead of `dist`. I prefer keeping the `Utils` suffixes on the `object`s though.

ryan-williams:
Actually getting in to the code, I decided to package them more granularly:
* `LociPartitionUtils` → new `loci` package.
* `WindowFlatMapUtils` → `windowing` package.
* `PileupFlatMapUtils` → `pileup` package.

I also put `LociMap` and `LociSet` into the new `loci` package.

timodonnell:
I see the logic here, but I liked the separation between Spark-independent functionality like `Pileup` and Spark-based distributed primitives like `pileupFlatmap`. @arahuja has mentioned even splitting these up into their own modules (#329), since there may be use cases for the basic functionality in non-Spark contexts. So my vote would be your original setup in this PR with a `dist` package, except call that package something a little clearer like `distributed` (I'm not a fan of calling it `utils`, since that doesn't make it clear this is the sub-package where distributed operations are implemented using Spark).

ryan-williams:
OK, moved everything into a new `distributed` package

ryan-williams:
Going to merge this in, lmk if there are further comments!

PR #428, opened 2016-04-06
remove unicode arrows

ryan-williams:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/428)


arahuja:
LGTM, why did we end up removing this now?

ryan-williams:
I agreed with @timodonnell to take them out of my recent PRs but some slipped in; at least one of these predates that but w/e.

PR #408, opened 2016-03-31
initial implementation of RNA support and filters in joint caller

timodonnell:
* Support for RNA-informed variant calling. Closes #388.
 * Basic infrastructure for filters, and a few basic filters. We will need to add more however. See #385.
 * Performance tweaks including changing some of the defs in AlleleEvidenceAcrossSamples to vals and changing some seqs to vectors.
 * More tests, including a new cancer exome + rna test dataset.
 * Some convenience arguments such as the --only-somatic argument to output only somatic variants.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/408)

Issue #385, opened 2016-02-16
joint caller: implement filters


timodonnell:
Some filters we should probably have (going by artifacts I've observed in real data)
 * Strand bias
 * Variants reads mostly start or end at the same place
 * Variants toward the end of homopolymers (perhaps take into account strand)
 * SNVs near indels
 * Reads in normal sample have lots of mismatches or poor alignments in the region

Some of these are implemented here https://github.com/hammerlab/guacamole/tree/master/src/main/scala/org/hammerlab/guacamole/filters and can be adapted for the joint caller.

For things directly involving base qualities, mapping qualities, and read depths I'd like to try integrating these directly into the likelihood computation when possible rather than having them as filters. For example, instead of having a threshold filter on read depth, it may be possible to parameterize our prior likelihood on there being a somatic variant in a more intuitive way involving minimum read depth.

Here are (some of) the filters Mutect supports: http://www.nature.com/nbt/journal/v31/n3/fig_tab/nbt.2514_T1.html

Since we're joint calling, one question is what sample should these filters run on. I think the simplest thing to do is run them on the pooled sequencing data, but another option to experiment with may be to run them on the samples that trigger a call, and then only keep the call if there are any samples that trigger the call and aren't filtered out.

Any filtered calls should be written out to a VCF with the FILTER field set to explain why it was filtered.

@e5c is down to take a stab at this once we have #384 merged and running on the cluster, assigning to her

timodonnell:
One of the most important filters to have: filter somatic calls that have evidence in the normal (i.e. cases where the variant is present in the normal at too low a fraction to result in a germline call but too high a fraction to be due to just chance). A possible extension here would be tolerating a parameterized amount of tumor contamination in normal, although can also leave that for later

jstjohn:
My PR https://github.com/hammerlab/guacamole/pull/394 has the filters you mentioned in the mutect-like somatic caller. Eventually these filters need to be tested and refactored out in such a way that the joint caller can use them easily.

timodonnell:
Mutect filters / annotations we need (much of this is based on [here](https://www.broadinstitute.org/gatk/events/slides/1506/GATKwr8-S-3-Variant_calling_with_MuTect.pdf))

- [X] Evidence in normal
- [X] Strand bias
- [X] triallelic site
- [ ] proximal gap
- [ ] poor mapping
- [ ] clustered position
[...]

PR #429, opened 2016-04-10
initial implementation of RNA support and filters in joint caller

### LINKED ISSUE #388: support RNA in joint caller
Presence of a variant in an RNA sample in addition to DNA should make us more likely to trigger a call. RNA evidence alone shouldn't result in a call.

One simple way to implement this is to have a fixed adjustment to the prior when there is evidence for the variant in any RNA sample.

timodonnell:
This PR supersedes #408. It includes what was in #408 plus some additional refactoring and tests as suggested by the in-person code review.

 * Support for RNA-informed variant calling. Closes #388.
 * Basic infrastructure for filters ("annotations"), and a few basic filters. See #385.
 * Rename the evidence classes in joint caller and move them into their own subpackage.
 * Performance tweaks including changing some of the defs in AlleleEvidenceAcrossSamples to vals and changing some seqs to vectors.
 * More tests, including a new cancer exome + rna test dataset.
 * Some convenience arguments such as the --only-somatic argument to output only somatic variants.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/429)


timodonnell:
Merging this since it's essentially just #408 rebased, which we code reviewed

Issue #106, opened 2014-06-16
Improve command line argument handling


timodonnell:
The Args4j (inherited from ADAM) library seems pretty clunky. We should figure out how to address some of these annoyances with it or switch to a better library if there is one (needs investigating):

 * The single dash (-) arguments feel ancient. Can it be made to work with GNU style arguments?
 * Args4j sorts all the arguments alphabetically, so we lose the logical grouping of arguments, and end up with e.g. obscure things at the top.
 * Args4j truncates the help message strings for the arguments, so we can't have detailed descriptions.
 * Need to be able to have a general usage statement / example at the top of the help output for a command instead of just the plain Args4j argument summary.

ryan-williams:
Some of this has been done (move to GNU-style double-dashed args), some won't be (fix lots of Args4j rough edges), and some will be addressed by #553 (supporting JSON configs), so I'm going to close this.

PR #430, opened 2016-04-10
Moving files/classes to their own packages/files or out of root guacamole package

ryan-williams:
* new `logging` package: `Common.progress`, `DelayedMessages`
* separate files for `ReadLoadingConfig`, `BamReaderAPI`, `ReadInputFilters`
- Arguments.BaseArgs → logging.DebugLogArgs
- Arguments.GenotypeOutput → variants.GenotypeOutputArgs
- Arguments.Loci → loci.LociArgs
- Arguments.ReadLoadingConfigArgs → reads.ReadLoadingConfigArgs
- Arguments.Reference → reference.ReferenceArgs
- Concordance → variants pkg
- Common.loadGenotypes → Concordance
- writeVariantsFromArguments → variants.VariantUtils.…
- use buggily-unused filters in VAFHistogram
- unroll other Common.Arguments members into Common; will move them
  out as well soon.
- GuacamoleKryoRegistrator → `kryo` package

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/430)

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/main/GeneratePartialFasta.scala
+++ b/src/test/scala/org/hammerlab/guacamole/main/GeneratePartialFasta.scala
@@ -7,7 +7,7 @@ import org.bdgenomics.utils.cli.Args4j
 import org.hammerlab.guacamole.Common.Arguments.{ReadLoadingConfigArgs, Reference}
 import org.hammerlab.guacamole.distributed.LociPartitionUtils
 import org.hammerlab.guacamole.logging.LoggingUtils.progress
-import org.hammerlab.guacamole.reads.Read.InputFilters
+import org.hammerlab.guacamole.reads.InputFilters
 import org.hammerlab.guacamole.reference.{ContigNotFound, ReferenceBroadcast}
 import org.hammerlab.guacamole.{Bases, Common, ReadSet}
 import org.kohsuke.args4j.{Argument, Option => Args4jOption}
--- a/src/test/scala/org/hammerlab/guacamole/main/GeneratePartialFasta.scala
+++ b/src/test/scala/org/hammerlab/guacamole/main/GeneratePartialFasta.scala
@@ -4,11 +4,10 @@ import java.io.{BufferedWriter, File, FileWriter}
 
 import org.apache.spark.Logging
 import org.bdgenomics.utils.cli.Args4j
-import org.hammerlab.guacamole.Common.Arguments.{ReadLoadingConfigArgs, Reference}
 import org.hammerlab.guacamole.distributed.LociPartitionUtils
 import org.hammerlab.guacamole.logging.LoggingUtils.progress
-import org.hammerlab.guacamole.reads.InputFilters
-import org.hammerlab.guacamole.reference.{ContigNotFound, ReferenceBroadcast}
+import org.hammerlab.guacamole.reads.{InputFilters, ReadLoadingConfigArgs}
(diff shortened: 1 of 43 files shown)


src/main/scala/org/hammerlab/guacamole/reads/ReadInputFilters.scala
```diff
@@ -0,0 +1,78 @@
...
+  val empty = ReadInputFilters()
+
+  /**
+   * See ReadInputFilters for full documentation.
+   *
+   * @param mapped include only mapped reads. Convenience argument that is equivalent to specifying all sites in
+   *               overlapsLoci.
+   */
+  def apply(mapped: Boolean = false,
+            overlapsLoci: LociSet.Builder = null,
```
timodonnell: why change to use null instead of option?

timodonnell:
I skimmed this kind of quickly but modulo my one comment (why un-option overlapsLoci parameter?) I think this looks reasonable

src/main/scala/org/hammerlab/guacamole/reads/ReadInputFilters.scala
ryan-williams: I wanted to support the interface `ReadInputFilters(overlapsLoci = loci)` instead of ``ReadInputFilters(overlapsLoci = Some(loci))`, so it's just a quick pass through null/Any space instead of `Option`-space

ryan-williams:
Newer commit https://github.com/hammerlab/guacamole/pull/430/commits/4ed001b774a6913871f6c1f5d553bcdfdc840b5e moves a bunch more things out of Common in case you want to have a look at that.

src/test/scala/org/hammerlab/guacamole/main/GeneratePartialFasta.scala
```diff
@@ -77,15 +80,6 @@ object GeneratePartialFasta extends Logging {
       })
     })
     writer.close()
-    Common.progress("Wrote: %s".format(args.output))
-  }
-
-  private def printUsage() = {
```
arahuja: This was unused - but perhaps makes sense to use it rather than drop it?

timodonnell:
LGTM

src/test/scala/org/hammerlab/guacamole/main/GeneratePartialFasta.scala
ryan-williams: It was unclear to me how best to use it; args4j already prints option/usage info that's automatically generated on parse failure, and I don't see a hook to customize that.

While looking into this, however, I did notice that the "stand-alone" main()s in src/test/scala/org/hammerlab/guacamole/main were not being packaged into our JAR when we run `mvn package`, making them seemingly impossible to build/run, so I moved them back under src/main/scala. I didn't see an obvious way to reconfigure Maven to allow them to be built as test JARs.

timodonnell: Just noting that I think this is one of the worst limitations of args4j. Would really like to replace it with something that allows us to give a usage message to users for each tool (#106) .

Also thanks for making sure that `GeneratePartialFasta` continues to be runnable. Need to ensure that we can always run https://github.com/hammerlab/guacamole/blob/master/scripts/generate-partial-fasta-for-tests whenever we need to regenerate the partial fastas.

ryan-williams: Sent out #451 to address running these while leaving them under `test`, so I think this is resolved.

PR #431, opened 2016-04-10
re-package distributed tests

ryan-williams:
I messed up at the end of #427 and didn't move the ex-`DistributedUtil` tests to `distributed` package.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/431)


ryan-williams:
I'm going to just merge this because it was supposed to be part of #427

PR #432, opened 2016-04-11
Factor out readset loading / seqdict checking

ryan-williams:
When loading multiple read sets, pushes sequence-dictionary consistency verification into the readset-loading machinery, and returns one along with the reads themselves.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/432)


src/main/scala/org/hammerlab/guacamole/Common.scala
```diff
@@ -156,15 +158,18 @@ object Common extends Logging {
...
-    filters: Read.InputFilters): ReadSet = {
-
-    ReadSet(
-      sc,
-      args.reads,
-      filters,
-      contigLengthsFromDictionary = !args.noSequenceDictionary,
-      config = ReadLoadingConfigArgs.fromArguments(args)
-    )
+    filters: Read.InputFilters): (RDD[MappedRead], ContigLengths) = {
```
timodonnell: Why not return a `ReadSet` here?

src/main/scala/org/hammerlab/guacamole/readsets/ReadSet.scala
```diff
@@ -0,0 +1,56 @@
...
+import org.bdgenomics.adam.models.SequenceDictionary
+import org.hammerlab.guacamole._
+import org.hammerlab.guacamole.reads.Read
+
+/**
+  * A ReadSet contains an RDD of reads along with some metadata about them.
+  *
+  * @param reads RDD of reads
+  */
+case class ReadSet(reads: ReadsRDD,
```
timodonnell: Why does the new `ReadSet` class not store the filename, filters, etc that were in the original class?

src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
```diff
@@ -0,0 +1,73 @@
...
+import org.hammerlab.guacamole._
+import org.hammerlab.guacamole.reads.Read
+
+case class ReadSets(readsRDDs: PerSample[ReadsRDD],
+                    sequenceDictionary: SequenceDictionary,
+                    contigLengths: ContigLengths) extends PerSample[ReadsRDD] {
+  override def length: Int = readsRDDs.length
+  override def apply(idx: Int): ReadsRDD = readsRDDs(idx)
+
+  val mappedReads = readsRDDs.map(_.mappedReads)
```
timodonnell: I realize Spark won't run this unless it is accessed but just checking that there's definitely no performance implications to this, e.g. memory set aside? Might want to make `lazy` if we're not sure

src/main/scala/org/hammerlab/guacamole/readsets/ReadSet.scala
ryan-williams: They weren't used anywhere… agreed that it feels reasonable to keep them around though, I'll add them back.

ryan-williams:
There's a seqdict conflict between `src/test/resources/cancer-wes-and-rna-celsr1/{normal_0,tumor_wes_2,tumor_rna_11}.bam`; the latter omits extra "unmapped" contigs found in the former two.

I'm going to change the seqdict-parsing in `ReadSets` to merge all the input files' sequence dictionaries, throwing an exception if any have conflicting info for a given contig name. I'll ping here when that's committed.

timodonnell:
Seems like a reasonable idea, but I think going from just a `ReadSet` class to a `ReadSet`, `ReadSets`, and `ReadRDD` classes is maybe introducing a bit more complexity than it saves. If the goal is to avoid having the callers have to assert that the sequence dictionaries match, maybe just add a function somewhere to do that check, and have the commands call that function? Or is there another problem this is addressing that I'm misunderstanding?

src/main/scala/org/hammerlab/guacamole/Common.scala
ryan-williams: Good point; all callers that load just one set of reads currently filter to mapped reads as well, so I was trying to expose a convenient interface to them.

I've changed this so that there are now special-case helpers in `ReadSets` for single-read-set cases ("mapped" and "all"), and removed `ReadSet` altogether.

src/main/scala/org/hammerlab/guacamole/readsets/ReadSet.scala
ryan-williams: I've added a `sourceFile` attribute to `ReadsRDD` and dropped `ReadSet` outright; it was just a convenience wrapper for `ReadSets` when there was exactly one input, but it makes more sense to just expose wrappers on `ReadSets` that unroll the relevant bits for callers when they are just loading reads from one file, so I've done that.

ryan-williams: I didn't add back the "filters" attribute because it's less clear whether that should be an attribute of individual `ReadsRDD`s, or the collective `ReadSets`; today all reads that are loaded share filters, but it doesn't seem worth exposing that as an attribute of a `ReadSets`, especially because the filters are unchanged from those passed in by the user to a `ReadSets`; everything else on a `ReadSets` is loaded from the underlying data.

src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
ryan-williams: We definitely don't need to worry about this causing any Spark jobs to run that wouldn't otherwise, and the RDD objects created here are tinydata, whether on the driver or executors. 

When I'm reading code and see a `lazy` qualifier, it signals to me that it is actually important that that `val` is only computed when it is used, presumably because it's expensive to compute and/or is frequently not used on instances of the class; I don't think either of those are the case here, so I dropped it.

I'm meaning to put a unit test somewhere that explores in depth how `lazy` behaves with Java and Kryo serialization; if it's important to you I can do that test before we merge this, though even if `lazy` doesn't come with any serde gotchas, I think I prefer this way, for the reasons discussed above.

ryan-williams:
> going from just a ReadSet class to a ReadSet, ReadSets, and ReadRDD classes is maybe introducing a bit more complexity than it saves.

This is a good point; on thinking about it more, `ReadSet` was extraneous and so I've removed it. `ReadSets` and `ReadsRDD` remain; they handle information that pertains to a collection of inputs vs. a particular one, resp.

Other changes:
* SequenceDictionary is now merged from all inputs, and stored in the `ReadSets`.
* Source filename of an `RDD[Read]` is now a member of `ReadsRDD`.
* Read-set loading has also been moved from `Common` into `ReadSets`.
  * Helpers exist for loading multiple readsets, single sets of reads (mapped or all), and tumor/normal sets.

src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
```diff
@@ -24,49 +24,76 @@ object ReadSets {
...
+          .reduceByKey(math.max)
+          .collectAsMap()
+          .toMap
       }
     }
 
     ReadSets(readRDDs.map(ReadsRDD(_)).toVector, sequenceDictionary, contigLengths)
   }
 
+  def mergeSequenceDictionaries(filenames: Seq[String], dicts: Seq[SequenceDictionary]): SequenceDictionary = {
```
timodonnell: I think a header comment here explaining what sequence dictionaries are and why we're merging them would be useful

src/main/scala/org/hammerlab/guacamole/readsets/ReadsRDD.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  lazy val mappedPairedReads: RDD[PairedRead[MappedRead]] =
+    reads.flatMap {
+      case rp: PairedRead[_] if rp.isMapped => Some(rp.asInstanceOf[PairedRead[MappedRead]])
+      case _                                => None
+    }
+}
+
+object ReadsRDD {
+  def apply(pair: (RDD[Read], String)): ReadsRDD = ReadsRDD(pair._1, pair._2)
```
timodonnell: this convenience constructor seems unnecessary, I think it'd be clearer to use the default constructor wherever this is getting used

ryan-williams: This constructor describes how to make a `ReadsRDD` from a `(RDD[Read], String)` tuple. I believe the class should own the logic for constructing instances of itself from variously-shaped inputs, rather than lysing such boilerplate out into call-sites 😄 .

FWIW, this is currently used in [one place](https://github.com/hammerlab/guacamole/pull/432/files#diff-91485b818b79f0449c217f8f2ad7580aR61), and I prefer letting it say

```
.map(ReadsRDD(_))
```

instead of

```
.map(t => ReadsRDD(t._1, t._2))
```

src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
ryan-williams: Ah, I missed this, done!

ryan-williams:
Responded to Tim's last [comment](https://github.com/hammerlab/guacamole/pull/432#discussion_r60482806), that I'd missed previously.

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/SomaticJointCaller.scala
```diff
@@ -66,15 +65,13 @@ object SomaticJoint {
   def inputsToReadSets(sc: SparkContext,
```
timodonnell: hadn't noticed this before -- I guess `InputCollection` has escaped the joint caller and is used by top level guacamole code? I'm not necessarily against that if it is useful elsewhere, but I think in general we should probably wait until the joint caller matures more before pulling its stuff out into general infrastructure

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -93,7 +81,7 @@ object Read extends Logging {
    * @param record
    * @return
    */
-  def fromSAMRecord(record: SAMRecord): Read = {
```
timodonnell: I actually kind of like the explicitness of `fromSAMRecord` and `fromADAMRecord` instead of just an overloaded `apply`, but if you feel strongly this is better i'm ok with it

src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
```diff
@@ -0,0 +1,350 @@
...
+  override def length: Int = readsRDDs.length
+  override def apply(idx: Int): ReadsRDD = readsRDDs(idx)
+
+  val mappedReads = readsRDDs.map(_.mappedReads)
+}
+
+object ReadSets {
+
+  /**
+   * Load just the mapped reads from an input ReadSet.
```
timodonnell: remove 'mapped' from comment?

src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
```diff
@@ -0,0 +1,350 @@
...
+      else
+        loadFromADAM(filename, sc, filters, config)
+
+    val reads = filterRDD(allReads, filters, sequenceDictionary)
+
+    (reads, sequenceDictionary)
+  }
+
+  /** Returns an RDD of Reads and SequenceDictionary from reads in BAM format **/
+  private def loadFromBAM(filename: String,
```
timodonnell: may want to keep these public as they may come in handy to e.g. apply different filters to different BAMs (or maybe the suggested way to do that is just load individual `ReadSets` for each bam?)

timodonnell:
posted some minor comments to consider but either way LGTM

src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
ryan-williams: good catch, done

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/SomaticJointCaller.scala
ryan-williams: I'm confused by this comment; this is `SomaticJointCaller.scala`, `InputCollection` is only used under the `jointcaller` package atm, and this PR doesn't change that.

My change here is just updating the joint-caller's usage of shared "read set" APIs.

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
ryan-williams: Interesting point. In general I don't like giving constructors names that are redundant with the types that they expect; the arguments will commonly have descriptive names already, like the sole caller of the ADAM version here where we now have:

```
adamRecords.map(Read(_))
```

instead of

```
adamRecords.map(Read.fromADAMRecord(_))
```

which is preferable to me, so I think I'll leave these changed.

src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
ryan-williams: Interesting. I think the right way to support what you're describing is to have an API that constructs a `ReadSets` from a list of filenames and corresponding list of filters. I think that would be worth adding.

Mimicking that functionality using a bunch of separate `load` calls should not be allowed because that directly undoes the small amount of worthwhile abstraction I've added here, namely the seqdict handling and APIs that treat a logical collection of read-sets collectively.

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
arahuja: I think I also prefer the overloaded constructors

src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
ryan-williams: OK, I've added a constructor with the API you've proposed! Good idea.

PR #433, opened 2016-04-11
Upgrade Hadoop BAM to use BAM index

### LINKED ISSUE #399: Upgrade HadoopBAM to use BAM index
Right now we use samtools for locally BAM reading since it can use the BAM index and directly access the reads of interest.

Hadoop BAM should support this now as well (https://github.com/HadoopGenomics/Hadoop-BAM/pull/59) and will allow us to only access a subset of the BAM as oppposed to filtering after loading when using the the `--loci` argument.

arahuja:
This fixes #399 

However, while we have been nominally targeting Java 8 before, the Hadoop BAM 7.4 requires it. I think this is a good move overall, but it does require a few steps to get your environment working again. I've tested this locally with Java 8 and will do so on our cluster - we can hold off on merging until then. I wanted to open the PR in advance to discuss the change/upgrade to Java 8 though.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/433)


src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -294,109 +261,56 @@ object Read extends Logging {
...
-      })
-      (sc.parallelize(reads.toSeq), sequenceDictionary)
+
+    val conf = sc.hadoopConfiguration
+    val samHeader = SAMHeaderReader.readSAMHeaderFrom(path, conf)
+    val sequenceDictionary = SequenceDictionary.fromSAMHeader(samHeader)
+
+    if (filters.overlapsLoci.isEmpty || filename.endsWith(".bam")) {
+      val lociString = filters.overlapsLoci.map(_.result(contigLengths(sequenceDictionary)).toString)
+      lociString.foreach(conf.set(BAMInputFormat.INTERVALS_PROPERTY,_))
```
ryan-williams: so our loci-string syntax is interoperable with hadoop-bam's? any 0- versus 1-based indexing gotchas?

ryan-williams:
This looks great. Moving to Java 8 and all the removed complexity here sounds awesome.

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
arahuja: Ah! Yea, I likely need to shift these... will fix and add tests for this

arahuja:
@ryan-williams Thanks for the quick review - updated with the proper coordinates

src/main/scala/org/hammerlab/guacamole/loci/LociSet.scala
```diff
@@ -92,7 +93,34 @@ case class LociSet(map: LociMap[Long]) {
...
+      .flatMap(
+        contig => {
+          map
+            .onContig(contig)
+            .asMap
+            .iterator
+            // We had 1 to the start to move to 1-based coordinates
+            // Since the `Interval` end is inclusive, we are adding and subtracting 1, no-op
+            .map(pieces => new Interval(contig, pieces._1.start.toInt + 1, pieces._1.end.toInt))
+        }).toList
```
timodonnell: minor point: any reason to make this a list? it will use more memory than a Vector

src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
```diff
@@ -36,39 +34,60 @@ class ReadSetSuite extends GuacFunSuite {
...
+      }
+    }
+
+    val lociAndExpectedCounts = Seq(
+      ("20:9999900", 0),
+      ("20:9999901", 1),
+      ("20:9999912", 2),
+      ("20:0-10000000", 9),
+      ("20:10270532", 1),
+      ("20:10270533", 0)
```
timodonnell: maybe worth testing on a LociSet that is more than just one interval?

timodonnell:
LGTM! Is performance about the same on loading a small subset of a local bam as with the previous implementation?

Thanks for doing this @arahuja !

src/test/scala/org/hammerlab/guacamole/reads/ReadSetSuite.scala
arahuja: Good idea! - will add

src/main/scala/org/hammerlab/guacamole/loci/LociSet.scala
arahuja: Not particularly, it's converted to a Java list anyways to pass to `BAMInputFormat`

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -294,109 +263,56 @@ object Read extends Logging {
...
-          Some(read)
-        }
-      })
-      (sc.parallelize(reads.toSeq), sequenceDictionary)
+
+    val conf = sc.hadoopConfiguration
+    val samHeader = SAMHeaderReader.readSAMHeaderFrom(path, conf)
+    val sequenceDictionary = SequenceDictionary.fromSAMHeader(samHeader)
+
+    if (filters.overlapsLoci.isEmpty || filename.endsWith(".bam")) {
```
ryan-williams: `filters.overlapsLoci.isEmpty` here is confusing to me given this branch only maps over `filters.overlapsLoci`

arahuja:
Will re-open this after it has been updated

PR #434, opened 2016-04-11
Add read-based path scoring

arahuja:
This updates the scoring mechanism of the germline assembly caller to score the paths based on how the reads align to them. The paths that explain the reads the best are chosen. 

Since this adds a large computational expense, there is now a `shortcut-assembly` flag to skip assembly in 'simple areas' (where there are no indels or multiple mismatches nearby)

I think the last test case displays some of this functionality the best - where we can successfully call the 3 variants without any of the false positives:
image

For a sample dataset (NA12878 chr20), this improve recall significantly, going from:
```
Recall / Sensitivity: 0.913427632367 Precision: 0.880323685189
```
to
```
Recall / Sensitivity: 0.951337972881 Precision: 0.838393433011
```

and specifically on indels, from
```
Recall / Sensitivity: 0.818181818182 Precision: 0.586783204394
```
to
```
Recall / Sensitivity: 0.856719367589 Precision: 0.61954385756
```

HaplotypeCaller on the same data, for all variants
```
Recall / Sensitivity: 0.983819243922 Precision: 0.832385755942
```

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/434)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
@@ -11,8 +11,9 @@ import org.hammerlab.guacamole.alignment.AffineGapPenaltyAlignment
 import org.hammerlab.guacamole.assembly.DeBruijnGraph
 import org.hammerlab.guacamole.distributed.LociPartitionUtils.partitionLociAccordingToArgs
 import org.hammerlab.guacamole.distributed.WindowFlatMapUtils.windowFlatMapWithState
-import org.hammerlab.guacamole.distributed.{LociPartitionUtils, WindowFlatMapUtils}
+import org.hammerlab.guacamole.likelihood.Likelihood
 import org.hammerlab.guacamole.loci.LociMap
+import org.hammerlab.guacamole.pileup.Pileup
 import org.hammerlab.guacamole.reads.{MappedRead, Read}
 import org.hammerlab.guacamole.reference.{ReferenceBroadcast, ReferenceGenome}
 import org.hammerlab.guacamole.variants.{Allele, AlleleConversions, AlleleEvidence, CalledAllele}
@@ -47,7 +48,7 @@ object GermlineAssemblyCaller {
     @Args4jOption(name = "--min-alignment-quality", usage = "Minimum alignment qualities of the read")
     var minAlignmentQuality: Int = 30
 
-    @Args4jOption(name = "--reference-fasta", required = false, usage = "Local path to a reference FASTA file")
+    @Args4jOption(name = "--reference-fasta", required = true, usage = "Local path to a reference FASTA file")
     var referenceFastaPath: String = null
 
     @Args4jOption(name = "--min-area-vaf", required = false, usage = "Minimum variant allele frequency to investigate area")
@@ -55,6 +56,12 @@ object GermlineAssemblyCaller {
 
     @Args4jOption(name = "--min-occurrence", required = false, usage = "Minimum occurrences to include a kmer ")
     var minOccurrence: Int = 3
+
+    @Args4jOption(name = "--min-likelihood", usage = "Minimum Phred-scaled likelihood. Default: 0 (off)")
+    var minLikelihood: Int = 0
+
+    @Args4jOption(name = "--shortcut-assembly", required = false, usage = "Skip assembly process in inactive regions")
+    var shortcutAssembly: Boolean = false
 
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -118,9 +125,7 @@ object GermlineAssemblyCaller {
      * @param minOccurrence Minimum times a kmer must appear to be in the DeBruijn graph
      * @param expectedPloidy Expected ploidy, or expected number of valid paths through the graph
      * @param maxPathsToScore Number of paths to align to the reference to score them
-     * @return A set of variants in the window
-     *         TODO: This currently passes along a graph as state, but rebuilds it each time
-     *         This can be updated to use the existing graph and just update it
+     * @return
```
ryan-williams: drop this line if the comment's gone?

timodonnell:
Maybe this is a good one to do an in-person review? I'd be interested in getting up to speed on this work and I think that would help

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -155,94 +159,30 @@ object GermlineAssemblyCaller {
...
+        paths
+      } else if (paths.size <= maxPathsToScore) {
+        val pathScores = DenseVector.zeros[Int](paths.size)
+        reads.foreach(
+          read => {
+            pathScores :+= DenseVector(paths.map(path => -AffineGapPenaltyAlignment.align(read.sequence, path).alignmentScore): _*)
           })
 
-        })
+        argtopk(pathScores, expectedPloidy).map(paths(_))
```
ryan-williams: could this block (up to [`pathScores` above](https://github.com/hammerlab/guacamole/pull/434/files#diff-1a28a5c8b1985b94c8856c89798c3aa5R173)) be replaced with the following?

```
        (for {
          path <- paths
        } yield {
          reads.map(
            read =>
              -AffineGapPenaltyAlignment.align(read.sequence, path).alignmentScore
          ).sum -> path
        }).sortBy(-_._1).take(expectedPloidy)
```

if so, why the `argtopk` and `DenseVector` bit? more mem-efficient than scala `Vector`?

ryan-williams: (implicit in this question is whether it makes a difference whether you iterate over all paths then reads, vs. reads then paths, OOC)

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -289,49 +231,81 @@ object GermlineAssemblyCaller {
...
 
             // Compute the number reads with variant bases from the reads overlapping the currentLocus
-            val currentLocusVAF = variableReads.toFloat / currentLocusReads.length
-            if (currentLocusVAF > minAreaVaf) {
-              val result = discoverHaplotypes(
+            val pileupAltReads = (pileup.depth - pileup.referenceDepth)
+            if (currentLocusReads.isEmpty || pileupAltReads < minAltReads) {
+              (lastCalledLocus, Iterator.empty)
+            } else if (shortcutAssembly &&
+              currentLocusReads.count(r => r.countOfMismatches > 1 || r.cigar.numCigarElements() > 1) / currentLocusReads.size.toFloat < minAreaVaf) {
```
arahuja: Yes, good catch, I had these named differently before and somehow mixed them up.

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -289,49 +231,81 @@ object GermlineAssemblyCaller {
...
+                window.referenceName,
+                referenceStart,
+                referenceEnd
+              )
+
             // Find the reads the overlap the center locus/ current locus
             val currentLocusReads =
               window
                 .currentRegions()
                 .filter(_.overlapsLocus(window.currentLocus))
```
arahuja: `window.currentRegions()` includes reads that overlap the window, but only  subset overlap the current locus.

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
arahuja: I think in this implementation it won't matter, so I can simplify it as you stated. This originally developed from building the full matrix of path/read scores ... but I will think about this a bit

src/main/scala/org/hammerlab/guacamole/reads/MappedRead.scala
```diff
@@ -54,21 +54,27 @@ case class MappedRead(
...
+
   /**
    * Number of mismatching bases in this read. Does *not* include indels: only looks at read bases that align to a
    * single base in the reference and do not match it.
    *
    * @param referenceContigSequence the reference sequence for this read's contig
    * @return count of mismatching bases
    */
+
+  var countOfMismatches = -1
```
ryan-williams: given the mixup above, I'd like to change this pattern somehow to ensure that this isn't referenced before it's initialized

arahuja:
@timodonnell Do you still want to do this in person? This change itself isn't large, but happy to give a full overview in person.

timodonnell:
No need to hold up this PR for it, but sounds good to do an overview in
person

JPFinnigan:
This might be a trivial point to bring up, but I believe the approach you're implementing here in weighting paths by coverage support, is the same general idea used to create StringTie's "flow network." Perhaps you guys have already taken a look, but just in case it might be useful for you, the formalism behind their approach is contained in the methods section of their Nature Biotechnology paper (Link: http://www.nature.com/nbt/journal/v33/n3/pdf/nbt.3122.pdf)

arahuja:
Thanks for the link @JPFinnigan, will take a look see if there anything we can use from that. Also will be nice to see if this is eventually useful with RNA as well

arahuja:
I updated the scoring based on @ryan-williams suggested simplification, goon to merge this is in now

PR #435, opened 2016-04-12
some mostly-style tweaks to #434

ryan-williams:
feel free to take or leave as much of this as you want; I was semi-consciously tweaking things in IJ while reviewing and got carried away. a couple of changes here:

- rearrange methods in GermlineAssembly.Caller in rough order of which ones call which others; unfortunately this DoS's the diff. I can separate that out into its own commit if you like.
- make some methods private
- move DBJ types into companion object
- misc comment / spacing / loop-structure tweaks

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/435)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
+++ b/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
@@ -1,15 +1,12 @@
 package org.hammerlab.guacamole.assembly
 
 import org.hammerlab.guacamole.Bases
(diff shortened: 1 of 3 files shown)


ryan-williams:
The reordering of the methods caused too much noise here, going to close this and send some of the other changes in a different PR.

arahuja:
Cool, thanks, that makes sense! I think some things here re: private methods and re-arranging can be held-off. A lot of this code will be re-used in the somatic caller and then might move out anyways. But no harm in making in private in the meantime I guess

PR #436, opened 2016-04-12
style nits

ryan-williams:
PR on top of #434 
- make some methods private
- comment spacing nits
- rearrange path/alignment-scores code in `buildVariantsFromPaths` for simplicity
- experiment: expand [complicated `if` condition in discoverGermlineVariants](https://github.com/hammerlab/guacamole/compare/rb-scoring-paths...ryan-williams:p434?expand=1#diff-1a28a5c8b1985b94c8856c89798c3aa5R281)
- factor two uses of `reference.getContig(referenceName)` into a `referenceContig` val

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/436)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
@@ -80,7 +80,7 @@ object GermlineAssemblyCaller {
      * @param endLocus End (exclusive) locus on the reference
      * @return Subsequence overlapping [startLocus, endLocus)
      */
-    def getSequenceFromRead(read: MappedRead, startLocus: Int, endLocus: Int) = {
+    private def getSequenceFromRead(read: MappedRead, startLocus: Int, endLocus: Int) = {
       read.sequence.slice(startLocus - read.unclippedStart.toInt, endLocus - read.unclippedStart.toInt)
     }
 
@@ -93,10 +93,10 @@ object GermlineAssemblyCaller {
      * @param minOccurrence Mininum number of times a subsequence needs to appear to be included
      * @return List of subsequences overlapping [startLocus, endLocus) that appear at least `minOccurrence` time
      */
-    def getConsensusKmer(reads: Seq[MappedRead],
-                         startLocus: Int,
-                         endLocus: Int,
-                         minOccurrence: Int): Iterable[Array[Byte]] = {
+    private def getConsensusKmer(reads: Seq[MappedRead],
+                                 startLocus: Int,
+                                 endLocus: Int,
+                                 minOccurrence: Int): Iterable[Array[Byte]] = {
 
       // Filter to reads that entirely cover the region
       // Exclude reads that have any non-M Cigars (these don't have a 1 to 1 base mapping to the region)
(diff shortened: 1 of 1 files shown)


src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -458,17 +466,18 @@ object GermlineAssemblyCaller {
...
     variants.toSet
   }
 
   /**
-    * Call variants using a pileup and genotype likelihoods
-    * @param pileup Pileup at a given locus
-    * @return Possible set of called variants
-    */
-  def callPileupVariant(pileup: Pileup): Set[CalledAllele] = {
+   * Call variants using a pileup and genotype likelihoods
```
arahuja: Hm - it seems like intellij makes them one way and we want them the other? No preference either way, whatever is easiest really. I think scalariform would move these around too

ryan-williams: yea, per [my comment](https://github.com/hammerlab/guacamole/pull/437#issuecomment-208975109) on #437, i think the defaults for scalariform and IJ conflicted, but IJ can go either way (maybe scalariform can too?).

I realigned them here partly because the method above had them this way.

I also added the blank line, which I thought was standard.

ryan-williams:
re: `private` methods; I'm fine to take that out if you prefer. 

One thing that interfered with my reading through of #434 was that there are 8 top-level public methods in `GermlineAssembly{,.Caller}`, almost all of them are used exactly once in the whole codebase (by one of the others in the same file), and they were not ordered in the file according to that structure.

There are a couple ways to mitigate that but keeping clear public, package-private, and private qualifiers is one good one.

I'd vote for physically placing them around each other as well.

I also don't necessarily understand why `buildVariantsFromPaths` and `callPileupVariant` are in the root `GermlineAssembly` object while the others are in `Caller`.

I'm happy to make those further changes here if you like; having them as their own commit should make things clearer.

arahuja:
> I also don't necessarily understand why buildVariantsFromPaths and callPileupVariant are in the root GermlineAssembly object while the others are in Caller.

No particularly, just an oversight on my part

> One thing that interfered with my reading through of #434 was that there are 8 top-level public methods in GermlineAssembly{,.Caller}, almost all of them are used exactly once in the whole codebase (by one of the others in the same file), and they are not ordered in the file according to the logical call-hierarchy.
There are a couple ways to mitigate that but keeping clear public, package-private, and private qualifiers is one good one.

Yea, agree that these could use better organization. There are two reasons for it now. 1) Some of the big functionality i.e. `discoverHaplotypes` and `buildVariants` happen inside the `discoverGermlineVariants` which has control over the sliding window and how to adjust it based on the results of those. Obviously, not great to let an implementation details affect the interface. 2) Some of them are factored out as they are used in the `somatic-assembly`. I'll get a PR on that and move some of the shared functionality to another package.

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -299,8 +308,8 @@ object GermlineAssemblyCaller {
                     currentReference,
                     window.currentRegions()
                   )
-                    .filter(variant => lastCalledLocus.forall(_ < variant.start)) // Filter variants before last called
-                    .toSet // Remove any duplicate variants due to paths that have overlapping segments
```
arahuja: Oh yea, I moved this `.toSet` up into the function

ryan-williams:
Cool, thanks Arun!

PR #437, opened 2016-04-12
style nits

ryan-williams:
PR on top of #433 
- remove unnecessary →iterator→list round trip per @timodonnell's comment
- comment typo

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/437)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/loci/LociSet.scala
+++ b/src/main/scala/org/hammerlab/guacamole/loci/LociSet.scala
@@ -97,20 +97,20 @@ case class LociSet(map: LociMap[Long]) {
   /**
     * Build a collection of HTSJDK Intervals which are closed [start, end], 1-based intervals
     */
-  def toHtsJDKIntervals: List[Interval] = {
+  def toHtsJDKIntervals: Seq[Interval] = {
     map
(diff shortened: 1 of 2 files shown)


arahuja:
LGTM - it seems we have an unofficial style guide going now - can we codify this somewhere? This is one area that scalariform did help. If there are fixed style rules having something enforce them rather than manual review would be nice

ryan-williams:
Agreed that we should probably codify style rules again/more somewhere post-scalariform.

The two style/linty things in this PR are in [these couple lines](https://github.com/hammerlab/guacamole/pull/437/files#diff-d2e9ed13f6b3c8bc7378acba75118592L91):
- blank `*` line between scaladoc initial message and `@param` doc-lines
- `*`s in scaladocs aligned with each other.
  - whether they go under the first or second `*` of the opening `/**` is something we've not been consistent about; scalariform used to align under the second, but IJ can be configured to do either:

  

  however, whichever we choose, they should be aligned within any given scaladoc block.

I'll file an issue to track what we prefer here and what existing tools can do.

The more important change in here is the seq/list/iterator bit, so maybe calling this "style nits" was misleading.

Issue #438, opened 2016-04-12
Upgrade to Scala 2.11


ryan-williams:
Eventually we should do this; collecting notes about how/when here.

- should we have a transition period where we support both?
- or should we switch over in one go (once we're sure things work on 2.11)?

Based on only high-level knowledge of the ramifications, I'm in favor of the latter.

Some context / notes:
- [2.10 has been EoL'd for ~18mos](http://www.scala-lang.org/news/2.12-roadmap/). 
- Of our dependencies, ADAM and Spark support Scala 2.11; if any don't that's worth finding out because it also probably means they're unmaintained.
- Spark currently [debating dropping Java 7 / Scala 2.10 in upcoming Spark 2.0 release](http://mail-archives.apache.org/mod_mbox/incubator-spark-dev/201604.mbox/%3CCAAsvFPmiJxva1UxFpHm9aSGYX4c0K8qByO1gAfv3sKPd8vJWig@mail.gmail.com%3E).
- @arahuja or @timodonnell mentioned that case-class boxing might be better in 2.11, as well as whatever else was added to scala in the last ~18mos :)

arahuja:
Also, a fan of the latter. Tried this in #433 but ran into some dependency issues (run time class not found errors),but  I think I had just needed Spark compiled for the right scala version though and didn't test further

ryan-williams:
Cool, gtk! Also the impending move to Java 8 seems like a nudge toward Scala 2.11, as I've heard that Java 8 interactions with Scala 2.10 are possibly wonky.

arahuja:
@ryan-williams Have you given this a shot? I tried just changing the config and an error I got was 
```
Error:(132, 47) diverging implicit expansion for type org.hammerlab.guacamole.reference.Position => Comparable[org.hammerlab.guacamole.reference.Position]
starting with method orderingToOrdered in object Ordered
      ReferenceRegion(contigName, start, end) <- rdd                                        ^
```

Seems to have to do with the `val ordering` defined in the `Position` companion object, but not sure how to proceed.

I'm guessing we will also need to handle https://github.com/hammerlab/magic-rdds/issues/23 first?

ryan-williams:
I tried this once a little while ago; one issue I remember is that it was unclear whether Zinc / incremental compilation works in Scala 2.11? This may be happening on https://github.com/bigdatagenomics/adam/pull/1142 atm too.

[...]

PR #440, opened 2016-04-12
bump Spark, Hadoop, Scala versions

ryan-williams:
- I use Spark 1.6.1 exclusively; an important crashy failure mode went away between 1.6.0 and 1.6.1 with Spark's move off of Akka, if my memory of my experiences serves me right.
- Bumping from Scala 2.10.3 to 2.10.5 shouldn't break anything; seems like a good small step toward #438.
- We use Hadoop 2.6.0+ (cdh5.5.1), so might as well codify that here.

If anyone feels like this is a dangerous change I'm open to trying to run some tests, but other things equal I'd bump these and then post-hoc troubleshoot if anything goes wrong.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/440)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,10 +19,10 @@
     <adam.version>0.19.0</adam.version>
     <bdg-formats.version>0.7.0</bdg-formats.version>
     <java.version>1.8</java.version>
-    <scala.version>2.10.3</scala.version>
+    <scala.version>2.10.5</scala.version>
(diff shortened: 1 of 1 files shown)


arahuja:
LGTM 👍

arahuja:
Good to merge?

ryan-williams:
Sure! Haven't tested on cluster but seems low risk

PR #441, opened 2016-04-13
Require kryo registration

ryan-williams:
Register a ton of currently unregistered classes.
- unregistered classes are currently serialized with their fully-qualified class name at the front of every record
- I copied a bunch of this from [work I'd done toward this end in Pageant](https://github.com/hammerlab/pageant/blob/045fae6c7f44691f7cac4f21882a46bdce9cd285/src/main/scala/org/hammerlab/pageant/kryo/PageantKryoRegistrar.scala).

As a follow-on, I may try a programmatic whittling down of the class-registrations added here: run `mvn test`, comment out one of them, see if it still passes, if not, uncomment it, run it again, etc.

There *shouldn't* be any missed classes in the below unless a large code path is not covered by any tests.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/441)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/Common.scala
+++ b/src/main/scala/org/hammerlab/guacamole/Common.scala
@@ -238,7 +238,7 @@ object Common extends Logging {
     }
 
     if (config.getOption("spark.kryo.registrator").isEmpty) {
-      config.set("spark.kryo.registrator", "org.hammerlab.guacamole.kryo.GuacamoleKryoRegistrator")
+      config.set("spark.kryo.registrator", "org.hammerlab.guacamole.kryo.GuacamoleKryoRegistrar")
     }
 
     if (config.getOption("spark.kryoserializer.buffer").isEmpty) {
--- a/src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrar.scala
+++ b/src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrar.scala
@@ -38,7 +38,7 @@ import org.hammerlab.guacamole.reads.{MappedRead, MappedReadSerializer, MateAlig
 import org.hammerlab.guacamole.reference.ReferenceBroadcast.MapBackedReferenceSequence
 import org.hammerlab.guacamole.variants._
 
-class GuacamoleKryoRegistrator extends KryoRegistrator {
+class GuacamoleKryoRegistrar extends KryoRegistrator {
   override def registerClasses(kryo: Kryo) {
 
     // Register ADAM serializers.
--- a/src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrator.scala
+++ b/src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrator.scala
@@ -19,35 +19,131 @@
 package org.hammerlab.guacamole.kryo
 
 import com.esotericsoftware.kryo.Kryo
+import org.apache.spark.serializer.KryoRegistrator
+import org.apache.spark.storage.BroadcastBlockId
+import org.bdgenomics.adam.models.{ReferenceRegion, SequenceDictionary, SequenceRecord}
 import org.bdgenomics.adam.serialization.ADAMKryoRegistrator
+import org.bdgenomics.formats.avro.{AlignmentRecord, Genotype => BDGGenotype}
+import org.hammerlab.guacamole.commands.VariantLocus
+import org.hammerlab.guacamole.commands.jointcaller.Parameters.SomaticGenotypePolicy
+import org.hammerlab.guacamole.commands.jointcaller.annotation.{InsufficientNormal, MultiSampleAnnotations, SingleSampleAnnotations, StrandBias}
+import org.hammerlab.guacamole.commands.jointcaller.evidence.{MultiSampleMultiAlleleEvidence, MultiSampleSingleAlleleEvidence, NormalDNASingleSampleSingleAlleleEvidence, TumorDNASingleSampleSingleAlleleEvidence, TumorRNASingleSampleSingleAlleleEvidence}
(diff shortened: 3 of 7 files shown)


timodonnell:
Cool, interested to see how this affects performance. For those changes in the joint caller, I actually originally had many of them as `defs` to minimize memory usage and data transfer but found that I got better performance by changing them to `vals` since the driver was spending a lot of time recomputing them when they were getting written out (they get computed on the cluster nodes then sent to the driver). I think which side of the CPU / memory tradeoff we end up taking there is going to depend on more tuning (they might well end up as `transient lazy val` later) but I'd like to retain the ability to easily switch them around. It'd also be nice to be able to directly compare the performance effect of registering the kryo classes for the joint caller without confounding it with those changes. Is it possible to do this PR in a way that leaves those as `vals`?

src/main/scala/org/hammerlab/guacamole/Serialization.scala
```diff
@@ -19,27 +19,78 @@
...
 
     // Register ADAM serializers.
-    super.registerClasses(kryo)
+    new ADAMKryoRegistrator().registerClasses(kryo)
+
+    kryo.register(classOf[Array[BDGGenotype]])
+    kryo.register(classOf[SequenceDictionary])
+    kryo.register(classOf[SequenceRecord])
+
+    kryo.register(classOf[PileupElement])
```
arahuja: I don't think we are ever serializing this - is there a reason we want this here?

src/main/scala/org/hammerlab/guacamole/Serialization.scala
```diff
@@ -48,5 +99,71 @@ class GuacamoleKryoRegistrator extends ADAMKryoRegistrator {
...
+    kryo.register(classOf[Array[AlignmentRecord]])
+    kryo.register(classOf[ReferenceRegion])
+    kryo.register(classOf[Array[ReferenceRegion]])
+
+    kryo.register(classOf[org.bdgenomics.formats.avro.Strand])
+    kryo.register(classOf[org.bdgenomics.adam.models.MultiContigNonoverlappingRegions])
+    kryo.register(classOf[org.bdgenomics.adam.models.NonoverlappingRegions])
+
+    // Added to Spark in 1.6.0; needed here for Spark < 1.6.0.
+    kryo.register(classOf[Array[Tuple1[Any]]])
```
arahuja: Can we just bump the Spark version to avoid this?

ryan-williams: Possibly; I guess there's a question about whether having our POM depend on Spark 1.6.1 implies we don't consider earlier versions of Spark to be supported.

I'm fine saying that and removing these.

src/main/scala/org/hammerlab/guacamole/Serialization.scala
ryan-williams: This one I specifically ran in to while making this PR; something in a test is ser'ing it.

I'll run my automated audit of all of these and then we should go through each and make sure (and probably document here) we understand the code paths and circumstances under which each one is ser'd, I think there are probably some surprises in store.

arahuja: >  automated audit

Say more?

ryan-williams:
@timodonnell:
> Is it possible to do this PR in a way that leaves those as vals?

I've taken my changes out of JointCaller, and am registering [3 anonymous classes](https://github.com/hammerlab/guacamole/pull/441/commits/613f937f0d2c921152a2db2ab229fa31da17ede8#diff-f4837c489d35507f52a07f828b317394R58) instead that get all the JointCaller objects registered.

src/main/scala/org/hammerlab/guacamole/Serialization.scala
ryan-williams: My plan is to farm out N processes that each build and run all guacamole tests, each one with a different `kryo.register` line commented out, to obtain the minimal necessary set of registrations.

I haven't decided whether I think that should block merging this PR, but I'll probably attempt it early this week.

src/main/scala/org/hammerlab/guacamole/Serialization.scala
ryan-williams: I removed these.

src/main/scala/org/hammerlab/guacamole/Serialization.scala
arahuja: Wouldn't setting require registration to true and then running our basic callers do the same? Anything unregistered would result in an error?

arahuja:
LGTM! Doesn't seem to be any harm in over-registering, but could be difficult to maintain

src/main/scala/org/hammerlab/guacamole/Serialization.scala
```diff
@@ -19,27 +19,83 @@
...
+    kryo.register(classOf[BroadcastBlockId])
+    kryo.register(classOf[Parameters.SomaticGenotypePolicy.Value])
+    kryo.register(classOf[StrandBias])
+    kryo.register(classOf[SingleSampleAnnotations])
+    kryo.register(classOf[AlleleAtLocus])
+    kryo.register(classOf[MultiSampleSingleAlleleEvidence])
+    kryo.register(classOf[MultiSampleMultiAlleleEvidence])
+    kryo.register(classOf[Array[MultiSampleMultiAlleleEvidence]])
+
+    kryo.register(classOf[Pileup])
```
timodonnell: another suspicious one -- we shouldn't be serializing these as @arahuja mentioned. ok by me though to address that in a separate PR

timodonnell:
LGTM

src/main/scala/org/hammerlab/guacamole/Serialization.scala
ryan-williams: Yes, and that's how I came up with many of the registrations here. The problem is that you generally only find out about one class per run (the first one that Kryo hits fails the stage), so it's very tedious to do dozens of runs to unearth all of these.

PR #443, opened 2016-04-18
LociSet / LociMap refactor

ryan-williams:
Individual commits should be self-contained and perhaps easier to review. Some highlights:

- https://github.com/hammerlab/guacamole/pull/443/commits/9f6126484c0e86a924c350586648322c34ed08b7: move `Loci*` classes into `set` and `map` pkgs, pull `Contig`/`Builder` classes out into own files.
- https://github.com/hammerlab/guacamole/pull/443/commits/78b7568cb62cac25dc3fd6375bbeed8db3fd80af: separate `LociSet` and `LociMap` implementations: `LociSet` previously wrapped a `LociMap[Long]` with all values set to `0`; now it wraps a `SortedMap[String, set.Contig]` analogous to `LociMap`'s `SortedMap[String, map.Contig]`. This allows clearer distinctions between set and map functionality, removes some circularity from the `Loci*` landscape that I found confusing.
- https://github.com/hammerlab/guacamole/pull/443/commits/c8e0b8d3e6e4394e356fd0c5c11a9ad51650cc8a: 
  - resolve `TODO` about `LociSet.take` implementation: assemble the two new `LociSet`s from existing `Contig`s, instead of completely deconstructing them into ranges and building two sets back up.
  - begin to distinguish between loci-string-parsing code paths (`LociParser`; contig lengths are required to obtain the `LociSet` via `.result`) and LociSet-building (`loci.set.Builder`; when creating LociSets from within existing LociSets or LociMaps).

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/443)


src/main/scala/org/hammerlab/guacamole/loci/SimpleRange.scala
```diff
@@ -0,0 +1,37 @@
...
+    override def hasNext: Boolean = i < end
+    override def next(): Long =
+      if (hasNext) { val result = i; i += 1; result }
+      else Iterator.empty.next()
+  }
+  /** Number of elements in the range. */
+  def length: Long = end - start
+
+  /** Comparisons between ranges. Order is DESCENDING (i.e. reversed) from by start. */
+  def compare(other: SimpleRange): Int = {
```
arahuja: Does this need an `override def`

src/main/scala/org/hammerlab/guacamole/loci/set/Contig.scala
```diff
@@ -0,0 +1,91 @@
...
+        doneTaking = true
+      }
+    }
+
+    (Contig(name, firstRanges), Contig(name, secondRanges))
+  }
+
+  /**
+    * Iterator over string representations of each range in the map, used to assemble (possibly truncated) .toString()
+    * output.
```
arahuja: Does this `(possibly truncated)`still apply?

src/main/scala/org/hammerlab/guacamole/loci/SimpleRange.scala
ryan-williams: It actually doesn't when it's overriding a `trait` method, but it's good to add on balance, good call; done!

src/main/scala/org/hammerlab/guacamole/loci/set/Contig.scala
ryan-williams: With the `TruncatedToString` interface I added, it does, indeed I added this comment I think; from here the "pieces" in the `Iterator[String]` don't know if the upstream caller is `truncatedString()` with a limit that will or won't result in truncation, or a `toString`, etc.

ryan-williams:
I added some comments and tweaks and squashed them in to existing commits, jfyi.

timodonnell:
Seems reasonable to me, thanks @ryan-williams .

Only suggestion is to maybe give `loci.map.Contig` and `loci.set.Contig` unique names so we don't have to always have import aliases like `import org.hammerlab.guacamole.loci.set.{Contig => LociSetContig}`. With or without that change I'm fine to merge this though

Separately, I think we're approaching diminishing returns on aesthetic refactoring to improve modularity or scala idiomaticness. Not saying that changes like this PR aren't valuable, just that where we are in the project right now I think getting Guac to run on large datasets, performance, and features on the issue list are better to focus on.

ryan-williams:
> Only suggestion is to maybe give loci.map.Contig and loci.set.Contig unique names so we don't have to always have import aliases like import org.hammerlab.guacamole.loci.set.{Contig => LociSetContig}.

I have gone back and forth on this…

- I like having separate `set` and `map` packages,
- I don't like the idea of repeating the package segments in class names (e.g. changing to `loci.set.LociSetContig`), and
- I like naming the classes with parallel structure across the sub-packages (i.e. `Contig`, `Builder`, `Serializer` in both).

Thinking on it further, aliasing the imports is only necessary in places where we're dealing with both `set` and `map` `Contig`s, which is currently:

1. in the kryo registrator, and
2. in `map.Contig` (the `inverse` method).

I've changed a third instance (a `WindowFlatMapUtils` usage of `LociSetContig`) to simply `Contig` in response to your comment, and I think addressing 1. by having GuacamoleKryoRegistrator delegate to sub-package-specific registrators is a good eventual follow-on improvement to what I have so far here (allowing for package-privatizing some internal classes as well as more clearly enumerating what classes are serde'd where, as has been discussed on #441).

So I think I'm going to leave it as is, for now, as it is a similar phenomenon to that which we are facing while juggling scala vs. java `Long`s in this part of the code-base. Using `import`s to call out which version we are using in a per-call-site-specific manner seems a better solution than sacrificing local brevity/clarity in what the classes themselves are named in the theoretically-narrowly-scoped packages they live in.

> Separately, I think we're approaching diminishing returns on aesthetic refactoring to improve modularity or scala idiomaticness. Not saying that changes like this PR aren't valuable, just that where we are in the project right now I think getting Guac to run on large datasets, performance, and features on the issue list are better to focus on.

I appreciate this perspective, but feel that this and other refactors are necessary to get Guac to run on large datasets, improve performance, and add features.

"Where we are in the project", at least wrt to the joint caller, is still "an order of magnitude slower than it should be + frequent OOM-deaths, neither of which has been root-caused".

When we started discussing ways to improve this a few weeks ago, our (many, good) ideas all involved delving into / rewriting core partitioning/iteration logic that is ~2yrs old, was written by versions of ourselves that understood Spark and Scala far less well than we do now, and the physical layout of which is/was a mess.

I'm confident in our ability to debug the Spark issues the joint caller is running into, but doing so will require abstractions that support varying our partitioning logic, tests that compare them, and instrumentation that helps us track down the mysterious orders-of-magnitude of memory and speed we're losing; those concerns have motivated the refactoring I've been doing, which I'm also anxious to start benefitting from.

PR #444, opened 2016-04-20
remove unused adam-cli dependency

ryan-williams:
h/t @fnothaft https://github.com/bigdatagenomics/adam/issues/1004#issuecomment-212478270

This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/444)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
     <java.version>1.8</java.version>
     <scala.version>2.10.5</scala.version>
     <scala.version.prefix>2.10</scala.version.prefix>
-    <hadoop.version>2.6.0</hadoop.version>
+    <hadoop.version>2.7.0</hadoop.version>
     <spark.version>1.6.1</spark.version>
     <maven.release.plugin.version>2.4.1</maven.release.plugin.version>
   </properties>
@@ -243,7 +243,13 @@
       <url>http://dl.bintray.com/non/maven</url>
     </repository>
   </repositories>
+
   <dependencies>
+    <dependency>
+      <groupId>com.esotericsoftware.kryo</groupId>
+      <artifactId>kryo</artifactId>
+      <version>2.21</version>
+    </dependency>
     <dependency>
       <groupId>args4j</groupId>
       <artifactId>args4j</artifactId>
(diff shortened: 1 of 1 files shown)


heuermh:
```
java.lang.IllegalAccessError: tried to access method
com.google.common.base.Stopwatch.<init>()V from class
org.apache.hadoop.mapreduce.lib.input.FileInputFormat
```
I've seen this one before, a dependency on Guava later than version 16.0.1 breaks Hadoop 2.6+ at runtime
https://github.com/bigdatagenomics/adam/blob/master/pom.xml#L514

arahuja:
This is in #433 as well with some other unused dependencies removed as well

timodonnell:
modulo travis error LGTM

ryan-williams:
Thanks for the tip @heuermh, that's interesting/annoying!

@arahuja any thoughts on why #433 is not failing in this way, given that it has the same change?

ryan-williams:
It seems that our workaround options are:

- bump guava down to 16.0.1
- bump hadoop down to 2.3.0

thoughts on those @timodonnell @arahuja ?

timodonnell:
My (pretty uninformed) thought would be that pinning to an old version of
guava is preferable to pinning to an old version of hadoop.

arahuja:
At some point, we shifted our guava shading from a few classes from `com.google.guava.common` to all of them. We should go back and selectively shade the guava classes we use.

ryan-williams:
@arahuja interesting, why would shading all of them be causing this issue? shouldn't that mean it doesn't matter what guava version we depended on, since we're shading it and our deps won't pick up our version transitively?

ryan-williams:
Turns out we can have it both ways by upgrading Hadoop to 2.7.0 (to pick up [HADOOP-11032](https://issues.apache.org/jira/browse/HADOOP-11032), which makes Hadoop play nice with Guava > 16.0.1) and declaring our commons-io dep as ≥ 2.3 (here I've used 2.4, [which Hadoop 2.7.0 depends on](https://github.com/apache/hadoop/blob/release-2.7.0/hadoop-project/pom.xml#L580-L584)). 

Previously we were relying on transitively picking up commons-io (version 1.3.2) from adam-core, despite using it directly; ADAM builds+tests fine with commons-io 2.4 so I may send them a PR to bump that version as well.

While here, I also declared our kryo dep, set to version 2.21, which Spark uses.

arahuja:
Cool @ryan-williams thanks for digging further! LGTM

fnothaft:
> Previously we were relying on transitively picking up commons-io (version 1.3.2) from adam-core, despite using it directly; ADAM builds+tests fine with commons-io 2.4 so I may send them a PR to bump that version as well.

That'd be great!

PR #445, opened 2016-04-20
Fix test orphan-tmp-file problem

ryan-williams:
was littering working directory with TestUtil* and .TestUtil* files.

This was originally in #443 but I've split it out here since it's unrelated.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/445)


timodonnell:
LGTM

PR #448, opened 2016-04-22
Small fixes to joint caller required to load vcf output into pyvcf

timodonnell:
A few joint caller VCF output fixes
 * change some attributes that had spaces in the values, which was confusing `pyvcf`
 * add reference fasta path to VCF output so `varcode` can guess the reference genome (required adding a field to `ReferenceGenome` to record this in)
 * support user specified metadata in VCF output (example use: record which guacamole git sha1 was used to generate the file)

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/448)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/jointcaller/SomaticJointCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/jointcaller/SomaticJointCaller.scala
@@ -157,6 +157,8 @@ object SomaticJoint {
                 includeFiltered: Boolean = false,
                 distributedUtilArguments: LociPartitionUtils.Arguments = new LociPartitionUtils.Arguments {}): RDD[MultiSampleMultiAlleleEvidence] = {
 
+    assume(loci.nonEmpty)
+
     // When mapping over pileups, at locus x we call variants at locus x + 1. Therefore we subtract 1 from the user-
     // specified loci.
     val broadcastForceCallLoci = sc.broadcast(forceCallLoci)
--- a/src/main/scala/org/hammerlab/guacamole/commands/jointcaller/SomaticJointCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/jointcaller/SomaticJointCaller.scala
@@ -45,6 +45,8 @@ object SomaticJoint {
     @Args4jOption(name = "-q", usage = "Quiet: less stdout")
     var quiet: Boolean = false
 
+    // For example:
+    //  --header-metadata kind=tuning_test version=4
     @Args4jOption(name = "--header-metadata",
       usage = "Extra header metadata for VCF output in format KEY=VALUE KEY=VALUE ...",
       handler = classOf[StringArrayOptionHandler])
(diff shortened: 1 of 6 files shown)


src/main/scala/org/hammerlab/guacamole/reference/ReferenceBroadcast.scala
```diff
@@ -156,6 +158,10 @@ object ReferenceBroadcast {
       result
     }
   }
+
+  def apply(broadcastedContigs: Map[String, ContigSequence]): ReferenceBroadcast = {
+    ReferenceBroadcast(broadcastedContigs, source = "unknown_source")
```
arahuja: Can we make this an option instead of putting in a random string?

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/SomaticJointCaller.scala
```diff
@@ -43,6 +44,11 @@ object SomaticJoint {
 
     @Args4jOption(name = "-q", usage = "Quiet: less stdout")
     var quiet: Boolean = false
+
+    @Args4jOption(name = "--header-metadata",
+      usage = "Extra header metadata for VCF output in format KEY=VALUE KEY=VALUE ...",
```
arahuja: Is there an example of this?

src/main/scala/org/hammerlab/guacamole/reference/ReferenceBroadcast.scala
timodonnell: Sure, done.

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/SomaticJointCaller.scala
timodonnell: Added an example in a comment

timodonnell:
Thanks for the feedback @arahuja , updated to address comments

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/VCFOutput.scala
```diff
@@ -79,7 +79,9 @@ object VCFOutput {
     extraHeaderMetadata.foreach(kv => {
       header.addMetaDataLine(new VCFHeaderLine(kv._1, kv._2))
     })
-    header.addMetaDataLine(new VCFHeaderLine("reference", reference.source))
+    if (reference.source.isDefined) {
```
arahuja: nit: this could also be `reference.source.foreach( header.addMetaDataLine(new VCFHeaderLine("reference", _)) )`

fine with leaving it as is

arahuja:
Cool, thanks @timodonnell minor nit, but fine leaving it

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/VCFOutput.scala
timodonnell: Yeah, totally subjective but I think as-is is actually clearer, so I think I'll leave it

ryan-williams:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/448#-:-KGDH7tesKvi4jr6POkd:1417088721)*

PR #451, opened 2016-04-25
document running test-"main" apps

ryan-williams:
Responding to concern that was raised in #430, this makes it clear how to run the "standalone" apps under `src/test`

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/451)


timodonnell:
LGTM, thanks @ryan-williams . I just made a PR that extends this current one to update the partial fasta generation based on these changes (and also fix a bug in partial fasta generation that had crept in when `LociSet.contigs` changed, which was making it silently produce invalid output). Feel free to merge my PR into master or merge my PR into this PR, whichever is easier

PR #452, opened 2016-04-25
move assertBases helpers into own class

ryan-williams:
move these helpers out of kitchen-sink `TestUtil`, make it not inherit from `Matchers`

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/452)


arahuja:
Any reason for this in particular?

ryan-williams:
Just hygience; having a kitchen-sink `TestUtil` object with lots of unrelated functionality is worse, imho, than having more narrowly-scoped objects wrapping more specific pieces of functionality.

While mucking around in #451 I wondered what `Matchers` functionality was being used in `TestUtil` to warrant the latter's inheriting the former, found these couple cases, and decided to pull them out.

arahuja:
LGTM

PR #453, opened 2016-04-26
factor out, comment some kryo registrations

ryan-williams:
Some follow-on work to #441:

- add hook for declaring test-specific kryo registration
- relegate `Pileup{,Element}` registration to the test that requires it; leave it out of production code.
- factor joint-caller registrations out to separate registrar.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/453)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/jointcaller/kryo/Registrar.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/jointcaller/kryo/Registrar.scala
@@ -1,0 +1,38 @@
+package org.hammerlab.guacamole.commands.jointcaller.kryo
+
+import com.esotericsoftware.kryo.Kryo
+import org.apache.spark.serializer.KryoRegistrator
+import org.hammerlab.guacamole.commands.jointcaller.Parameters.SomaticGenotypePolicy
+import org.hammerlab.guacamole.commands.jointcaller.annotation.{InsufficientNormal, MultiSampleAnnotations, SingleSampleAnnotations, StrandBias}
+import org.hammerlab.guacamole.commands.jointcaller.evidence.{MultiSampleMultiAlleleEvidence, MultiSampleSingleAlleleEvidence, NormalDNASingleSampleSingleAlleleEvidence, TumorDNASingleSampleSingleAlleleEvidence, TumorRNASingleSampleSingleAlleleEvidence}
+import org.hammerlab.guacamole.commands.jointcaller.{AlleleAtLocus, Input, InputCollection, Parameters}
+
+class Registrar extends KryoRegistrator {
+  override def registerClasses(kryo: Kryo): Unit = {
+    kryo.register(classOf[NormalDNASingleSampleSingleAlleleEvidence])
+    kryo.register(classOf[TumorDNASingleSampleSingleAlleleEvidence])
+
+    // These are needed for MultiSampleSingleAlleleEvidence; perTumorDnaSampleTopMixtures and probably other fields.
+    kryo.register(Class.forName("scala.collection.immutable.MapLike$$anon$2"))  // MapLike.MappedValues?
+    kryo.register(Class.forName("org.hammerlab.guacamole.commands.jointcaller.evidence.MultiSampleSingleAlleleEvidence$$anonfun$16"))
+    kryo.register(Class.forName("org.hammerlab.guacamole.commands.jointcaller.evidence.MultiSampleSingleAlleleEvidence$$anonfun$10"))
+
+    kryo.register(classOf[Parameters])
+    kryo.register(Class.forName("scala.Enumeration$Val"))
+    kryo.register(SomaticGenotypePolicy.getClass)
+    kryo.register(classOf[MultiSampleAnnotations])
+    kryo.register(classOf[InputCollection])
+    kryo.register(classOf[Input])
+    kryo.register(Input.Analyte.getClass)
+    kryo.register(Input.TissueType.getClass)
+    kryo.register(classOf[InsufficientNormal])
+    kryo.register(classOf[TumorRNASingleSampleSingleAlleleEvidence])
+    kryo.register(classOf[Parameters.SomaticGenotypePolicy.Value])
+    kryo.register(classOf[StrandBias])
+    kryo.register(classOf[SingleSampleAnnotations])
+    kryo.register(classOf[AlleleAtLocus])
+    kryo.register(classOf[MultiSampleSingleAlleleEvidence])
+    kryo.register(classOf[MultiSampleMultiAlleleEvidence])
+    kryo.register(classOf[Array[MultiSampleMultiAlleleEvidence]])
+  }
+}
--- a/src/main/scala/org/hammerlab/guacamole/commands/jointcaller/kryo/Registrar.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/jointcaller/kryo/Registrar.scala
@@ -16,6 +16,7 @@ class Registrar extends KryoRegistrator {
     kryo.register(Class.forName("scala.collection.immutable.MapLike$$anon$2"))  // MapLike.MappedValues?
     kryo.register(Class.forName("org.hammerlab.guacamole.commands.jointcaller.evidence.MultiSampleSingleAlleleEvidence$$anonfun$16"))
     kryo.register(Class.forName("org.hammerlab.guacamole.commands.jointcaller.evidence.MultiSampleSingleAlleleEvidence$$anonfun$10"))
+    kryo.register(Map.empty.getClass)
 
     kryo.register(classOf[Parameters])
     kryo.register(Class.forName("scala.Enumeration$Val"))
--- a/scripts/guacamole-spark-defaults.conf
+++ b/scripts/guacamole-spark-defaults.conf
@@ -1,4 +1,4 @@
 spark.serializer               org.apache.spark.serializer.KryoSerializer
-spark.kryo.registrator         org.hammerlab.guacamole.kryo.GuacamoleKryoRegistrator
+spark.kryo.registrator         org.hammerlab.guacamole.kryo.Registrar
 spark.kryoserializer.buffer.mb 4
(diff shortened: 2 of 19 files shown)


src/main/scala/org/hammerlab/guacamole/commands/jointcaller/kryo/Registrar.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.hammerlab.guacamole.commands.jointcaller.{AlleleAtLocus, Input, InputCollection, Parameters}
+
+class Registrar extends KryoRegistrator {
+  override def registerClasses(kryo: Kryo): Unit = {
+    kryo.register(classOf[NormalDNASingleSampleSingleAlleleEvidence])
+    kryo.register(classOf[TumorDNASingleSampleSingleAlleleEvidence])
+
+    // These are needed for MultiSampleSingleAlleleEvidence; perTumorDnaSampleTopMixtures and probably other fields.
+    kryo.register(Class.forName("scala.collection.immutable.MapLike$$anon$2"))  // MapLike.MappedValues?
+    kryo.register(Class.forName("org.hammerlab.guacamole.commands.jointcaller.evidence.MultiSampleSingleAlleleEvidence$$anonfun$16"))
```
timodonnell: Any idea what this is about? Didn't realize we had to have kryo registrations for closures (?). These anonymous names seem prone to changing as seemingly unrelated code changes

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/kryo/Registrar.scala
```diff
@@ -0,0 +1,38 @@
+package org.hammerlab.guacamole.commands.jointcaller.kryo
+
+import com.esotericsoftware.kryo.Kryo
+import org.apache.spark.serializer.KryoRegistrator
+import org.hammerlab.guacamole.commands.jointcaller.Parameters.SomaticGenotypePolicy
+import org.hammerlab.guacamole.commands.jointcaller.annotation.{InsufficientNormal, MultiSampleAnnotations, SingleSampleAnnotations, StrandBias}
+import org.hammerlab.guacamole.commands.jointcaller.evidence.{MultiSampleMultiAlleleEvidence, MultiSampleSingleAlleleEvidence, NormalDNASingleSampleSingleAlleleEvidence, TumorDNASingleSampleSingleAlleleEvidence, TumorRNASingleSampleSingleAlleleEvidence}
+import org.hammerlab.guacamole.commands.jointcaller.{AlleleAtLocus, Input, InputCollection, Parameters}
+
+class Registrar extends KryoRegistrator {
```
timodonnell: just curious, why make this a class instead of an object?

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/kryo/Registrar.scala
```diff
@@ -0,0 +1,38 @@
...
+    kryo.register(Class.forName("org.hammerlab.guacamole.commands.jointcaller.evidence.MultiSampleSingleAlleleEvidence$$anonfun$16"))
+    kryo.register(Class.forName("org.hammerlab.guacamole.commands.jointcaller.evidence.MultiSampleSingleAlleleEvidence$$anonfun$10"))
+
+    kryo.register(classOf[Parameters])
+    kryo.register(Class.forName("scala.Enumeration$Val"))
+    kryo.register(SomaticGenotypePolicy.getClass)
+    kryo.register(classOf[MultiSampleAnnotations])
+    kryo.register(classOf[InputCollection])
+    kryo.register(classOf[Input])
+    kryo.register(Input.Analyte.getClass)
```
timodonnell: Also curious what the difference between `classOf` and `getClass` is here

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/kryo/Registrar.scala
ryan-williams: Yea I'm not even sure what anonymous inner-classes or methods these are referring to; I discussed this a little bit when I added them: https://github.com/hammerlab/guacamole/pull/441#issuecomment-211189236. 

I'm not sure what kinds of changes would cause these magic strings to change, but they would break tests so we shouldn't have to worry about them changing without us noticing.

Other options are:
* `@transient` or `def` some of the `val`s in `MultiSampleSingleAlleleEvidence`.
* write a dedicated `Serializer` for `MultiSampleSingleAlleleEvidence` and/or classes that are currently getting ser'd as part of its closure.

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/kryo/Registrar.scala
ryan-williams: `classOf` is what makes this work when registering `class`s, but it's not defined for `object`s. Conversely, `getClass` gets the class of an `object` but doesn't work for `class`s.

```
scala> object Foo
scala> classOf[Foo]
<console>:12: error: not found: type Foo
       classOf[Foo]
               ^
scala> Foo.getClass
res1: Class[_ <: Foo.type] = class Foo$

scala> class Bar
scala> classOf[Bar]
res2: Class[Bar] = class Bar
scala> Bar.getClass
<console>:12: error: not found: value Bar
       Bar.getClass
       ^
```

I learned this through some trial and error and looking at other projects' Kryo-registration flows around the internet 😄

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/kryo/Registrar.scala
ryan-williams: I thought about `object`ing this, but I like the consistency with other "registrators", and also having an eye towards a possible future where modules handle their own kryo registration, and can e.g. run their tests with just their registrar being instantiated, not the global guacamole one that registers everything. It's a very minor point though, either way.

any registrar that is passed to Spark must be a zero-argument class, not an object, due to [the way spark instantiates them](https://github.com/apache/spark/blob/v1.6.1/core/src/main/scala/org/apache/spark/serializer/KryoSerializer.scala#L123).

e5c:
Reviewed 8 of 8 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrar.scala, line 99 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/453#-KGNfVLOvnR2Ca4ASlQd:-KGNfVLOvnR2Ca4ASlQe:-946956390) ([raw file](https://github.com/hammerlab/guacamole/blob/6b52856aba47d3bee6b031f7e4201000c9e1df01/src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrar.scala#L99)):*
OOC are there benefits to registering these explicitly instead of as `Array[_]`? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/453)*

ryan-williams:
*[src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrar.scala, line 99 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/453#-KGNfVLOvnR2Ca4ASlQd:-KGNjp70Tux_qRdByw-o:1302117005) ([raw file](https://github.com/hammerlab/guacamole/blob/6b52856aba47d3bee6b031f7e4201000c9e1df01/src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrar.scala#L99)):*
good q; there's something I don't understand about the implications of each alternative. I believe I've observed cases where the `[_]` is registered and yet I get unregistered-class errors about a specific type. Let me see if I can reproduce that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/453)*

ryan-williams:
*[src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrar.scala, line 99 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/453#-KGNfVLOvnR2Ca4ASlQd:-KGOti9b1MpokiiBzqMB:1366831064) ([raw file](https://github.com/hammerlab/guacamole/blob/6b52856aba47d3bee6b031f7e4201000c9e1df01/src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrar.scala#L99)):*
OK, I tested it and am having a hard time finding cases where the `[_]` actually fixes any not-registered errors.

I needed it in [a couple of places in `PileupFlatMapUtilsSuite`](https://github.com/hammerlab/guacamole/pull/453/files#diff-43113659962f2789a700092a0b533b9aR42); a generic `Map` seems to get pulled in when serializing some `ClassTag`s, and the `Array[Seq[_]]` maybe works because at two levels deep type erasure has kicked in…?

There's an outstanding `PairedRead[_]` in `GuacamoleKryoRegistrar` that seems to be having an effect; it's still a bit mysterious.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/453)*

ryan-williams:
*[src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrar.scala, line 99 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/453#-KGNfVLOvnR2Ca4ASlQd:-KGOuWAhFGNhNmKZF2H2:583640279) ([raw file](https://github.com/hammerlab/guacamole/blob/6b52856aba47d3bee6b031f7e4201000c9e1df01/src/main/scala/org/hammerlab/guacamole/kryo/GuacamoleKryoRegistrar.scala#L99)):*
Also note that most of these registrations are gone in my newest, most improved HEAD of this PR!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/453)*

arahuja:
@ryan-williams this good to merge?

ryan-williams:
as far as I'm concerned, yes! sgty?

PR #455, opened 2016-04-27
Fix varlens test deletion parsing

arahuja:
I think there was an issue with the CSV parsing in the tests here. I'd again advocate to store the data in a  canonical format (VCF or with standard coordinates) rather than relying on custom test code to load it.

@timodonnell Is this right here? I'm not sure which side of the interbase coordinates are inclusive

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/455)

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/VariantComparisonUtils.scala
+++ b/src/test/scala/org/hammerlab/guacamole/VariantComparisonUtils.scala
@@ -29,10 +29,10 @@ case class VariantFromVarlensCSV(
       (interbaseStart, ref, alt)
     } else {
       // Deletion.
-      val refSeqeunce = Bases.basesToString(
+      val refSequence = Bases.basesToString(
         reference.getReferenceSequence(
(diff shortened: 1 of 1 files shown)


arahuja:
Also, is there any documentation on how to generate these test files? There didn't seem to be any insertions in the CSV file, is that right? And the decoy ones are spiked in?

timodonnell:
Yeah, I agree it'd be better to store these as VCFs. The only reason it's done this way is that I didn't have VCFs for these samples, only csv files, and had no convenient way of generating a VCF. Definitely open to doing a conversion though.

I think you're right there is something wrong here, will look into it. In general I'm taking "interbase" to mean 0-indexed, inclusive on start, exclusive on end (i.e. what we do in guacamole). This part of the code is complicated by the fact that we're also trying to convert variants like `ref='A'` `alt=''` into VCF-style variants like `ref=CA` `alt=C`

timodonnell:
I just spot checked this for 3 variants and I think your change is correct. LGTM to merge

PR #418, opened 2016-04-01
bash script improvements

ryan-williams:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/418)

PR #457, opened 2016-04-28
Bash script improvements

ryan-williams:
Rebased #418; I ran all the modified scripts at least far enough to verify that I've not broken anything bash-wise; I can't actually get the `run_other_callers` commands to be happy with reference-chromosome names, and I killed `generate-partial-fasta…` after it was running for a while.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/457)


ryan-williams:
This was previously lgtm'd by @timodonnell modulo sanity-running the scripts, which I've now done, so I'm going to merge this!

PR #458, opened 2016-04-28
Use shared spark context in tests

ryan-williams:
Uses @holdenk's spark-testing-base.

I ran `mvn test` 10 times (after first doing a `mvn test-compile`) with and without this change, to measure the timing difference:

Before:

```
01:47 min
01:47 min
01:46 min
01:47 min
01:47 min
01:47 min
01:49 min
01:52 min
01:53 min
01:43 min
```

After:

```
01:29 min
01:32 min
01:31 min
01:32 min
01:29 min
01:33 min
01:31 min
01:30 min
01:32 min
01:33 min
```

~15% faster!

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/458)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -306,19 +306,6 @@
       <version>3.0</version>
     </dependency>
     <dependency>
-      <groupId>org.scalatest</groupId>
-      <artifactId>scalatest_${scala.version.prefix}</artifactId>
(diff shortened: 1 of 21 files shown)


arahuja:
LGTM! Thanks!

PR #460, opened 2016-04-29
LociPartitioningUtils cleanups

ryan-williams:
- be clear about where we are using `Long`s (micro-partitions) vs. `Int`s
  (Spark partitions)
- documentation touch-ups
- PerSample-ify RDDs
- use "partition" instead of "task"

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/460)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -276,6 +276,11 @@
       <version>1.2.0</version>
     </dependency>
     <dependency>
+      <groupId>org.spire-math</groupId>
+      <artifactId>spire_2.10</artifactId>
+      <version>0.11.0</version>
+    </dependency>
+    <dependency>
       <groupId>org.apache.spark</groupId>
       <artifactId>spark-core_${scala.version.prefix}</artifactId>
       <version>${spark.version}</version>
--- a/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
@@ -8,10 +8,9 @@ import org.hammerlab.guacamole.Common
 import org.hammerlab.guacamole.Common.GermlineCallerArgs
 import org.hammerlab.guacamole.alignment.AffineGapPenaltyAlignment
 import org.hammerlab.guacamole.assembly.AssemblyUtils
-import org.hammerlab.guacamole.distributed.LociPartitionUtils.partitionLociAccordingToArgs
+import org.hammerlab.guacamole.distributed.LociPartitionUtils.{LociPartitioning, partitionLociAccordingToArgs}
 import org.hammerlab.guacamole.distributed.WindowFlatMapUtils.windowFlatMapWithState
 import org.hammerlab.guacamole.likelihood.Likelihood
-import org.hammerlab.guacamole.loci.map.LociMap
 import org.hammerlab.guacamole.logging.DelayedMessages
 import org.hammerlab.guacamole.logging.LoggingUtils.progress
 import org.hammerlab.guacamole.pileup.Pileup
@@ -119,7 +118,7 @@ object GermlineAssemblyCaller {
                                  minOccurrence: Int,
                                  minAreaVaf: Float,
                                  reference: ReferenceBroadcast,
-                                 lociPartitions: LociMap[Long],
+                                 lociPartitions: LociPartitioning,
                                  minAltReads: Int = 2,
                                  minPhredScaledLikelihood: Int = 0,
                                  shortcutAssembly: Boolean = false): RDD[CalledAllele] = {
--- a/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala
@@ -99,8 +99,7 @@ object SomaticStandard {
       val lociPartitions = partitionLociAccordingToArgs(
         args,
         loci.result(normalReads.contigLengths),
-        tumorReads.mappedReads,
-        normalReads.mappedReads
+        Vector(tumorReads.mappedReads, normalReads.mappedReads)
       )
 
       var potentialGenotypes: RDD[CalledSomaticAllele] =
--- a/src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
@@ -10,9 +10,8 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.storage.StorageLevel
 import org.hammerlab.guacamole._
 import org.hammerlab.guacamole.distributed.LociPartitionUtils
-import org.hammerlab.guacamole.distributed.LociPartitionUtils.partitionLociAccordingToArgs
+import org.hammerlab.guacamole.distributed.LociPartitionUtils.{LociPartitioning, partitionLociAccordingToArgs}
 import org.hammerlab.guacamole.distributed.PileupFlatMapUtils.pileupFlatMap
-import org.hammerlab.guacamole.loci.map.LociMap
 import org.hammerlab.guacamole.logging.LoggingUtils.progress
 import org.hammerlab.guacamole.pileup.Pileup
 import org.hammerlab.guacamole.reads.{InputFilters, MappedRead, ReadLoadingConfigArgs}
@@ -123,7 +122,7 @@ object VAFHistogram {
       val lociPartitions = partitionLociAccordingToArgs(
         args,
         loci.result(readSets(0).contigLengths),
-        readSets(0).mappedReads // Use the first set of reads as a proxy for read depth
+        readSets(0).mappedReads  // Use the first set of reads as a proxy for read depth
       )
 
       val minReadDepth = args.minReadDepth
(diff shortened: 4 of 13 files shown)


arahuja:
LGTM - a few questions.

OOC, what does the Spire `Intergral` type add? If we are adding a some Spire dependency we should also had use the non-empty Seq to replace the `PerSample` since that will never be empty and we do a fair amount of asserts on it.

Also, if we are in here, should we just drop `--parallelism` to remove the confusion with `spark.default.parlallelism`? 

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/460#-:-KGlzILQo5AKnWN55fd0:1452617147)*

arahuja:
*[src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala, line 175 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/460#-KGlyfoAIts3jA-Zwj7J:-KGlyfoAIts3jA-Zwj7K:-125223901) ([raw file](https://github.com/hammerlab/guacamole/blob/24dad84011373de8b6c4117759d5fac900c42f71/src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala#L175)):*
Not in this change, but it's too late for this assert given the `regionRDDs(0).sparkContext` above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/460)*

ryan-williams:
> what does the Spire Intergral type add?

it allows us to call `partitionLociUniformly` parameterized by either `Int` or `Long`, as the call-site dictates. I want to make that explicit in the types, as the call-site implications of the `Long` (micro-partition) version are distinct from those of the `Int` (partition) version, though the code is the same (when expressed this way, at least!).

I fussed with Scala `Integral`/`Numeric` types+implicits and concluded they can't be made to do what I want here; spire seems dedicated to filling in the holes in the Scala stdlib's support for this sort of patter. lmk if you have other thoughts about this!

> we should also had use the non-empty Seq to replace the PerSample since that will never be empty and we do a fair amount of asserts on it.

I like the sound of this, am happy to do it, and think it should happen as another PR :)

> should we just drop --parallelism to remove the confusion with spark.default.parlallelism?

good q; again I'd vote for making that another change, if we want it. I don't feel a pressing need to drop that param in favor of `spark.default.parallelism`, largely because I've grown less fond of relying on `spark.default.parallelism` in general.

The assumption that there is a sane default number of partitions that can be applied to arbitrary different stages in an app is pretty flawed and is the root of some of our troubles over the years, I've come to feel. 

In a perfect world, we'd plumb through num-partitions values to our different RDD operations, having derived them from rough estimates of the desired size of partitions (in either records or bytes), which would in turn be specified by cmdline params.

That's all further-future work, but shifting something we already specify directly (that is narrowly scoped to one sort of RDD operation) back to the `spark.default.parallelism` catch-all would actually be a step backward, according to my current thinking.

---

Review status: 0 of 13 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala, line 175 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/460#-KGlyfoAIts3jA-Zwj7J:-KGmPiwegoDOs9Mwc5NY:-499146022) ([raw file](https://github.com/hammerlab/guacamole/blob/24dad84011373de8b6c4117759d5fac900c42f71/src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala#L175)):*
Done!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/460#-:-KGmPvq_sM-rDrJk16Bc:-414665295)*

timodonnell:
Review status: 0 of 13 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala, line 11 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/460#-KGmUe7Q_g170XAsQcDe:-KGmUe7R9yUGYfYTJqs0:-1445042622) ([raw file](https://github.com/hammerlab/guacamole/blob/24dad84011373de8b6c4117759d5fac900c42f71/src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala#L11)):*
My preference would be to avoid implicit conversions and just call them explicitly. I know implicit conversions are common in scala but I think they lead to less clear code, interested what others think

---

*[src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala, line 36 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/460#-KGmVEHHDViaIvbAixf0:-KGmVEHHDViaIvbAixf1:-316437389) ([raw file](https://github.com/hammerlab/guacamole/blob/24dad84011373de8b6c4117759d5fac900c42f71/src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala#L36)):*
I'd rather spell out `Index`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/460)*

arahuja:
> I like the sound of this, am happy to do it, and think it should happen as another PR :)

Ah yea, mostly thinking out loud

>  but shifting something we already specify directly (that is narrowly scoped to one sort of RDD operation) back to the spark.default.parallelism catch-all would actually be a step backward, according to my current thinking.

OK, I could see that. Going the other way then (and again not related to this) perhaps we remove the fallback and make the argument required.

ryan-williams:
@arahuja 

> OK, I could see that. Going the other way then (and again not related to this) perhaps we remove the fallback and make the argument required.

In some world where we've removed our reliance on `spark.default.parallelism` for everything else, sure. I don't feel like we need to mess with it for now though.

ryan-williams:
Review status: 0 of 13 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala, line 11 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/460#-KGmUe7Q_g170XAsQcDe:-KGmd7gPCVf6YmL1lVoM:990664606) ([raw file](https://github.com/hammerlab/guacamole/blob/24dad84011373de8b6c4117759d5fac900c42f71/src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala#L11)):*
mm, these are maybe not the kind of implicits you're thinking of.

[Here is a dummy commit I made that unrolls them all](https://github.com/ryan-williams/guacamole/commit/9128dec4c2b46fe59af09e571d7abb257b21cc85). I think most people would consider it an anti-pattern to use a library like spire but work around the implicits it provides; the implicits are basically the point of the library, and get you strong type guarantees for concepts like `Int|Long`, without messing up your code.

---

*[src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala, line 36 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/460#-KGmVEHHDViaIvbAixf0:-KGmd0jfyYgZM8LaRa4y:1395134907) ([raw file](https://github.com/hammerlab/guacamole/blob/24dad84011373de8b6c4117759d5fac900c42f71/src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala#L36)):*
Done. This should have actually been `NumPartitions` to be more clear, so I've made it so, but unrolled the `Idx`s elsewhere anyway.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/460)*

arahuja:
*[src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala, line 175 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/460#-KGlyfoAIts3jA-Zwj7J:-KGwLLrONmfIAZcebSeU:165686682) ([raw file](https://github.com/hammerlab/guacamole/blob/24dad84011373de8b6c4117759d5fac900c42f71/src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala#L175)):*
No reason to hold up the PR on this since it's not really related - but I didn't see anything changed on this? Just in case you were expecting it to be

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/460)*

ryan-williams:
Review status: 0 of 13 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala, line 175 \[r2\]](https://reviewable.io:443/reviews/hammerlab/guacamole/460#-KGlyfoAIts3jA-Zwj7J:-KGwO-m1Tiafmm1imk16:-1432849092) ([raw file](https://github.com/hammerlab/guacamole/blob/24dad84011373de8b6c4117759d5fac900c42f71/src/main/scala/org/hammerlab/guacamole/distributed/LociPartitionUtils.scala#L175)):*
I think you'll see it if you view more recent diffs on this PR; I think the permalink to this thread maybe goes to its original revision range?

I see the updated version when I go here https://reviewable.io/reviews/hammerlab/guacamole/460



---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/460)*

arahuja:
@ryan-williams It still seems to be there if I look Github file's changed:
image

Maybe we are just talking about different things? I had commented on `assume(regionRDDs.nonempty)`

ryan-williams:
You noted that the `assume(regionRDDs.nonempty)` was coming after a reference to `regionRDDs(0)`:



Now I've reordered things so that they're in the correct order:



That's the right outcome, right? I guess you're saying you meant I should remove the `assume`?

arahuja:
ah ok yea - I was looking at that thinking the statement was still there and didn't notice the other line moved, sounds good!

PR #462, opened 2016-04-30
move Bases, CigarUtils into `util` pkg

ryan-williams:
…and out of the root pkg

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/462)


arahuja:
👍

PR #463, opened 2016-04-30
rename HasReferenceRegion → reference.Region

ryan-williams:
also move region/lociset intersection check into lociset

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/463)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/HasReferenceRegion.scala
+++ b/src/main/scala/org/hammerlab/guacamole/HasReferenceRegion.scala
@@ -38,16 +38,6 @@ trait HasReferenceRegion {
   val end: Long
 
   /**
-   * Does this region overlap any of the given loci, with halfWindowSize padding?
-   *
-   * The padding is inclusive on both ends, so for example if halfWindowSize=1, then a region would overlap if it
-   * included any of these three positions: locus - 1, locus, and locus + 1.
-   */
-  def overlapsLociSet(loci: LociSet, halfWindowSize: Long = 0): Boolean = {
-    loci.onContig(referenceContig).intersects(math.max(0, start - halfWindowSize), end + halfWindowSize)
-  }
-
-  /**
    * Does the region overlap the given locus, with halfWindowSize padding?
    */
   def overlapsLocus(locus: Long, halfWindowSize: Long = 0): Boolean = {
--- a/src/main/scala/org/hammerlab/guacamole/commands/jointcaller/evidence/MultiSampleMultiAlleleEvidence.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/jointcaller/evidence/MultiSampleMultiAlleleEvidence.scala
@@ -4,7 +4,7 @@ import org.hammerlab.guacamole._
 import org.hammerlab.guacamole.commands.jointcaller.pileup_summarization.{MultiplePileupStats, PileupStats}
 import org.hammerlab.guacamole.commands.jointcaller.{AlleleAtLocus, InputCollection, Parameters}
 import org.hammerlab.guacamole.pileup.Pileup
-import org.hammerlab.guacamole.reference.ReferenceBroadcast
+import org.hammerlab.guacamole.reference.{ReferenceBroadcast, ReferenceRegion}
 import org.hammerlab.guacamole.util.Bases
 
 /**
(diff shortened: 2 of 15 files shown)


arahuja:
Can you explain this rename further? Was never a huge fan of `HasReferenceRegion` but I don't think this is clearer either. At least including what it is a region of would be useful? Would `ReferenceCoordinaets` or `ReferenceRegion` be better? `Region` is fairly ambiguous when reading the code

ryan-williams:
The argument for `reference.Region` over `reference.ReferenceRegion` is:

- the package name is already `reference`. Repeating pieces of the path in the basename is generally an anti-pattern imho; it's not the job of a class to name itself something that will remain unambiguous when downstream users import it, as its basename, alongside arbitrary other libraries. Within the context of a `reference` package, `Region` is unambiguous and the right name.
- we ubiquitously refer to "regions" around the codebase, e.g. [WindowFlatMapUtils](http://cl.ly/300O2Z2k2K1M/Screen%20Shot%202016-04-30%20at%207.20.39%20PM.png), [LociPartitionUtils](http://cl.ly/1j3G2e0U0C1q/Screen%20Shot%202016-04-30%20at%207.24.02%20PM.png), etc. There isn't really another "region" in play, afaik.
- a similar discussion happened on #443, because we have `loci.map.Contig` and `loci.set.Contig`.
  - that's a trickier one since the collision is between classes in two sibling packages, not in e.g. two separate repos.
  - even so, I think it makes more sense to name them in a way that is most sane within their local context.
    - generally code doesn't import two things from anywhere with the same basename.
    - if it does, it's converting between them or something and should use the `=>` "import as" syntax to disambiguate.

Anyway, it's a minor point and I can make it `ReferenceRegion` if you the above has not convinced you! 😃

arahuja:
> it's not the job of a class to name itself something that will remain unambiguous when downstream users import it, as its basename, alongside arbitrary other libraries. Within the context of a reference package, Region is unambiguous and the right name

I'd probably disagree with this since when reading other code, you are not reading the imports first and then reading the class names. Making the class names unambiguous aids in reading the code and understanding snippets without reading the file from start to finish, at least to me. 

> we ubiquitously refer to "regions" around the codebase, e.g. WindowFlatMapUtils, LociPartitionUtils, etc. There isn't really another "region" in play, afaik.

Yea I think (and was going to comment on the other PR) this is pretty confusing throughout too. I always trip up when reading this section. It used to be reads and then was 'reference-mapped objects' and then we debated and moved it to regions. But it's not really regions we are mapping but thing have coordinates or reference fields.

Anyways, I'm fine with whatever, just my .02

ryan-williams:
Cool, I've renamed it to `ReferenceRegion`. A couple inlines, for discussion's sake:

> Making the class names unambiguous aids in reading the code…

I agree with you that readers should never have to wonder (or scroll up to import statements to find out) which package's `Region`, [`Allele`](https://github.com/hammerlab/guacamole/blob/fd61a07e5ddc874e53856368591c1bd3499dbd9c/src/test/scala/org/hammerlab/guacamole/variants/VariantComparisonTest.scala#L40), [`HashMap`](https://github.com/hammerlab/guacamole/blob/1548a14bb9c801c75a2d38f0e93e20cb5528a23b/src/main/scala/org/hammerlab/guacamole/reference/ReferenceBroadcast.scala#L57), [`Contig`](https://github.com/hammerlab/guacamole/blob/ea30984aeb04537fc5f96f660838a4c0f0e4766d/src/main/scala/org/hammerlab/guacamole/loci/map/Contig.scala#L53), [`Range`, or `Long`](https://github.com/hammerlab/guacamole/blob/832f7cdea0e9a0f828a8272479537934ec6e04ba/src/main/scala/org/hammerlab/guacamole/loci/set/LociSet.scala#L132) is being referred to by code they're reading.

However, I would rephrase your comment as: making _imported_ names unambiguous is important when reading code that has imported them (unambiguity at the declaration site is also important but I don't think `reference.Region` being unambiguous at the declaration site is being disputed here).

To that end, how much detail is required to `import` a given class unambiguously depends on circumstances at the `import` site, and so should be decided at each `import` site. We _have to_ use the `=>` syntax when two classes with the same basename are in scope and, shy of that, if a basename still feels ambiguous in a given context, we have the option to respond in two ways:

- `import ….reference` then use `reference.Region` (like we frequently do with e.g. `mutable.HashMap`), or 
- `import ….reference.{Region => ReferenceRegion}`
  - `import ….reference.{Region => GuacamoleReferenceRegion}` etc. also available for more extreme cases…
  - surely overkill for denormalizing into the class declaration though!

Of course, a lot of this isn't even relevant to `reference.Region` because it (uniquely among the examples I linked above) doesn't share a basename with anything else we use in the project. Anyway, that was my thinking here!

> But it's not really regions we are mapping but thing have coordinates or reference fields.

Yea, I've struggled at points with our abstracting away of reads as "regions" but have eventually made peace with it; better names are always welcome but this seems like the best we've got so far.

PR #464, opened 2016-04-30
rename Main, move things from root pkg, Common

ryan-williams:
commits itemize the changes; mostly moving some longer-tail things out of kitchen-sinks (the root package, and `Common`)

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/464)


arahuja:
I think shortening the NA12878 and CancerWGS test class names to drop utils is more confusing, attaching the suffix would probably be clearer. But it probably doesn't matter much.

ryan-williams:
cool, added the `TestUtil` suffixes back, thanks for all the quick reviews!

arahuja:
👍

PR #467, opened 2016-05-02
Move shared assembly utilities to AssemblyUtils

arahuja:
This PR is mostly moving code and test files.

It
- creates `AssemblyUtils` to move a few functions useful for `somatic-assembly` as well
- adds a 180 base window around the variants in `NA12878.subset.vcf` to alter the NA12878 BAM file.
- moves all the BAM files to git-lfs (Not sure what impact this will have on the PR) might move this to another PR

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/467)


arahuja:
> moves all the BAM files to git-lfs (Not sure what impact this will have on the PR) might move this to another PR

I ended moving this last commit here as it seems like it causes some Travis issues. I'll try it in another PR

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -123,13 +125,13 @@ class DeBruijnGraph(val kmerSize: Int,
   }
 
   /**
-   *  Searches forward or backward from a node to find those connected by a unique path
+    *  Searches forward or backward from a node to find those connected by a unique path
```
ryan-williams: I ran some scripts recently and like 90% of the codebase aligns these the other way, and I think that's more standard in the world at large, so I'd vote we do that! I was going to send a quick PR doing it everywhere but maybe undo this for now?

Make sure you just check the box in intellij and then you'll never have to think about it again.

src/main/scala/org/hammerlab/guacamole/assembly/AssemblyUtils.scala
```diff
@@ -0,0 +1,168 @@
...
+import org.hammerlab.guacamole.reads.MappedRead
+import org.hammerlab.guacamole.reference.{ContigSequence, ReferenceGenome}
+import org.hammerlab.guacamole.util.CigarUtils
+import org.hammerlab.guacamole.variants.ReferenceVariant
+import org.hammerlab.guacamole.windowing.SlidingWindow
+
+import scala.collection.JavaConversions._
+
+
+object AssemblyUtils extends Logging {
```
ryan-williams: I would name this `….assembly.Utils` instead of `….assembly.AssemblyUtils`, per discussion on #463, but if you remain unconvinced then this wfm 😄

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
arahuja: Hah I thought we were switching this the _other_ way (_this way_) so went through switched them as I saw them (switched the intellij setting) but I can switch it back

ryan-williams: sorry, I think I recently did the change this direction in at least one place, so that's probably my fault. happy to help with the munging if you like

arahuja: Should be switched back. Looks like I had some weird re-basing here where half were off line

ryan-williams: [ugh](http://docs.scala-lang.org/style/scaladoc.html)

> Note, especially for those coming from Java, that the left-hand margin of asterisks falls under the _third_ column, not the second, as is customary in Java.

arahuja: What is the ugh for? You don't want to do that or you do? The intellij option is whether or not to use the scaladoc format and scalariform pushed them that way

ryan-williams: "ugh" because I had just convinced myself that first-asterisk was what we should standardize on, and then that says it should be second asterisk. cf #469, lmk which you want to do.

arahuja: I'll leave it as-is now (which should be first asterisk). I'm indifferent on the issue otherwise. As with the other style things, probably something else (intellij/scalariform) should worry about them rather than us, so I'll set my settings appropriately after #469

ryan-williams: Definitely a job for intellij, just need to decide what to tell it we want.

In other news, github seems completely confused/wrong about what the diff is here; I see that your underlying code has everything aligned to first asterisk and yet github showing me





wat.

anyway, this is all resolved here, thx

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -301,26 +303,35 @@ class DeBruijnGraph(val kmerSize: Int,
...
+    kmerCounts.keys.filter(parents(_).isEmpty).map(_.take(kmerSize))
   }
 
   /**
-   * Find all children of a given node
+   * Find all nodes that have out-degree = 0
    *
-   * @param node Kmer to find parents of
-   * @return  List of Kmers where their prefix matches this node's suffix
+   * @return
```
ryan-williams: kill this line and the one above it, or complete it?

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -354,17 +367,103 @@ object DeBruijnGraph {
...
     val head = sequences.headOption.getOrElse(Seq.empty)
     val rest = sequences.tail.flatMap(sequence => sequence.takeRight(sequence.length - overlapSize + 1))
     head ++ rest
   }
 
+  /**
+   * Extract a subsequence from a read sequence
+   *
+   * NOTE: This assumes that the read entirely covers the reference region and does not check for insertions
+   * or deletion
```
ryan-williams: `s/deletion/deletions/`.

Can you expound in the comment on why it's safe to assume no insertions or deletions here?

This feels like it should be named something like `sliceRead` or even made a method on `MappedRead` called `slice`. `getSequenceFromRead` sounds like it would just return `read.sequence`.

ryan-williams: incorporating "subsequence" into the name would also wfm

arahuja: Sure I can rename it `getSubSequence` or something like that. It's very specific to this use case those as noted in the comments.  It doesn't properly check where the read aligns to extract the overlapping sequence.  But what _should_ be the overlapping sequence if the read did not have any insertions or deletions. If the read does have insertions or deletions, this will basically return some random sequence.  

By the way, this was just moved over from `GermlineAssemblyCaller` before.

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -354,17 +367,103 @@ object DeBruijnGraph {
...
+
+    // Filter to reads that entirely cover the region
+    // Exclude reads that have any non-M Cigars (these don't have a 1 to 1 base mapping to the region)
+    val overlapping = reads
+      .filter(!_.cigarElements.exists(_.getOperator != CigarOperator.M))
+      .filter(r => r.overlapsLocus(startLocus) && r.overlapsLocus(endLocus - 1))
+
+    // Extract the sequences from each region
+    val sequences = overlapping
+      .map(r => getSequenceFromRead(r, startLocus, endLocus))
```
ryan-williams: OK, I see that the one use case is just here, and you've filtered out reads with non-`M` cigar-elements just before calling `getSequenceFromRead`. I would:

- inline `getSequenceFromRead`; having a glorified `MappedRead.slice` that assumes as a precondition that reads with non-`M` cigars have been filtered out seems not worth the boilerplate of making it its own method.
- `s/.filter(!…/.filterNot(…` above
- total style nit that you're free to disregard, but I prefer starting a chain of `filter`/`map`s like you have here/above with the initial object on its own line as well, so I'd rewrite this and the block above as:

```scala
    // Filter to reads that entirely cover the region.
    // Exclude reads that have any non-M Cigars (these don't have a 1-to-1 base mapping to the region).
    val sequences = 
      reads
        .filterNot(_.cigarElements.exists(_.getOperator != CigarOperator.M))
        .filter(r => r.overlapsLocus(startLocus) && r.overlapsLocus(endLocus - 1))
        .map(r => r.sequence.slice(startLocus - r.unclippedStart.toInt, endLocus - r.unclippedStart.toInt))
```

or

```scala
    // Filter to reads that entirely cover the region.
    // Exclude reads that have any non-M Cigars (these don't have a 1-to-1 base mapping to the region).
    val sequences =
      for {
        read <- reads
        if !read.cigarElements.exists(_.getOperator != CigarOperator.M)
        if read.overlapsLocus(startLocus) && read.overlapsLocus(endLocus - 1)
        unclippedStart = read.unclippedStart.toInt
      } yield {
        read.sequence.slice(startLocus - unclippedStart, endLocus - unclippedStart)
      }
```

lmk if that doesn't seem better to you

arahuja: Cool, will do. I think something like `filter` vs `filterNot` is a style thing as well; we should add it to the styleguide in #436 `filterNot(p(_))` is just implemented as `filter(!p(_)`

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -367,4 +381,89 @@ object DeBruijnGraph {
...
+  /**
+   * For a given set of reads identify all kmers that appear in the specified reference region
+   *
+   * @param reads  Set of reads to extract sequence from
+   * @param startLocus Start (inclusive) locus on the reference
+   * @param endLocus End (exclusive) locus on the reference
+   * @param minOccurrence Minimum number of times a subsequence needs to appear to be included
+   * @return List of subsequences overlapping [startLocus, endLocus) that appear at least `minOccurrence` time
+   */
+  private def getConsensusKmer(reads: Seq[MappedRead],
```
ryan-williams: this is unused? what happened to its call-sites? all intentional?

arahuja: Yes it is. It was used in `discoverHaplotypes` earlier. I'm gone back and forth between using just the reference as source and sinks kmers vs generating a set of from the reads using this vs just using the `sources` and `sinks` from the graphs. Each has their advantages and disadvantages in terms of performance (both variant calling and speed) so still experimenting with which will be best.

Right now using the reference kmers seems fine and faster in germline calling but am testing the alternatives in the somatic caller.

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
ryan-williams: OK. `.filter(!…` is fine with me too; I can't even find the IntelliJ inspection that I used to see warn about this, so maybe they removed it.

src/test/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCallerSuite.scala
```diff
@@ -163,6 +163,7 @@ class GermlineAssemblyCallerSuite extends GuacFunSuite with BeforeAndAfterAll {
   test (
     "test assembly caller: illumina platinum tests; homozygous insertion 2") {
     verifyVariantsAtLocus(1302671) (
+      ("chr1",1302554,"C", "T"),
```
ryan-williams: need a space before `"C"`

ryan-williams:
This lgtm, looking forward to seeing the somatic counterpart!

arahuja:
Thanks for the review @ryan-williams, pushed an update with your suggestions

ryan-williams:
lgtmerge pending travis!

ryan-williams:
hm, I just rebased #460 on this and pushed and it is pushing 10s of MBs; this didn't add large files to the .git repo did it? Current git size I'm seeing is 435MB, iirc it was in the 200-300MB range recently.

ryan-williams:
my `.git` on another server that hasn't had this PR (and possibly other things!) pushed to it in ~2d is 185MB

arahuja:
@ryan-williams Yes it did add one large file `src/test/resources/illumina-platinum-na12878/NA12878.10k_variants.plus_chr1_3M-3.1M.bam` which existed before, but I expanded the number of reads in it. It should 99MB from ~35MB before. Not sure if you saw the note above but I tried using git-lfs for them, but ran into issues where travis wasn't properly parsing the files.

PR #469, opened 2016-05-02
Consistently align scaladocs

ryan-williams:
Per discussion on #467, we align to the first asterisk of the opening `/**` of a scaladoc most of the time:

```
$ gfs '.scala' | p "po -hM ' *\/\*\*.*?\n *\*' {}" | cls | tts | subtract | hist
  26 2
 299 1
```

That's 26 blocks aligned to the 2nd asterisk, 299 aligned to the 1st.

On the other hand, apparently [the official scala edict](http://docs.scala-lang.org/style/scaladoc.html) puts them under the 2nd one.

I am fine to make it either way; this PR currently puts them all under the first one. Let's decide on one and use this PR to make it so and then never think about it again.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/469)


arahuja:
👍

Issue #439, opened 2016-04-12
Codify style rules


ryan-williams:
I recently removed Scalariform due to conflicts between it, my personal preferences, and (most importantly) IntelliJ that I wasn't able to resolve.

In #435, #436, and #437, some questions have been raised about what styles we should enforce, disallow, allow existing violations of, and even allow new deviations from.

I'm personally fine with a maximally hands-off posture where:
- we agree on preferred style patterns,
- changes that improve code according to that hierarchy are welcomed,
- existing and new deviations are not required to be fixed.

In all cases, the ability of IntelliJ to format code according to a style automatically is a strong argument in favor of a given style.

To seed discussion, here's the ScalaDoc tab of my IntelliJ Code Style > Scala preferences:

![](http://cl.ly/1N3O1F0h363Q/Screen%20Shot%202016-04-12%20at%2012.14.06%20PM.png)

Let me know if you have something different and feel we should standardize on that.

ryan-williams:
I bias towards complying with IntelliJ's style warnings, but one that I'd not really thought about is `.size` vs. `.length` on `String`s and `Array`s; the "more info" makes a reasonably compelling case, though:

![](http://f.cl.ly/items/1H2z0e3q2t273H472K2i/Screen%20Shot%202016-04-29%20at%2010.48.59%20AM.png)

arahuja:
For those that might be interested (likely no one else) `scalariform` does not need to run through maven but can also run as a command line tool: https://github.com/scala-ide/scalariform/

Some properties of use that seem to come up may be:
`alignSingleLineCaseStatements` to `true`
`placeScaladocAsterisksBeneathSecondAsterisk ` to `false` (depending on result of #469 )
`newlineAtEndOfFile` to `true`
`danglingCloseParenthesis` to `Force`
`spaceInsideBrackets` to `false`
`spacesAroundMultiImports` to `false` (I think this was the original property most did not like from scalariform)

ryan-williams:
Per [discussion on #475](https://github.com/hammerlab/guacamole/pull/475/files/3d66c4feb0d1c8d726215af426cec67aa9b60c77#r62576952), I prefer avoiding wildcard imports, the most common exception being when importing `implicit`s from an `object`.

IntelliJ makes this easy to automate by setting a high number here:

[...]

PR #475, opened 2016-05-03
Prune kmer graphs based on read quality scores

arahuja:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/475)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
+++ b/src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
@@ -1,7 +1,9 @@
 package org.hammerlab.guacamole.assembly
 
+import breeze.linalg.DenseVector
+import breeze.stats.mean
 import htsjdk.samtools.CigarOperator
-import org.hammerlab.guacamole.reads.MappedRead
+import org.hammerlab.guacamole.reads.{MappedRead, Read}
 import org.hammerlab.guacamole.util.Bases
 
 import scala.collection.mutable
@@ -338,22 +340,31 @@ object DeBruijnGraph {
   type Sequence = DeBruijnGraph#Sequence
   type Kmer = DeBruijnGraph#Kmer
 
-  def apply(sequences: Seq[Sequence],
+  def apply(reads: Seq[Read],
             kmerSize: Int,
             minOccurrence: Int = 1,
+            minMeanKmerBaseQuality: Int = 0,
             mergeNodes: Boolean = false): DeBruijnGraph = {
 
     val kmerCounts = mutable.Map.empty[DeBruijnGraph#Kmer, Int]
 
-    sequences
-      .filter(Bases.allStandardBases(_))
-      .foreach(
-        _.sliding(kmerSize)
-          .foreach(seq => {
-            val count = kmerCounts.getOrElse(seq, 0)
-            kmerCounts.update(seq, count + 1)
-          })
-      )
+    reads.foreach(
+      read => {
+        read.sequence
+          .sliding(kmerSize)
+          .zipWithIndex
+          .foreach(
+            { case (seq, index) =>
+              val baseQualities = DenseVector(read.baseQualities.slice(index, index + kmerSize).map(_.toFloat).toArray)
+              val meanBaseQuality = mean(baseQualities)
+              if (meanBaseQuality > minMeanKmerBaseQuality) {
+                val count = kmerCounts.getOrElse(seq, 0)
+                kmerCounts.update(seq, count + 1)
+              }
+            }
+          )
+      }
+    )
 
     val graph = new DeBruijnGraph(kmerSize, kmerCounts.filter(_._2 >= minOccurrence))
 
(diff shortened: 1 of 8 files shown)


arahuja:
Rebased this after the lastest commits

arahuja:
Pushed a small update as well - if anyone can review!

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
```diff
@@ -338,22 +339,31 @@ object DeBruijnGraph {
...
-      )
+    reads.foreach(
+      read => {
+        read.sequence
+          .sliding(kmerSize)
+          .zipWithIndex
+          .foreach(
+            { case (seq, index) =>
+              val baseQualities = read.baseQualities.slice(index, index + kmerSize).map(_.toFloat)
+              val meanBaseQuality = mean(baseQualities)
```
ryan-williams: these qualities are log-scaled right? does it make sense to arithmetically average them?

ryan-williams: follow-up question: is it worth optimizing this to compute/maintain a "streaming" mean, so that you aren't averaging (k-1) of the same qualities at each position that you just did at the previous position?

src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala
```diff
@@ -15,15 +17,15 @@ class DeBruijnGraphSuite extends GuacFunSuite {
 
   test("build graph") {
 
-    val sequence = Bases.stringToBases("TCATCTCAAAAGAGATCGA")
+    val sequence = makeRead(Bases.stringToBases("TCATCTCAAAAGAGATCGA"))
```
ryan-williams: nit: change these names from `sequence` to `read`?

src/test/scala/org/hammerlab/guacamole/util/TestUtil.scala
```diff
@@ -27,7 +27,7 @@ import org.apache.spark.SparkContext
 import org.hammerlab.guacamole.ReadSet
 import org.hammerlab.guacamole.loci.set.LociParser
 import org.hammerlab.guacamole.pileup.Pileup
-import org.hammerlab.guacamole.reads.{InputFilters, MappedRead, MateAlignmentProperties, PairedRead, Read, ReadLoadingConfig}
+import org.hammerlab.guacamole.reads._
```
ryan-williams: nit: I prefer avoiding wildcard imports except when importing `implicit def`s from `object`s. You can prevent IntelliJ from using wildcard imports here:



will add to #439 if you agree

src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala
```diff
@@ -38,15 +40,67 @@ class DeBruijnGraphSuite extends GuacFunSuite {
...
     val graph = DeBruijnGraph(Seq(sequence), kmerSize = 3)
 
     graph.kmerCounts === Map("TCA" -> 1, "CAT" -> 2, "AAA" -> 3)
   }
 
+  test("build graph and prune sequence based on kmer quality") {
+
+    val highQualitySequence = makeRead(
+      "TCATCTTAAAAGACATAAA",
+      qualityScores = Some((0 until 19).map(_ => 30)) // All base qualities are 30
```
ryan-williams: nit: I like 2 spaces before inline line-comments; again, lmk if you agree and I'll add to #439

ryan-williams:
looks good overall; one question and a few style comment/questions

src/main/scala/org/hammerlab/guacamole/assembly/DeBruijnGraph.scala
arahuja: > these qualities are log-scaled right? does it make sense to arithmetically average them?

Yes and no. Yes, as in it's commonly done for filtering and if most of the bases are 30 and some are 0 this will average ~low 20s and we can filter on that. No, in that there may be better ways to filter. Most of the times it's exactly those cases to filter >3 bases with 'low' base quality.

> follow-up question: is it worth optimizing this to compute/maintain a "streaming" mean, so that you aren't averaging (k-1) of the same qualities at each position that you just did at the previous position?

Yea, I toyed with this - there wasn't a noticeable performance difference on the tests and small runs, but I could imagine on it's own it's much faster. My implementation was a bit messy so I stuck with this, but can invest time in optimizing if this feature stays.

src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala
arahuja: Sure, will do

src/test/scala/org/hammerlab/guacamole/util/TestUtil.scala
arahuja: Sure, as always I just rely on automated tools, will add the config

src/test/scala/org/hammerlab/guacamole/assembly/DeBrujinGraphSuite.scala
arahuja: Not sure it matters in any way? Would probably disagree then, since it's not easy to automate away and not worth watching for it on PRs?

ryan-williams: OK, can punt until we get an automated story back.

ryan-williams:
lgtm!

arahuja:
Updated based on the feedback! thanks @ryan-williams!

PR #477, opened 2016-05-13
Move test files to git-lfs

arahuja:
This moves the test BAM files to git-lfs. It does require some changes to travis as well to download the files with `git lfs pull`. However, travis does not have `git lfs` installed so we need to use `sudo` to install it. Travis did attempt to add `git lfs` to its container environment but rolled it back, so hopefully it is coming soon.

Also, this does not remove the BAMs from the repo history. We can do that if we'd like, there are instructions here using BFG: https://github.com/github/git-lfs/wiki/Tutorial#using-bfg-to-migrate-repo

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/477)


timodonnell:
Thanks @arahuja ! Maybe update the README as part of this PR to describe how to check out the code with the lfs files and run the tests?

Just fwiw when I check out this branch and run the tests, without doing anything to make git lfs work, I get many errors like below. I assume these are due to my not having configured git lfs. Possibly we should note in the README that if you see errors like this you need to setup lfs?

```
- read evidence for simple snvs *** FAILED ***
  java.lang.RuntimeException: Cannot use index file with textual SAM file
  at htsjdk.samtools.SamReaderFactory$SamReaderFactoryImpl.open(SamReaderFactory.java:275)
  at htsjdk.samtools.SamReaderFactory$SamReaderFactoryImpl.open(SamReaderFactory.java:133)
  at org.hammerlab.guacamole.readsets.ReadSets$.loadFromBAM(ReadSets.scala:190)
  at org.hammerlab.guacamole.readsets.ReadSets$.load(ReadSets.scala:156)
  at org.hammerlab.guacamole.readsets.ReadSets$$anonfun$4.apply(ReadSets.scala:112)
  at org.hammerlab.guacamole.readsets.ReadSets$$anonfun$4.apply(ReadSets.scala:110)

arahuja:
>  Maybe update the README as part of this PR to describe how to check out the code with the lfs files and run the tests?

Sure, will add

>  I assume these are due to my not having configured git lfs

Yes, installing `git lfs` and pulling using `git lfs pull` are needed to download the file as added in travis. I will add that to the README as well.

PR #478, opened 2016-05-16
Moving shared assembly args to AssemblyUtils

arahuja:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/478)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/assembly/AssemblyUtils.scala
+++ b/src/main/scala/org/hammerlab/guacamole/assembly/AssemblyUtils.scala
@@ -2,39 +2,17 @@ package org.hammerlab.guacamole.assembly
 
 import htsjdk.samtools.CigarOperator
 import org.apache.spark.Logging
-import org.bdgenomics.utils.cli.Args4jBase
 import org.hammerlab.guacamole.alignment.{AffineGapPenaltyAlignment, ReadAlignment}
 import org.hammerlab.guacamole.reads.MappedRead
 import org.hammerlab.guacamole.reference.{ContigSequence, ReferenceGenome}
 import org.hammerlab.guacamole.util.CigarUtils
 import org.hammerlab.guacamole.variants.ReferenceVariant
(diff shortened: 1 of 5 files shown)


arahuja:
Going to merge this in, unless there are any comments?

src/main/scala/org/hammerlab/guacamole/assembly/AssemblyUtils.scala
```diff
@@ -2,18 +2,40 @@ package org.hammerlab.guacamole.assembly
...
 import org.hammerlab.guacamole.variants.ReferenceVariant
 import org.hammerlab.guacamole.windowing.SlidingWindow
+import org.kohsuke.args4j.{Option => Args4jOption}
 
 import scala.collection.JavaConversions._
 
 
 object AssemblyUtils extends Logging {
 
+  trait Arguments extends Args4jBase {
```
ryan-williams: I'm not crazy about nesting classes like this inside of objects; would prefer making this `assembly.Arguments` rather than `assembly.AssemblyUtils.Arguments` but we've disagreed about that a few times recently; middle ground might be e.g. `assembly.AssemblyArgs` but this wfm if you prefer it

ryan-williams:
Sry, missed this, one comment but lgtm either way

src/main/scala/org/hammerlab/guacamole/assembly/AssemblyUtils.scala
arahuja: Yea I'm fine doing that - I actually wasn't sure which way we were going, but followed `LociPartitionUtils` which has the `Arguments` trait nested. But I'll move this to `AsssemblyArgs`

ryan-williams: Good point! I think those just ended up there after coming out of being nested in Common but my thinking on this has likely been fluid recently so take with salt 😂

PR #480, opened 2016-05-19
Add slice method to MappedRead

arahuja:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/480)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/reads/MappedRead.scala
+++ b/src/main/scala/org/hammerlab/guacamole/reads/MappedRead.scala
@@ -18,7 +18,7 @@
 
 package org.hammerlab.guacamole.reads
 
-import htsjdk.samtools.Cigar
+import htsjdk.samtools.{Cigar, CigarElement}
 import org.bdgenomics.adam.util.PhredUtils
 import org.hammerlab.guacamole.pileup.PileupElement
 import org.hammerlab.guacamole.reference.{ContigSequence, ReferenceRegion}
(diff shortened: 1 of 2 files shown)


src/main/scala/org/hammerlab/guacamole/reads/MappedRead.scala
```diff
@@ -102,6 +102,63 @@ case class MappedRead(
...
+   * Returns None if the provided coordinates do not overlap the read
+   *
+   * @param from reference locus to start the new read
+   * @param until reference locus to slice up to (exclusive end)
+   * @param referenceContigSequence reference sequence for the read's contig
+   * @return A read which spans [from, until) or None if the provided coordinates do not overlap the read
+   */
+  def slice(from: Long, until: Long, referenceContigSequence: ContigSequence): Option[MappedRead] = {
+    if (from >= end || until < start) {
+      None
```
ryan-williams: I feel like more common semantics would be to return an "empty" (0-length) read in cases like this. Don't feel strongly though.

src/main/scala/org/hammerlab/guacamole/reads/MappedRead.scala
```diff
@@ -102,6 +102,63 @@ case class MappedRead(
...
+    } else if (from <= start && until >= end) {
+      Some(this)
+    } else {
+
+      val referenceStart = math.max(from, start)
+      val referenceEnd = math.min(until - 1, end - 1)
+
+      val el = PileupElement(this, start, referenceContigSequence)
+      val startElement = el.advanceToLocus(referenceStart)
```
ryan-williams: I don't love having to instantiate a PileupElement to get this functionality; the dependency graph of the project should generally not have Reads depending on Pileup things, I think. Might make sense to pull this functionality out or put it into MappedRead and let PileupElement get it from there?

ryan-williams:
lgtm, couple suggestions

src/main/scala/org/hammerlab/guacamole/reads/MappedRead.scala
arahuja: Yea I could see this and was randomly thinking about it this morning before seeing this. It seems weird, but PileupElement has most of the logic to advance the sequence on a read. Really that piece should be factored out a bit so we can view a read base at any reference position probably.

Otherwise I think just moving this to a third place that depends on both works, but I don't think that really adds anything either

src/main/scala/org/hammerlab/guacamole/reads/MappedRead.scala
arahuja: Yea, thought for a while on that. I didn't think a 0-length read had any meaning really though so returning None would be cleaner? On a standard sequence it's easy to check for an empty sequence as well (essentially acting as null/None) but here the read the would still exist but its main info (the sequence) would be null.

PR #482, opened 2016-06-21
pom profiles

ryan-williams:
New Maven profiles:
- default (`-Pguac`): build a guac jar with relocated guava classes shaded.
- `-Pdeps`: build a shaded jar of all deps, but not including guac or relocated guava.
- `-Puber`: build a JAR that is all of the above (current default).

This solves two problems:
- when iteratively building guac, jars have to be repackaged for every trivial change.
  - the uber jar takes >70s locally when not even responding to any compilation changes.
  - the new default guac jar takes ~20s locally.
- the `guacamole-<version>.jar` JAR that Maven will pick up by anything depending on `org.hammerlab.guacamole:guacamole:<version>` will now have shaded+relocated guava in it; the previous lack of it made the JAR basically unusable via this route, afaict.

Suggested workflow:
```
alias mvp="mvn package -DskipTests"
mvp -Pdeps  # once, or only when dependencies change
mvp         # while developing
version=0.0.1-SNAPSHOT
main=org.hammerlab.guacamole.Main
guac_jar=target/guacamole-$version.jar
guac_deps_jar=target/guacamole-deps-only-$version.jar
spark-submit <spark confs> --class $main $guac_jar --jars $guac_deps_jar <guac args>
```

To build both the "guac" and "deps" JARs in one cmd:

```
mvp -Pguac,deps
```

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/482)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,18 +27,41 @@ inspiration from the [Avocado][] project.
 
 For hacking Guacamole, see our [code docs][].
 
-# Running Guacamole on a Single Node
+## Building
 
-Guacamole requires [Apache Maven][maven] (version 3.0.4 or higher).
+Building Guacamole requires [Apache Maven][maven] (version 3.0.4 or higher).
 
-Build:
+You'll generally want two JARs to be on your classpath when you run Guacamole:
+
+- the main Guacamole JAR, and
+- a shaded JAR with all of Guacamole's transitive dependencies.
+
+The former can be built in any of the following ways:
 
 ```
 mvn package
+mvn package -DskipTests  # don't run tests, just package the JAR.
+mvn package -Pguac       # this is the default Maven profile, but can be useful for combining with other profiles as examples below demonstrate.
 ```
 
-This will build a guacamole JAR file in the `target` directory. You can use the
-`guacamole` script to run locally:
+The latter can be built using the `deps` profile:
+
+```
+mvn package -Pdeps
+mvn package -Pdeps -DskipTests
+```
+
+You can also build both at once:
+
+```
+mvn package -Pguac,deps
+mvn package -Pguac,deps -DskipTests
+```
+
+## Running
+
+### Running Locally
+`scripts/guacamole` makes it easy to run locally:
 
 ```
 scripts/guacamole somatic-joint \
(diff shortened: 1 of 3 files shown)


timodonnell:
I don't have a great understanding of these issues, but at first glance, why change the default? Would it work to just leave the default as is (I guess `-Puber`) and then also add a faster thing (`-Pguac` and `-Pdeps`) that people who know what they're doing can use for faster compilations? it seems worthwhile to me that a simple `mvn package` builds the full package, including dependencies.

ryan-williams:
> it seems worthwhile to me that a simple mvn package builds the full package, including dependencies.

"The full package" of Guacamole is the guac classes (and select shaded guava classes) and the POM specifying guac's dependencies. That's what we ship to Maven*, what downstream tools and deployment methods (including the shade-plugin which builds our uber-jar) consume, and what `mvn package` will produce by default as of this PR.

Various methods exist in the JVM ecosystem for gathering all transdeps of a library for deployment, and "building a 200MB uber jar" is one of the worst such methods.

As of this PR, that method remains available via a Maven profile, alongside the "one 3MB guac jar and one 197MB deps jar" method, which improves upon the former by decoupling the building of guac from the bundling of its transdeps, allowing for much faster build times in the common case that no dependencies have changed. It also separates the two kinds of shading we do: that of certain guava classes (which is crucial for guacamole to work properly in Hadoop environments), and the shading of all transdeps (which is one of several possible ways to deploy guacamole, and which we inherited while copying ADAM's configurations, afaict).

Another method for deploying Guacamole can be seen in spark-notebook, which parses POMs and builds a classpath that points to all trans-dep JARs in the local Maven cache, so it needs to consume just the "thin" guac jar. Spark-notebook itself is not an especially crucial use case, but it illustrates a common and scalable route by which Guacamole {can,will,should} be {deployed,integrated} with other tools.

Finally, [the Maven dependency plugin](http://maven.apache.org/plugins/maven-dependency-plugin/build-classpath-mojo.html) allows for printing out (or saving to a file) a classpath pointing at existing local copies of trans-dep JARs; this is a capability I'd like to add to Guacamole as well. Presumably it's what Spark-notebook does under the hood, but I've not checked.

So there are a bunch of ways to put Guacamole classes on a classpath with its dependencies, and the one-uber-jar method has large drawbacks, is unsuitable for some applications (which leverage Maven to resolve deps themselves, which arguably makes more sense), and should not be thought of as a default in any meaningful sense, so I kept it available but made it non-default.

*It's worth noting that the Guac artifact currently on Maven central does not include the shaded Guava classes and so will not run correctly in a Hadoop environment, IIUC.

timodonnell:
Cool, thanks for the explanation @ryan-williams . Since we seem to be going from a situation where we need to keep track of one jar to one where where we have to have two (a guacamole jar and a deps jar) for any downstream tool that wants to run guacamole, I just want to make sure that this increase in complexity is the standard thing to do. Sounds like our old way was broken though and this is correct way to do it from what you're saying, so LGTM.

ryan-williams:
> Since we seem to be going from a situation where we need to keep track of one jar to one where where we have to have two…

That's not really accurate. This PR makes the option to build/deploy Guacamole as one JAR equally accessible as the new option to deploy it as two: they each are enabled via a profile, befitting their status as specific deployment strategies that encode assumptions about build/deployment environment beyond the minimal packaging of [the library and info about its dependencies] that is published to Maven and generally used by any downstream dependents.

The third notable strategy is to use the Maven dependency plugin to build a classpath that points to all transdeps' JAR paths in the local Maven cache; it also has tradeoffs relative to the other two and is worth adding. I just filed #494 to track this.

I think they're all worth supporting, but the right default `mvn package` behavior is to build just the guac classes, which is the minimal deliverable of this repo, free from additional assumptions about deployment methodology.

arahuja:
I'd be in favor of this I think. I very consistently use `-Pcluster` currently to avoid packaging up Hadoop and Spark JARs (and might even want those split out as well). Those JARs are usually already available and it helps avoid any version collisions with the cluster JARs.

ryan-williams:
Interesting @arahuja, that makes sense, but I also didn't consciously change the spark/hadoop provided-ness here; should I fold spark/hadoop `provided` into the `deps` profile here?

arahuja:
@ryan-williams depends? For example, I think for iterating on the cluster the best scenario is having all non-hadoop deps built once available somewhere, a guac jar that you can change and take all hadoop/spark things from the cluster. 

But locally you want 3 - non-hadoop deps, hadoop+spark, guac. 

If you can have profiles override (you can right?) I'd roll them together where `-Pdeps` does what you have and `-Pdeps,cluster` builds a deps jar with `provided` on the others.

ryan-williams:
Cool, makes sense, I'll merge this in as-is then.

PR #484, opened 2016-07-10
fix somatic joint integration test arg-parsing

### LINKED ISSUE #483: SomaticJointCallerIntegrationTests are broken
The SomaticJointCallerIntegrationTests (which have been refactored to be a standalone entry point instead of a test since since the last time I used them), give these instructions for how to run them:

```
 * To run:
 *
 *   mvn package
 *   mvn test-compile
 *   java \
 *     -cp target/guacamole-with-dependencies-0.0.1-SNAPSHOT.jar:target/scala-2.10.5/test-classes \
 *     org.hammerlab.guacamole.main.SomaticJointCallerIntegrationTests
```

However, when I do that, I get this usage error followed by the joint caller help message:

```
[tim@Tims-MacBook ~/sinai/git/guacamole]$  java \
>       -cp target/guacamole-with-dependencies-0.0.1-SNAPSHOT.jar:target/scala-2.10.5/test-classes \
>       org.hammerlab.guacamole.main.SomaticJointCallerIntegrationTests
Option "--reference-fasta" is required
```

Also, it's unclear to me how to run these from IntelliJ.

I'm thinking we should probably roll back the change that made these non-tests. Even though they don't assert anything, by the fact that they run successfully they effectively test a number of things. Having them as tests also makes sure we don't end up in the current situation where they bit rot.

ryan-williams:
fixes #483

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/484)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/SparkCommand.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/SparkCommand.scala
@@ -3,6 +3,8 @@ package org.hammerlab.guacamole.commands
 import org.apache.spark.{SparkConf, SparkContext}
 import org.bdgenomics.utils.cli.Args4jBase
 
+import scala.collection.mutable
+
 abstract class SparkCommand[T <% Args4jBase: Manifest] extends Command[T] {
   override def run(args: T): Unit = {
     val sc = createSparkContext(appName = name)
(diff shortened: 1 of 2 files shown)


timodonnell:
Cool, thanks @ryan-williams ! LGTM

PR #489, opened 2016-07-11
use type aliases for Locus, ContigName, etc.

ryan-williams:
also rename some vars to be more consistent across classes

This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/489)


timodonnell:
LGTM

PR #490, opened 2016-07-11
rename some instance vars to be more consistent across classes

ryan-williams:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/490)


ryan-williams:
Actually I'm going to fold this into #489

PR #491, opened 2016-07-11
ensure contig ordering in LociSet/LociMap

ryan-williams:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/491)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/loci/map/LociMap.scala
+++ b/src/main/scala/org/hammerlab/guacamole/loci/map/LociMap.scala
@@ -19,7 +19,7 @@
 package org.hammerlab.guacamole.loci.map
 
 import org.hammerlab.guacamole.loci.set.{LociSet, Builder => LociSetBuilder}
-import org.hammerlab.guacamole.reference.{ContigName => ContigName}
+import org.hammerlab.guacamole.reference.ContigName
 import org.hammerlab.guacamole.strings.TruncatedToString
 
 import scala.collection.immutable.TreeMap
@@ -90,6 +90,6 @@ object LociMap {
     LociMap(
       TreeMap(
         contigs.map(contig => contig.name -> contig).toSeq: _*
-      )
+      )(ContigName.ordering)
     )
 }
--- a/src/main/scala/org/hammerlab/guacamole/loci/set/Builder.scala
+++ b/src/main/scala/org/hammerlab/guacamole/loci/set/Builder.scala
@@ -1,21 +1,19 @@
 package org.hammerlab.guacamole.loci.set
 
-import scala.collection.SortedMap
+import scala.collection.mutable.ArrayBuffer
 
 /**
  * Build a LociSet out of Contigs.
  */
 private[loci] class Builder {
-  private val map = SortedMap.newBuilder[String, Contig]
+  private val map = ArrayBuffer[Contig]()
 
   def add(contig: Contig): this.type = {
     if (!contig.isEmpty) {
-      map += ((contig.name, contig))
+      map += contig
     }
     this
   }
 
-  def result: LociSet = {
-    LociSet(map.result())
-  }
+  def result: LociSet = LociSet.fromContigs(map)
 }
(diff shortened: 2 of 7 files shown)


timodonnell:
Why do we care about the ordering?

ryan-williams:
I've had a hunch for a little while that we secretly/unintentionally depend on the contig order in sorted BAMs matching the (lexicographical) order that the contigs end up in our `Loci{Set,Map}`'s `SortedMap`s, due to [this block](https://github.com/hammerlab/guacamole/blob/a0296f95a5124d6a6a503b3e019840c9c0cd3664/src/main/scala/org/hammerlab/guacamole/distributed/WindowFlatMapUtils.scala#L237-L241):

```scala
taskLoci.contigs.flatMap(contig => {
  val regionIterator: PerSample[Iterator[R]] = regionSplitByContigPerSample.map(_.next(contig.name))
  val windows: PerSample[SlidingWindow[R]] = regionIterator.map(SlidingWindow[R](contig.name, halfWindowSize, _))
  generateFromWindows(contig, windows)
}).iterator
```

I'll try to write a test that exposes this now, stay tuned, or lmk if I'm missing something.

In any case, I figured we may as well keep them sorted in a saner way than lexicographically, makes pretty-printing nicer without an additional sort, :isaacshrug:.

ryan-williams:
Yea, check out [this test](https://github.com/ryan-williams/guacamole/commit/f4552bc3d69286941e619915bfcb298e4f96f6eb) which passes, demonstrating the bug.

```scala
val readChr2  = TestUtil.makeRead("AAAAA", chr =  "chr2", start = 10)
val readChr10 = TestUtil.makeRead("AAAAA", chr = "chr10", start = 10)

val loci = LociSet("chr2:0-1000,chr9:0-1000,chr10:0-1000")

val rbc = new RegionsByContig(Iterator(readChr2, readChr10))

intercept[AssertionError]({
  loci.contigs.map(contig => rbc.next(contig.name).toList).toList
}).getMessage should include("Regions are not sorted by contig")
```

Also:

```scala
val loci = LociSet("chr2:0-1000,chr10:0-1000")

val rbc = new RegionsByContig(Iterator(readChr2, readChr10))

loci.contigs.map(contig => rbc.next(contig.name).toList).toList should be(
  List(
    List(),
    List(readChr2)
  )
)
```

timodonnell:
Sorry I'm quite likely missing something, but I'm not sure I follow. Are we getting incorrect results currently? Or are you saying that lexicographic ordering is just unintuitive?

The first test would seem to me to be correct, because, lexicographically sorted, chr10 does come before chr2.

The nice thing about lexicographic ordering is that we don't have to assume anything about what reference / species we're using and how contigs are named.

ryan-williams:
No, I believe this to be a bug that is live in guac today.

You're right that [the first test case](https://github.com/ryan-williams/guacamole/blob/contig-order-error/src/test/scala/org/hammerlab/guacamole/distributed/RegionsByContigSuite.scala#L12-L25) is fine on its own, but it hints at the problem that the [second](https://github.com/ryan-williams/guacamole/blob/contig-order-error/src/test/scala/org/hammerlab/guacamole/distributed/RegionsByContigSuite.scala#L27-L35) and [third](https://github.com/ryan-williams/guacamole/blob/contig-order-error/src/test/scala/org/hammerlab/guacamole/distributed/RegionsByContigSuite.scala#L37-L48) cases confirm.

I've added a [fourth case](https://github.com/ryan-williams/guacamole/blob/contig-order-error/src/test/scala/org/hammerlab/guacamole/distributed/RegionsByContigSuite.scala#L50-L62) that illustrates the problem directly as it is most likely affecting us today: the partition that spans the boundary between `chr9` and `chr10` will have all its `chr10` reads skipped, effectively leaving us with a varyingly-sized blind-spot at the start of `chr10`.

ryan-williams:
Ah actually now I think you're right and that I was mistaken; we partition the reads by TaskPosition upstream of RegionsByContig, making they'll already be in lex order by contig, so the existing behavior is probably ok, and moreover without changing the ordering on TaskPosition this PR would in fact add the bug I thought I was fixing! Sorry for the false alarm. I guess it's ok that we have partitions that span chr1+chr10, chr19+chr2, and chr2+chr20.

PR #493, opened 2016-07-11
some style nits

ryan-williams:
- import cleanup
- make `val`s `def`s in some `trait`s
- space out some code blocks vertically more for clarity; i think this it the root desire driving a lot of my drive-by nits historically, as i've thought about it more!

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/493)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/distributed/RegionsByContig.scala
+++ b/src/main/scala/org/hammerlab/guacamole/distributed/RegionsByContig.scala
@@ -20,22 +20,31 @@ import org.hammerlab.guacamole.reference.{ContigName, ReferenceRegion}
  * @param regionIterator regions, sorted by contig and start locus.
  */
 class RegionsByContig[R <: ReferenceRegion](regionIterator: Iterator[R]) {
+
   private val buffered = regionIterator.buffered
+
   private var seenContigs = List.empty[ContigName]
-  private var prevIterator: Option[SingleContigRegionIterator[R]] = None
+  private var curIterator: SingleContigRegionIterator[R] = _
+
   def next(contigName: ContigName): Iterator[R] = {
-    // We must first march the previous iterator we returned to the end.
-    while (prevIterator.exists(_.hasNext)) prevIterator.get.next()
+
+    // Drop any remaining regions from the previous contig (which drops them from the underlying regionIterator), as the
+    // new SingleContigRegionIterator assumes it receives regionIterator pointing at the start of a fresh contig.
+    while (curIterator != null && curIterator.hasNext) curIterator.next()
 
     // The next element from the iterator should have a contig we haven't seen so far.
-    assert(buffered.isEmpty || !seenContigs.contains(buffered.head.contigName),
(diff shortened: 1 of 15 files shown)


src/main/scala/org/hammerlab/guacamole/distributed/RegionsByContig.scala
```diff
@@ -20,22 +20,30 @@ import org.hammerlab.guacamole.reference.ReferenceRegion
...
 class RegionsByContig[R <: ReferenceRegion](regionIterator: Iterator[R]) {
+
   private val buffered = regionIterator.buffered
   private var seenContigs = List.empty[String]
-  private var prevIterator: Option[SingleContigRegionIterator[R]] = None
+  private var curIterator: SingleContigRegionIterator[R] = _
+
   def next(contig: String): Iterator[R] = {
-    // We must first march the previous iterator we returned to the end.
-    while (prevIterator.exists(_.hasNext)) prevIterator.get.next()
```
timodonnell: Seems like potentially more than a stylistic change. Is this intended to change semantics?

ryan-williams: no semantic change intended; I lost a bit in my rebasing here which caused the last travis failure. should be fixed now.

I've gone back and forth on using `null`s vs. `None`s in various Iterator implementations; I used to be big on `Option`s for these cases but a lot of `Iterator`s are not written that way and it seems cleaner not to in a lot of cases.

For example, here the `null` is only there for the very first time the Iterator runs; the variable is logically not optional in the steady state, but is more like lazily-initialized. Anyway, just a gut thing, if you don't like it I can revert it. My initial idea here was to call it `curIterator` instead of `prevIterator` which seems clearer.

src/main/scala/org/hammerlab/guacamole/pileup/PileupElement.scala
```diff
@@ -216,7 +216,7 @@ case class PileupElement(
    */
   @tailrec
   final def advanceToLocus(newLocus: Long): PileupElement = {
-    assume(newLocus >= locus, "Can't rewind to locus %d from %d. Pileups only advance.".format(newLocus, locus))
+    assume(newLocus >= locus, s"Can't rewind to locus $newLocus from $locus. Pileups only advance. $read")
```
arahuja: Any reason this is printing this specific read?

arahuja:
👍

arahuja:
👍  assuming the issue brought up by @timodonnell was addressed

src/main/scala/org/hammerlab/guacamole/pileup/PileupElement.scala
ryan-williams: I think I added that just to help track down situations where this error state was hit, while debugging things. NBD either way if you want to remove it.

ryan-williams:
Yea I believe it was, merging.

PR #495, opened 2016-07-11
improve LociMap/Set serde

ryan-williams:
in the course of broadcasting LociMaps and LociSets in a spark repl,
they can be java serialized, which was throwing errors.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/495)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/loci/set/Contig.scala
+++ b/src/main/scala/org/hammerlab/guacamole/loci/set/Contig.scala
@@ -1,5 +1,6 @@
 package org.hammerlab.guacamole.loci.set
 
+import java.io.{ObjectInputStream, ObjectOutputStream}
 import java.lang.{Long => JLong}
 
 import com.google.common.collect.{RangeSet, TreeRangeSet, Range => JRange}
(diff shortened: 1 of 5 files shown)


src/main/scala/org/hammerlab/guacamole/loci/set/Contig.scala
```diff
@@ -14,7 +15,32 @@ import scala.collection.mutable.ArrayBuffer
 /**
  * A set of loci on a contig, stored/manipulated as loci ranges.
  */
-case class Contig(name: ContigName, private val rangeSet: RangeSet[JLong]) extends TruncatedToString {
+case class Contig(var name: ContigName, private var rangeSet: RangeSet[JLong]) extends TruncatedToString {
+
+  private def readObject(in: ObjectInputStream): Unit = {
```
arahuja: Do these need to `override` or implement anything? `kryo.readObject` will just find these?

ryan-williams: yea, weird semantics of Java `Serializable`, you make them `private` and not an `override`.

ryan-williams: to be clear, this is just for Java serde, nothing to do with Kryo.

arahuja: Ah ok, and you don't need extends Serializable?

ryan-williams: Good q, had to double-check, but `case class`es are always `Serializable`:

```scala
scala> case class A(n: Int)
scala> import java.io.ObjectOutputStream
scala> import java.io.ByteArrayOutputStream
scala> val oos = new ObjectOutputStream(new ByteArrayOutputStream())
scala> oos.writeObject(A(4))
scala> class B(n: Int)
scala> oos.writeObject(new B(4))
java.io.NotSerializableException: B
  at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1183)
  at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:347)
  ... 32 elided
```

arahuja:
👍

PR #485, opened 2016-07-11
loci-partitioning + pileup-creation refactoring, and new strategies

ryan-williams:
**See other PRs where I've broken functionality from this one into more manageable bits**.

Several high-level things happening here:

- puts our three loci-partitioning strategies (existing "uniform" and "approximate", as well as new "exact") behind a shared `LociPartitioner` interface.
- adds a `PartitionedRegions` class which wraps a `LociPartitioning` and an `RDD` of reads *after they've been partitioned* (meaning some reads exist more than once in the RDD).
- both of the above allow for persisting to disk, and printing a variety of summary stats.
- new pileup-creation machinery (`PileupsRDD`) keeps clean abstractions between each piece of the pipeline, from `{reads, loci}` to [partitioned reads] to {locus, reads overlapping locus} to {pileups}, each emits an `Iterator` that is consumed by the next step's `Iterator`.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/485)

PR #496, opened 2016-07-25
a bunch of style nits

ryan-williams:
pulled these out while separating #485 into smaller PRs.

some typo fixes, some `import` cleanups, some separating blocks of code out onto more lines in cases where it felt crowded

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/496)


arahuja:
A lot of these make sense to me, but many do not as well. I'm not sure it's worth commenting on all individually, but as for many of these style items ( line lengths, dangling parens, how function args should look) if we are following a specific style guide we should probably be explicit about that somewhere and effort would be better spent just codifying that (see #439). Otherwise, likely, these will just move back in some other PR.

ryan-williams:
Sounds good, I'll see about which of these are worth articulating/discussing more explicitly e.g. on #439.

ryan-williams:
Having thought about this a bit I think we should hash these out here.

I'll go leave some comments on a few cases explaining why I thought it was worth changing.

Also, to be clear, I'm not saying your or others' guac code needs to be formatted similarly, I just unconsciously reformat stuff in a way that is clearer to me while reading code, and so on the cases you agree we should make use of my having already done the work, rather than throwing it away.

The cases you disagree about I think we should actually discuss, and it makes more sense to do it here with real context than over on #439. If consensus emerges on things we can put them over there.

src/main/scala/org/hammerlab/guacamole/assembly/AssemblyArgs.scala
```diff
@@ -1,9 +1,8 @@
 package org.hammerlab.guacamole.assembly
 
-import org.bdgenomics.utils.cli.Args4jBase
 import org.kohsuke.args4j.{Option => Args4jOption}
 
-trait AssemblyArgs extends Args4jBase {
```
ryan-williams: while moving `Args` traits around I decided the right place to mix in `Args4jBase` is at lower levels of the inheritance tree, e.g. in the actual `SparkCommand`s' Arguments that use them.

I've ended up with a few traits that serve as interfaces to "args" traits that implement fields as `@Argument`s, which also helped clarify that we don't need to be specifying every thing up the tree as a concrete "args" object.

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -74,17 +75,19 @@ object GermlineAssemblyCaller {
           .getPartitioner(mappedReads, args.assemblyWindowRange)
           .partition(loci.result(contigLengths))
 
-      val genotypes: RDD[CalledAllele] = discoverGermlineVariants(
```
ryan-williams: A theme that has emerged for me is that I like only having one thing happening on a given line; scala is *so* dense, and lines are so much wider than they are tall (in terms of pixels on screen), that things tend to be clearer the more lines they are split over, at least starting from where we are today.

So here, I like naming a variable on one line, opening the function call on the next, specifying each arg on its own line, and closing the function scope with a `)` line.

Again, not saying you have to do things this way, but if you are at worst neutral on it, I really prefer it. Many more examples to come.

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -118,7 +121,7 @@ object GermlineAssemblyCaller {
           initialState = None,
           (lastCalledLocus, windows) => {
             val window = windows.head
-            val referenceName = window.contigName
```
ryan-williams: starting to standardize around `contigName` for this kind of variable; open to changing it, but a good first step is homogenizing uses.

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -131,49 +134,64 @@ object GermlineAssemblyCaller {
...
-            val referenceContig = reference.getContig(referenceName)
+            val referenceContig = reference.getContig(contigName)
 
             val regionReads = window.currentRegions()
             // Find the reads the overlap the center locus/ current locus
             val currentLocusReads =
               regionReads
                 .filter(_.overlapsLocus(window.currentLocus))
 
-            val pileup = Pileup(currentLocusReads, referenceName, window.currentLocus, referenceContig)
```
ryan-williams: example where more lines is cleaner imho: having to count 3 or 4 args across a line is much harder than scanning down per-arg lines.

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -243,11 +262,11 @@ object GermlineAssemblyCaller {
         val probability = math.exp(mostLikelyGenotypeAndProbability._2)
         genotype
           .getNonReferenceAlleles
-          .toSet // Collapse homozygous genotypes
+          .toSet  // Collapse homozygous genotypes
           .filter(_.altBases.nonEmpty)
           .map(allele => {
             CalledAllele(
-              pileup.head.read.sampleName,
+              pileup.elements.head.read.sampleName,
```
ryan-williams: In many places we use the first element of an array to represent values we assume are the same for all elements in the array; this is generally smelly to me and you get better APIs and code by restructuring things to avoid that, IME. This is a step towards removing `Pileup.head` which is an example of this.

src/main/scala/org/hammerlab/guacamole/commands/SomaticStandardCaller.scala
```diff
@@ -207,7 +212,9 @@ object SomaticStandard {
         Likelihood.likelihoodsOfAllPossibleGenotypesFromPileup(
           filteredNormalPileup,
           Likelihood.probabilityCorrectIgnoringAlignment,
-          normalize = true).toMap
```
ryan-williams: passing an arg, closing a function call, and doing a full collection transformation, all on one line / as an afterthought: too much, imho.

Something like `.toMap` is many lines of code in some languages; scala's concision for the option of writing very little, and having amazing one-liners, but obviously "can" != "should" in all cases.

src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
```diff
@@ -130,6 +128,8 @@ object VAFHistogram {
           .getPartitioner(readsets.allMappedReads)
           .partition(loci.result(contigLengths))
 
+      val reference = ReferenceBroadcast(args.referenceFastaPath, sc)
```
ryan-williams: `val` was declared a lot earlier than it was used

src/main/scala/org/hammerlab/guacamole/commands/VariantSupport.scala
```diff
@@ -110,23 +109,26 @@ object VariantSupport {
...
-      alleles.map(kv => AlleleCount(pileup.sampleName,
-        pileup.contigName,
-        pileup.locus,
-        Bases.basesToString(kv._1.refBases),
-        Bases.basesToString(kv._1.altBases),
-        kv._2.size)).iterator
-    }
+    def pileupToAlleleCounts(pileup: Pileup): Iterator[AlleleCount] =
+      (for {
+        (allele, elements) <- pileup.elements.groupBy(_.allele)
```
ryan-williams: `for`s generally allow writing a lot less cruft, so I prefer them; when you want a destructuring assignment as well, they are even better.

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/AlleleAtLocus.scala
```diff
@@ -104,34 +104,38 @@ object AlleleAtLocus {
       val subsequenceCounts =
         ReadSubsequence.nextAlts(pileup.elements)
           .filter(subsequence => !onlyStandardBases || subsequence.sequenceIsAllStandardBases)
-          .groupBy(x => (x.endLocus, x.sequence))
-          .map(pair => (pair._2.head -> pair._2.length))
+          .groupBy(x => (x.endLocus, x.sequence()))
```
ryan-williams: if a method's declaration has `()` then IntelliJ warns against not calling with `()`; I don't know the significance otomh but it feels reasonable to be consistent with the declaration. In particular, you can't go the other direction.

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/SomaticJointCaller.scala
```diff
@@ -185,35 +186,38 @@ object SomaticJoint {
...
+        onlySomatic,
+        includeFiltered
+      ).toIterator
+    }
+
     pileupFlatMapMultipleRDDs(
       readsets.mappedReads,
       lociPartitions,
       skipEmpty = true,  // TODO: shouldn't skip empty positions if we might force call them. Need an efficient way to handle this.
-      rawPileups => {
```
ryan-williams: factored this code out to a callback so that this `pileupFlatMapMultipleRDDs` call can be a little less gnarly

src/main/scala/org/hammerlab/guacamole/distributed/PileupFlatMapUtils.scala
```diff
@@ -6,15 +6,16 @@ import org.hammerlab.guacamole.loci.partitioning.LociPartitioning
...
-import org.hammerlab.guacamole.reference.{ReferenceGenome, _}
+import org.hammerlab.guacamole.reference.{ContigSequence, ReferenceGenome}
 import org.hammerlab.guacamole.windowing.SlidingWindow
 
 import scala.reflect.ClassTag
 
 object PileupFlatMapUtils {
   /**
    * Helper function. Given optionally an existing Pileup, and a sliding read window return a new Pileup at the given
-   * locus. If an existing Pileup is given as input, then the result will share elements with that Pileup for efficiency.
```
ryan-williams: I thought we were agreed that >120 chars was fair game for breaking onto multiple lines.

src/main/scala/org/hammerlab/guacamole/distributed/PileupFlatMapUtils.scala
```diff
@@ -104,18 +111,29 @@ object PileupFlatMapUtils {
       skipEmpty,
       halfWindowSize = 0,
       initialState = None,
-      function = (maybePileups: Option[PerSample[Pileup]], windows: PerSample[SlidingWindow[MappedRead]]) => {
-        val advancedPileups = maybePileups match {
-          case Some(existingPileups) => {
```
ryan-williams: this block is so crowded that I found it unreadable

src/main/scala/org/hammerlab/guacamole/distributed/WindowFlatMapUtils.scala
```diff
@@ -14,34 +14,36 @@ import scala.reflect.ClassTag
 object WindowFlatMapUtils {
 
   /**
-    * FlatMap across loci, and any number of RDDs of regions, where at each locus the provided function is passed a
```
ryan-williams: Aligning to the first asterisk like (almost?) everything else in this repo.

src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -243,11 +262,11 @@ object GermlineAssemblyCaller {
         val probability = math.exp(mostLikelyGenotypeAndProbability._2)
         genotype
           .getNonReferenceAlleles
-          .toSet // Collapse homozygous genotypes
+          .toSet  // Collapse homozygous genotypes
```
ryan-williams: I prefer 2 spaces before line comment

src/test/scala/org/hammerlab/guacamole/commands/SomaticStandardCallerSuite.scala
```diff
@@ -204,15 +210,15 @@ class SomaticStandardCallerSuite extends GuacFunSuite with TableDrivenPropertyCh
       TestUtil.makeRead("TCGATCGA", "8M", 0)
     )
 
-    val normalPileup = Pileup(normalReads, "chr1", 2, referenceContigSequence = simpleReference.getContig("chr1"))
+    val normalPileup = makePileup(normalReads)
```
ryan-williams: I really like reducing repetitive boilerplate in tests, so that they are close to [blobs of data in] -> [expected blobs of data out].

src/test/scala/org/hammerlab/guacamole/commands/jointcaller/ReadSubsequenceSuite.scala
```diff
@@ -23,12 +23,12 @@ class ReadSubsequenceSuite extends GuacFunSuite {
...
       TestUtil.makeRead("TNGAGCGA", "8M", 1) // contains N base
     )
-    val pileups = reads.map(read => Pileup(Seq(read), "chr1", 1, simpleReference.getContig("chr1")))
+    val pileups = reads.map(read => makePileup(Seq(read), "chr1", 1))
 
-    ReadSubsequence.ofFixedReferenceLength(pileups(0).head, 1).get.sequence should equal("C")
-    ReadSubsequence.ofFixedReferenceLength(pileups(0).head, 2).get.sequence should equal("CG")
-    ReadSubsequence.ofFixedReferenceLength(pileups(1).head, 6).get.sequence should equal("CGACCCTCG")
-    ReadSubsequence.ofFixedReferenceLength(pileups(3).head, 1).get.sequenceIsAllStandardBases should equal(false)
+    ReadSubsequence.ofFixedReferenceLength(pileups(0).elements.head, 1).get.sequence should equal("C")
```
ryan-williams: after a change above in the assembly caller, this was the last use of `Pileup.head`, the existence of which is a smell, so I unrolled these accesses.

src/main/scala/org/hammerlab/guacamole/reads/Read.scala
```diff
@@ -84,57 +85,66 @@ object Read extends Logging {
...
-        record.getReferenceName.intern,
-        record.getMappingQuality,
-        record.getAlignmentStart - 1,
-        cigar = record.getCigar,
-        failedVendorQualityChecks = record.getReadFailsVendorQualityCheckFlag,
-        isPositiveStrand = !record.getReadNegativeStrandFlag,
-        isPaired = record.getReadPairedFlag
-      )
+    val sampleName =
+      (if (record.getReadGroup != null && record.getReadGroup.getSample != null)
```
ryan-williams: I like omitting curly-braces for expressions like this; we shouldn't accumulate any / a lot of state in inline blocks, so the default should be single expressions, and omitting curly-braces reduces clutter significantly imho.

ryan-williams:
OK, on third thought I am breaking this out into a bunch of other PRs!

PR #497, opened 2016-07-25
break readsets into sub-pkgs

ryan-williams:
- some argument classes to `readsets.args`
- some read-loading functionality to `readsets.loading`
- `ReadsRDD` to `readsets.rdd`

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/497)


arahuja:
LGTM code wise, two comments, but up to you:
- `loading`: would `io` be a more standard name?
- `SingleSampleArgs`: this somewhat drops the idea that this is for loading reads (BAMs or ADAM)? as opposed to variants, or fastq input

ryan-williams:
> would `io` be a more standard name?

good idea, done!

> `SingleSampleArgs`: this somewhat drops the idea that this is for loading reads (BAMs or ADAM)? as opposed to variants, or fastq input

good q; imho `SingleSampleArgs` residing in the `readsets` package makes its scoped-ness to "reads" data sufficiently clear.

PR #498, opened 2016-07-25
add LociPartitioner interface

ryan-williams:
currently two implementations, `{Approximate,Uniform}Partitioner`, corresponding to `partitionLoci{Uniform,Approximate}ly` methods.

A third, `ExactPartitioner`, is in #501 which includes this change; feel free to review starting here or all at once there.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/498)


src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
```diff
@@ -39,7 +39,7 @@ object GermlineAssemblyCaller {
     var minAlignmentQuality: Int = 30
 
     @Args4jOption(name = "--reference-fasta", required = true, usage = "Local path to a reference FASTA file")
-    var referenceFastaPath: String = null
+    var referenceFastaPath: String = _
```
timodonnell: I'd rather be explicit about what we're setting this value to, I don't remember off the top of my head if strings default to null or ""

src/main/scala/org/hammerlab/guacamole/loci/map/LociMap.scala
```diff
@@ -76,6 +90,26 @@ object LociMap {
   /** Construct an empty LociMap. */
   def apply[T](): LociMap[T] = LociMap(TreeMap[ContigName, Contig[T]]())
 
+  def load(is: InputStream): LociMap[PartitionIndex] = {
+    fromLines(LinesIterator(is))
+  }
+
+  def fromLines(lines: TraversableOnce[String]): LociMap[PartitionIndex] = {
```
timodonnell: need a docstring

src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala
```diff
@@ -0,0 +1,58 @@
...
+
+trait LociPartitionerArgs
+  extends LociArgs
+  with ApproximatePartitionerArgs
+  with UniformPartitionerArgs {
+
+  @Args4JOption(
+    name = "--loci-partitioning",
+    usage = "Directory path within which to save the partitioned reads and accompanying LociMap partitioning.",
+    forbids = Array("--partitioning-dir"),
```
timodonnell: Is `partitioning-dir` used?

Also might be helpful to clarify if this is only an output parameter (i.e., it's only written but never read, in which case maybe call it something like ``--out-loci-partitioning``) or both an output-and-an-input argument. Also from the name alone it would seem to me it's saving only loci, so if it's also saving reads another name (`--out-partitioned-reads`?) might be appropriate

src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioning.scala
```diff
@@ -0,0 +1,76 @@
...
+import org.hammerlab.guacamole.loci.map.LociMap
+import org.hammerlab.guacamole.loci.partitioning.LociPartitioner.PartitionIndex
+import org.hammerlab.guacamole.loci.set.LociSet
+import org.hammerlab.guacamole.reference.{Locus, ReferenceRegion}
+import org.hammerlab.magic.util.{Saveable, Stats}
+
+import scala.reflect.ClassTag
+
+case class LociPartitioning(map: LociMap[PartitionIndex]) extends Saveable {
+  @transient lazy val partitionsMap: Map[PartitionIndex, LociSet] = map.inverse
```
timodonnell: A comment about why these are `@transient lazy val` would be useful (guessing it's serialization?) Also a class-level docstring with a sentence for what this class represents would be helpful

timodonnell:
LGTM modulo a couple small comments

ryan-williams:
Review status: 0 of 25 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala, line 42 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/498#-KNV4HjgIM9MQ4sP7Snn-r1-42:-KNXp96NWnIPi7sDyYjo:-1135031845) ([raw file](https://github.com/hammerlab/guacamole/blob/2d0f8fa953199bd3a88f549c218d78ba4fd75db8/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala#L42)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> I'd rather be explicit about what we're setting this value to, I don't remember off the top of my head if strings default to null or ""
</details>
`String`s become `null`, but I reverted anyway. This was just an IJ suggestion but I disabled that inspection now.

---

*[src/main/scala/org/hammerlab/guacamole/loci/map/LociMap.scala, line 97 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/498#-KNV4HjhZhpHrcZlvSAl-r1-97:-KNXpLWPpMPUhOzo69Aq:318479528) ([raw file](https://github.com/hammerlab/guacamole/blob/2d0f8fa953199bd3a88f549c218d78ba4fd75db8/src/main/scala/org/hammerlab/guacamole/loci/map/LociMap.scala#L97)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> need a docstring
</details>
good point, done

---

*[src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala, line 20 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/498#-KNV4HjhZhpHrcZlvSAw-r1-20:-KNXpNyVmesYECK_jZmh:1167746807) ([raw file](https://github.com/hammerlab/guacamole/blob/2d0f8fa953199bd3a88f549c218d78ba4fd75db8/src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala#L20)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> Is `partitioning-dir` used?
> 
> 
> 
> Also might be helpful to clarify if this is only an output parameter (i.e., it's only written but never read, in which case maybe call it something like ``--out-loci-partitioning``) or both an output-and-an-input argument. Also from the name alone it would seem to me it's saving only loci, so if it's also saving reads another name (`--out-partitioned-reads`?) might be appropriate
</details>
ah, yea this had some cruft from later changes, sorry.

I've updated it now: `--loci-partitioning` loads a partitioning if the path exists; if not then it computes a partitioning and writes it to this path. If the option is not set it does neither.

---

*[src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioning.scala, line 16 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/498#-KNV4HjhZhpHrcZlvSAx-r1-16:-KNXqf_Wl9fO3KTJziHX:802894410) ([raw file](https://github.com/hammerlab/guacamole/blob/2d0f8fa953199bd3a88f549c218d78ba4fd75db8/src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioning.scala#L16)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> A comment about why these are `@transient lazy val` would be useful (guessing it's serialization?) Also a class-level docstring with a sentence for what this class represents would be helpful
</details>
good calls. I've removed all these `val`s as they actually only come in to play later. added a docstring for now too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/498)*

ryan-williams:
hm reviewable did not put my comment responses on github, only reviewable i guess, in any case I think I responded to everything @timodonnell, thx for the catches, was able to clean things up a fair amount here!

src/main/scala/org/hammerlab/guacamole/loci/partitioning/ApproximatePartitioner.scala
```diff
@@ -0,0 +1,190 @@
...
+ * @param regions: RDD of reads to base the partitioning on.
+ * @param numPartitions Number of partitions to split reads into.
+ * @param microPartitionsPerPartition Long >= 1. Number of micro-partitions generated for each of the `numPartitions`
+ *                                    Spark partitions that will be computed. Higher values of this will result in a
+ *                                    more exact but more expensive computation.
+ *                                    In the extreme, setting this to greater than the number of loci (per partition)
+ *                                    will result in an exact calculation.
+ * @return LociMap of locus -> partition assignments.
+ */
+class ApproximatePartitioner[R <: ReferenceRegion: ClassTag](regions: RDD[R],
```
timodonnell: I'm not sure if `ApproximatePartitioner` is the right name for this. Maybe `ApproximateRegionDepthPartitioner`? Then your new partitioner would be something like `ApproximateRegionDepthWithThresholdPartitioner`. Not that these huge names are that lovely, but I don't think Approximate vs. Exact is really the distinction between this partitioner and yours, unless I'm misunderstanding yours.

timodonnell:
*[src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala, line 20 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/498#-KNV4HjhZhpHrcZlvSAw-r1-20:-KNcplRoMiyd4J7Fpnt4:1053855624) ([raw file](https://github.com/hammerlab/guacamole/blob/2d0f8fa953199bd3a88f549c218d78ba4fd75db8/src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala#L20)):*
<details><summary><i>Previously, ryan-williams (Ryan Williams) wrote…</i></summary>
> ah, yea this had some cruft from later changes, sorry.
> 
> I've updated it now: `--loci-partitioning` loads a partitioning if the path exists; if not then it computes a partitioning and writes it to this path. If the option is not set it does neither.
</details>
That behavior should be documented in the help string for this argument. Also if it's not already done the code should print a progress about whether that file is being read or written

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/498)*

src/main/scala/org/hammerlab/guacamole/loci/partitioning/ApproximatePartitioner.scala
ryan-williams: Good point, these names are not great. 

I changed it to `MicroRegionPartitioner`, which I think is good on descriptiveness and brevity.

lmk if you have a different preference.

ryan-williams:
Review status: 0 of 25 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala, line 20 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/498#-KNV4HjhZhpHrcZlvSAw-r1-20:-KNcthiUfq-HGI-mxvDB:1151240227) ([raw file](https://github.com/hammerlab/guacamole/blob/2d0f8fa953199bd3a88f549c218d78ba4fd75db8/src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala#L20)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> That behavior should be documented in the help string for this argument. Also if it's not already done the code should print a progress about whether that file is being read or written
</details>
cool, documented in help string and log `progress`es now

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/498)*

ryan-williams:
OK, I think everything's addressed @timodonnell

timodonnell:
Cool, LGTM

ryan-williams:
oops, only renamed `ApproximatePartitionerArgs` and not `ApproximatePartitioner`, 1s

ryan-williams:
OK I think this is all set, will merge on green

PR #501, opened 2016-07-25
Add "capped regions" loci partitioner

ryan-williams:
builds on #498, feel free to review all of it here or start there then pick this back up once it's merged.

The commits here should also stand alone if you want to review them one by one.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/501)


src/main/scala/org/hammerlab/guacamole/loci/partitioning/ExactPartitioner.scala
```diff
@@ -0,0 +1,113 @@
...
+        if (runs.length == 1)
+          s"$rs"
+        else {
+          val total = runs.map(_._1.toLong).sum
+          s"${runs.length} runs, $total loci (avg %.1f): $rs".format(total.toDouble / runs.length)
+        }
+      }
+
+      progress(
+        s"$validLoci (%.1f%%) valid loci, $invalidLoci invalid ($totalLoci total of ${loci.count} eligible)${overflowMsg}"
```
timodonnell: nit: might be clearer to say explicitly "sites with depth < $maxRegionsPerPartition" instead of "invalid loci"

src/main/scala/org/hammerlab/guacamole/readsets/rdd/RegionRDD.scala
```diff
@@ -0,0 +1,192 @@
...
+object RegionRDD {
+  private val rddMap = mutable.Map[Int, RegionRDD[_]]()
+  implicit def rddToRegionRDD[R <: ReferenceRegion: ClassTag](rdd: RDD[R]): RegionRDD[R] =
+    rddMap.getOrElseUpdate(
+      rdd.id,
+      new RegionRDD[R](rdd)
+    ).asInstanceOf[RegionRDD[R]]
+
+  /**
+   * Given an RDD output by {{RegionRDD.partitionDepths}}, count the number of loci below (resp. above) the depth
```
timodonnell: I think you mean "keyed by" in `Given an RDD output by`? If you could be a little more explicit in the comment on what `depthRuns` contains that'd be helpful

src/main/scala/org/hammerlab/guacamole/reference/ContigIterator.scala
```diff
@@ -0,0 +1,33 @@
...
+      regions.head
+    else
+      throw new NoSuchElementException
+
+  override def hasNext: Boolean = {
+    regions.hasNext && regions.head.contigName == contigName
+  }
+
+  override def next(): R = {
+    val n = regions.head
```
timodonnell: Did you mean `val n = head` instead of `val n = regions.head`? I think as-is it might not throw if the next region has a different contig

src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala
```diff
@@ -0,0 +1,55 @@
+package org.hammerlab.guacamole.reference
+
+/**
+ * Helpers for dealing with contig-name strings.
+ */
+object ContigName {
+
+  // Sort order: chr1, chr2, …, chr22, chrX, chrY, <other>.
```
timodonnell: Ah I'm realzing now this PR is actually a bit dated, since we already talked about this contig ordering stuff -- I'll hold off on reviewing any more of it now. Sorry about that!

ryan-williams:
thx for the comments Tim, I addressed them all and removed a few other bits of functionality that slipped in from future PRs to make this all cleaner.

---

Review status: 0 of 58 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/main/scala/org/hammerlab/guacamole/loci/partitioning/ExactPartitioner.scala, line 87 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-KNVInfR22nF8vZNj0Ze-r1-87:-KNY-9H92OqPnzDdJ9dQ:-595262294) ([raw file](https://github.com/hammerlab/guacamole/blob/149b4fd3b2e88a1efe6d32710ab5e3b3aa6b30d0/src/main/scala/org/hammerlab/guacamole/loci/partitioning/ExactPartitioner.scala#L87)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> nit: might be clearer to say explicitly "sites with depth < $maxRegionsPerPartition" instead of "invalid loci"
</details>
sure, reworded

---

*[src/main/scala/org/hammerlab/guacamole/readsets/rdd/RegionRDD.scala, line 177 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-KNVInfR22nF8vZNj0Zm-r1-177:-KNY-CXo2PH4VYgDL1Mx:-843586734) ([raw file](https://github.com/hammerlab/guacamole/blob/149b4fd3b2e88a1efe6d32710ab5e3b3aa6b30d0/src/main/scala/org/hammerlab/guacamole/readsets/rdd/RegionRDD.scala#L177)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> I think you mean "keyed by" in `Given an RDD output by`? If you could be a little more explicit in the comment on what `depthRuns` contains that'd be helpful
</details>
no, I meant "an RDD *that was* output by `partitionDepths`", but I've reworked and better-commented this API, it was definitely clunky, lmk if the new way doesn't make sense.

---

*[src/main/scala/org/hammerlab/guacamole/reference/ContigIterator.scala, line 22 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-KNVInfR22nF8vZNj0Zn-r1-22:-KNY2GfzBcZ4tF9Qptv9:-80385028) ([raw file](https://github.com/hammerlab/guacamole/blob/149b4fd3b2e88a1efe6d32710ab5e3b3aa6b30d0/src/main/scala/org/hammerlab/guacamole/reference/ContigIterator.scala#L22)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> Did you mean `val n = head` instead of `val n = regions.head`? I think as-is it might not throw if the next region has a different contig
</details>
good catch! this scared me for a second since I thought it meant a bunch of stuff that uses this had not noticed this bug, but then I realized that since `hasNext` is written correctly here, the incorrect behavior would only happen if a caller repeatedly called `next` on this iterator without checking `hasNext`, which nothing else does (or should do), but I've updated it now so that that case can't happen, and added tests both of typical behavior and this case in particular; thanks!

---

*[src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-KNVInfR22nF8vZNj0Zo-r1-8:-KNY3GEtGPpu3GpgkiuS:1369570486) ([raw file](https://github.com/hammerlab/guacamole/blob/149b4fd3b2e88a1efe6d32710ab5e3b3aa6b30d0/src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala#L8)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> Ah I'm realzing now this PR is actually a bit dated, since we already talked about this contig ordering stuff -- I'll hold off on reviewing any more of it now. Sorry about that!
</details>
heh, actually this is intentional, sorry I wasn't clear. I do actually want logical-chr order in my debug-printlns about the loci-partitioning, which is what uses this code.

the locimap/lociset contig-ordering is still lexicographic as of this PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-:-KNY4LbP24cIpEBGm1A3:-341923255)*

timodonnell:
*[src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-KNVInfR22nF8vZNj0Zo-r1-8:-KNcqil7ZB3XN_dHvbXJ:1124503180) ([raw file](https://github.com/hammerlab/guacamole/blob/149b4fd3b2e88a1efe6d32710ab5e3b3aa6b30d0/src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala#L8)):*
<details><summary><i>Previously, ryan-williams (Ryan Williams) wrote…</i></summary>
> heh, actually this is intentional, sorry I wasn't clear. I do actually want logical-chr order in my debug-printlns about the loci-partitioning, which is what uses this code.
> 
> the locimap/lociset contig-ordering is still lexicographic as of this PR.
</details>
I'm not a fan of this code. It's not clear to me what it will end up affecting since it's defining a new ordering for ContigName 's which are used throughout the code. My understanding is anything imports this will use it, is that right? It's also the only place we have logic based on what the reference contig names are. If it's only used in debug messages would it be possible to just add a helper that prints out the message you want with the desired sort order without doing anything involving with potentially global implications? If some version of this has to stay we should put a prominent message that it's only used for debug output and we are not actually assuming a particular reference / species in guacamole

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/501)*

ryan-williams:
*[src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-KNVInfR22nF8vZNj0Zo-r1-8:-KNd0RjmhWoZiQ6CROCG:243309424) ([raw file](https://github.com/hammerlab/guacamole/blob/149b4fd3b2e88a1efe6d32710ab5e3b3aa6b30d0/src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala#L8)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> I'm not a fan of this code. It's not clear to me what it will end up affecting since it's defining a new ordering for ContigName 's which are used throughout the code. My understanding is anything imports this will use it, is that right? It's also the only place we have logic based on what the reference contig names are. If it's only used in debug messages would it be possible to just add a helper that prints out the message you want with the desired sort order without doing anything involving with potentially global implications? If some version of this has to stay we should put a prominent message that it's only used for debug output and we are not actually assuming a particular reference / species in guacamole
</details>
the `implicit Ordering` defined here will not be used implicitly unless it is in scope at some call-site that wants an implicit `Ordering[String]`.

nothing in the codebase at this point imports `ContigName.ordering` or `ContigName._`, so it's never in scope in that way.

it is only used, explicitly, [once](https://github.com/hammerlab/guacamole/pull/501/files#diff-62a886c44c86a5d2ce8866c4070b3f6aR62), by the print statements I referenced before.

you're right that having it be `implicit` is unnecessary and makes it easier to have some unexpected behavior, so I've removed the `implicit` here, now.

[here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/501)*

ryan-williams:
*[src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-KNVInfR22nF8vZNj0Zo-r1-8:-KNd0RjmhWoZiQ6CROCG:2131503689) ([raw file](https://github.com/hammerlab/guacamole/blob/149b4fd3b2e88a1efe6d32710ab5e3b3aa6b30d0/src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala#L8)):*
<details><summary><i>Previously, ryan-williams (Ryan Williams) wrote…</i></summary>
> the `implicit Ordering` defined here will not be used implicitly unless it is in scope at some call-site that wants an implicit `Ordering[String]`.
> 
> nothing in the codebase at this point imports `ContigName.ordering` or `ContigName._`, so it's never in scope in that way.
> 
> it is only used, explicitly, [once](https://github.com/hammerlab/guacamole/pull/501/files#diff-62a886c44c86a5d2ce8866c4070b3f6aR62), by the print statements I referenced before.
> 
> you're right that having it be `implicit` is unnecessary and makes it easier to have some unexpected behavior, so I've removed the `implicit` here, now.
> 
> [here
</details>
grr reviewable lost the end of that

[here is a gist](https://gist.github.com/ryan-williams/34b60c39679cfd305a5387a9d0a0178e) showing how an implicit `Ordering[String]` has no effect if it's not directly imported; just importing the type alias / companion object does not do it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/501)*

timodonnell:
*[src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-KNVInfR22nF8vZNj0Zo-r1-8:-KNd6Fab6vaiO0RqNDmG:1009767829) ([raw file](https://github.com/hammerlab/guacamole/blob/149b4fd3b2e88a1efe6d32710ab5e3b3aa6b30d0/src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala#L8)):*
<details><summary><i>Previously, ryan-williams (Ryan Williams) wrote…</i></summary>
> the `implicit Ordering` defined here will not be used implicitly unless it is in scope at some call-site that wants an implicit `Ordering[String]`.
> 
> nothing in the codebase at this point imports `ContigName.ordering` or `ContigName._`, so it's never in scope in that way.
> 
> it is only used, explicitly, [once](https://github.com/hammerlab/guacamole/pull/501/files#diff-62a886c44c86a5d2ce8866c4070b3f6aR62), by the print statements I referenced before.
> 
> you're right that having it be `implicit` is unnecessary and makes it easier to have some unexpected behavior, so I've removed the `implicit` here, now.
> 
> [here
</details>
Yeah, but relying on not accidentally importing this in the wrong way is crazy error prone. I realize that without the implicit that is no longer an issue, but I would still rather see this code compressed into something less verbose (it really should be almost a one liner to do a numerical sort) and used directly in the debug-printing code. Having it here is just confusing as it makes it seem we rely on it. I'd be ok punting on it though to get the new partitioner in

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/501)*

ryan-williams:
Review status: 0 of 41 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-KNVInfR22nF8vZNj0Zo-r1-8:-KNd0RjmhWoZiQ6CROCG:-1030939924) ([raw file](https://github.com/hammerlab/guacamole/blob/149b4fd3b2e88a1efe6d32710ab5e3b3aa6b30d0/src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala#L8)):*
> it really should be almost a one liner to do a numerical sort

Sure, it's just a numerical sort, that also normalizes `chr1` vs. `1`, and also puts `X`/`chrX` and `Y`/`chrY` after the numbers, and also sorts everything else after that.

If you have a way you'd rather that be written, I'm all ears.

Having a small library that neatly encapsulates this logical ordering of the contigs we deal with constantly doesn't seem that crazy to me; the desire for this has come up multiple times around the lab, various tools in the ecosystem depend on specific sort orders being used, this is the most logical and widely used one, etc.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/501)*

timodonnell:
*[src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-KNVInfR22nF8vZNj0Zo-r1-8:-KNh54yqQeZxdmDLmluQ:1593835558) ([raw file](https://github.com/hammerlab/guacamole/blob/149b4fd3b2e88a1efe6d32710ab5e3b3aa6b30d0/src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala#L8)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> Yeah, but relying on not accidentally importing this in the wrong way is crazy error prone. I realize that without the implicit that is no longer an issue, but I would still rather see this code compressed into something less verbose (it really should be almost a one liner to do a numerical sort) and used directly in the debug-printing code. Having it here is just confusing as it makes it seem we rely on it. I'd be ok punting on it though to get the new partitioner in
</details>
I seem to somehow have lost your most recent reply on reviewable, but regarding another possible way to write this, would something like this work?

```
// Sort contig names like "chr4" or "7" first by a prefix string ("chr") then numerically on the number (4 or 7). 
// This way, chr10 comes after chr9.
val stringThenNumberPattern = """([A-Za-z]*)([0-9]+)""".r
contigs.sortBy(x => stringThenNumberPattern.findFirstMatchIn(x) match {
    case None => (x, -1)
    case Some(result) => (result.group(1), result.group(2).toInt)
})
```

I think this would give basically the same result without resorting to hard coding any assumptions about reference genomes. (Attempts to do things like remove "chr" prefix has really been a mistake in varcode, which is probably one reason I'm worried about this.)

Thanks for putting up with these nit picks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/501)*

timodonnell:
Not necessarily worth blocking on this PR, but just wanted to note that in trying to understand some of this PR I ended up looking at this code in the `magic-rdds` repo:

https://github.com/hammerlab/magic-rdds/blob/9f38cc826bc71e562f0bbe904be30e8a5d0e9873/src/main/scala/org/hammerlab/magic/iterator/GroupRunsIterator.scala

Which is undocumented. Do you see the `magic-rdds` as being a permanent dependency of guacamole? If so we should document it just like guacamole. If it's just a transient thing while we get stuff working and you plan to move stuff out of that repo (which would be my preference if it's not a lot of stuff) then I'm ok to punt on it for now.

ryan-williams:
Thx for the nudge on magic-rdds @timodonnell, I added a ton of documentation (and a Travis CI), hopefully it's more grokkable now.

magic-rdds contains a bunch of stuff that has nothing to do with genomics, most of which finds a use in Pageant or Guac, so I envision it being a permanent dependency of Guacamole.

This PR is currently pointing at a snapshot of magic-rdds, but I'll hopefully get a proper release cut in the next few days, just need sonatype to whitelist it.

ryan-williams:
Review status: 0 of 53 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-KNVInfR22nF8vZNj0Zo-r1-8:-KO3bM_xBP965B7LSNd5:-1185196207) ([raw file](https://github.com/hammerlab/guacamole/blob/149b4fd3b2e88a1efe6d32710ab5e3b3aa6b30d0/src/main/scala/org/hammerlab/guacamole/reference/ContigName.scala#L8)):*
<details><summary><i>Previously, timodonnell (Tim O'Donnell) wrote…</i></summary>
> I seem to somehow have lost your most recent reply on reviewable, but regarding another possible way to write this, would something like this work?
> 
> ```
> // Sort contig names like "chr4" or "7" first by a prefix string ("chr") then numerically on the number (4 or 7). 
> // This way, chr10 comes after chr9.
> val stringThenNumberPattern = """([A-Za-z]*)([0-9]+)""".r
> contigs.sortBy(x => stringThenNumberPattern.findFirstMatchIn(x) match {
>     case None => (x, -1)
>     case Some(result) => (result.group(1), result.group(2).toInt)
> })
> ```
> 
> I think this would give basically the same result without resorting to hard coding any assumptions about reference genomes. (Attempts to do things like remove "chr" prefix has really been a mistake in varcode, which is probably one reason I'm worried about this.)
> 
> Thanks for putting up with these nit picks!
</details>
Yea, that would do the same thing, but I think inlining that is worse on a couple axes than just having a cleanly-scoped sub-library for a logical task like this, especially since I think it will likely be reused.

Happy to hear more about your takeaways from varcode, but from what you've said here I'm not that worried; callers of this code will not have any prefixes removed, they're just going to get a sane (and widely standardized-on) sort order.

As one attempt to mitigate what might be bugging you about this, I folded the `Map` declaration into some code that generates the `Map`:

```
val map: Map[String, Int] = ((1 to 22).map(_.toString) ++ List("X", "Y")).zipWithIndex.toMap
```

I think this is slightly worse, but it saves some lines.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/501)*

timodonnell:
Cool, thanks!

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-:-KO7Gvxco6K_KLwPgpEc:580104278)*

timodonnell:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/501#-:-KO7H2g4DfHfdGZyO1ws:1417088721)*

PR #502, opened 2016-07-25
VAFHistogram/VariantSupport: handle multiple samples in one RDD pipeline

ryan-williams:
Previously these commands were processing N RDDs basically in sequence. Where possible, it's good to multiplex a given sequence of RDD operations, processing all the samples at once, keeping their elements distinguishable from one another's using some field as a key, to improve efficiency, allowing executors to flexibly perform tasks from any sample, etc.

I've done that here to VariantSupport and to some of the VAFHistogram pipeline.

However, VAFHistogram's mixture-model-building step uses an mllib API that doesn't allow for this sort of multiplexing, meaning it will necessarily process samples sequentially.

In such a case it makes more sense to handle processing multiple samples elsewhere in a pipeline than within one Spark app, which can only process them in serial; cf. discussion [here](https://github.com/bigdatagenomics/adam/issues/833).

So, this PR makes VAFHIstogram only create one model in a run. Previously this would have corresponded to only processing one sample per run, but I've also added functionality that allows it to process multiple samples at once and build a model across all of them, just to round out the distinctions about what kind of multi-sample processing makes sense where.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/502)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
@@ -124,8 +124,6 @@ object VAFHistogram {
 
       val ReadSets(_, _, contigLengths) = readsets
 
-      val mappedReadsRDDs = readsets.mappedReadsRDDs
-
       val partitionedReads =
         PartitionedRegions(
           readsets.mappedReadsRDDs,
@@ -189,7 +187,11 @@ object VAFHistogram {
       }
 
       if (args.cluster) {
-        val variantsPerSample = variantLoci.keyBy(_.sampleId).splitByKey(numVariantsPerSample)
+        val variantsPerSample =
+          variantLoci
+            .keyBy(_.sampleId)
+            .splitByKey(numVariantsPerSample)
+
         for {
           (sampleId, variants) <- variantsPerSample
         } {
@@ -202,9 +204,10 @@ object VAFHistogram {
   /**
    * Generates a count of loci in each variant allele frequency bins
    *
-   * @param variantAlleleFrequencies RDD of loci with variant allele frequency > 0
-   * @param bins Number of bins to group the VAFs into
-   * @return Map of rounded variant allele frequency to number of loci with that value
+   * @param variantAlleleFrequencies RDD of loci with variant allele frequency > 0.
+   * @param bins Number of bins to group the VAFs into.
+   * @return Map from sample name to per-sample sorted [[Vector]] of tuples containing a [rounded variant allele
+   *         frequency] and [the number of loci with that VAF].
    */
   def generateVAFHistograms(variantAlleleFrequencies: RDD[VariantLocus],
                             bins: Int): Map[SampleId, Vector[(Int, Long)]] = {
--- a/src/main/scala/org/hammerlab/guacamole/reads/MappedRead.scala
+++ b/src/main/scala/org/hammerlab/guacamole/reads/MappedRead.scala
@@ -49,7 +49,6 @@ case class MappedRead(
     isPositiveStrand: Boolean,
     isPaired: Boolean)
   extends Read
-
     with ReferenceRegion {
 
   assert(baseQualities.length == sequence.length,
--- a/src/test/scala/org/hammerlab/guacamole/commands/VariantSupportSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/commands/VariantSupportSuite.scala
@@ -40,10 +40,11 @@ class VariantSupportSuite extends GuacFunSuite with TableDrivenPropertyChecks {
     val computedAlleleCounts =
       (for {
         AlleleCount(_, _, _, ref, alternate, count) <- VariantSupport.Caller.pileupsToAlleleCounts(Vector(pileup))
-      } yield (ref, alternate, count)
+      } yield
+        (ref, alternate, count)
       )
-        .toArray
-        .sortBy(x => x)
+      .toArray
+      .sortBy(x => x)
 
     computedAlleleCounts should be(alleleCounts.sortBy(x => x))
   }


src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
```diff
@@ -202,24 +202,27 @@ object VAFHistogram {
    * @param bins Number of bins to group the VAFs into
    * @return Map of rounded variant allele frequency to number of loci with that value
```
arahuja: Update the return docs? This is now map of `sampleName -> Map[Int, Long]` might be nice to have a type alias of `type SampleName = String`

src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
```diff
@@ -228,56 +231,70 @@ object VAFHistogram {
...
+            ),
         reference
       )
+
     if (printStats) {
       variantLoci.persist(StorageLevel.MEMORY_ONLY)
 
-      val numVariantLoci = variantLoci.count
-      progress(s"$numVariantLoci non-zero variant loci in sample $sampleName")
+      val numVariantLociBySample = variantLoci.map(_.sampleName -> 1L).reduceByKey(_ + _).collect()
```
arahuja: Is this different than `countByValue` since it is collected at the end anyways?

src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
```diff
@@ -202,24 +202,27 @@ object VAFHistogram {
...
     def roundToBin(variantAlleleFrequency: Float) = {
       val variantPercent = (variantAlleleFrequency * 100).toInt
       variantPercent - (variantPercent % (100 / bins))
     }
+
     variantAlleleFrequencies
-      .map(vaf => roundToBin(vaf.variantAlleleFrequency) -> 1L)
+      .map(vaf => (vaf.sampleName, roundToBin(vaf.variantAlleleFrequency)) -> 1L)
       .reduceByKey(_ + _)
+      .map(t => t._1._1 -> Map(t._1._2 -> t._2))
```
arahuja: can `t` be renamed here?

src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
```diff
@@ -182,14 +181,15 @@ object VAFHistogram {
...
+        } {
           histogram.toSeq.sortBy(_._1).foreach(kv => println(histogramToString(kv)))
-        )
+        }
       }
 
       if (args.cluster) {
-        val numClusters = args.numClusters
-        variantLoci.foreach(buildMixtureModel(_, numClusters))
+        buildMixtureModel(variantLoci, args.numClusters)
```
arahuja: This seems like different behavior now? Instead of building a model per sample it is just building one with all variant loci?

src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
ryan-williams: good point, changed

src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
ryan-williams: good catch and good idea, both done

src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
ryan-williams: done, unrolled into a `case` unapply

src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
ryan-williams: yea, that's right and basically intentional, though I should have explained my thinking!

I view this as two changes in sequence:

1. spark apps that are pushing multiple datasets (samples) through a pipeline where each is independent of the others, like this was doing (e.g. `.map` over an array of `RDD`s) should be changed to only process one sample at a time, and the calling scripts / workflows should handle that multiplexing over multiple samples instead. In my experience, pushing that [sequential processing of multiple RDDs] into one Spark app has disadvantages / Spark is the wrong part of the stack to do that. https://github.com/bigdatagenomics/adam/issues/833 has some context where I wanted to move ADAM cmds toward performing multiple independent "commands" within a given Spark app, but was talked out of it and convinced that the opposite approach is superior. Spark doesn't parallelize its processing of an array of `RDD`s, so it's really the worst possible place to multiplex on samples, since no parallelization is possible.

2. given the above, and imagining this command now only operates on a single sample, I've generalized it to be able to handle multiple samples in one pass / as part of one mixture model; this is different than the previous "multiple independent samples/models in one app" setup and does actually let Spark operate on multiple samples at once, albeit to answer a different question than was being answered before, but per 1. the previous question is better answered by parallelizing elsewhere in the stack.

Sorry for the wall of text, lmk if that doesn't make sense.

ryan-williams:
Responded to everything, thx @arahuja lmk if this looks good

src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
arahuja: I don't think I get this completely. It seems like a lot of the code changes here are to more efficiently handle multiple inputs, which seems great, but then this part produces a different result than it did before.  It seems either this should support multiple samples (and then give multiple models) or just a single sample. But now it seems structured to take multiple samples and produce a single model, which wouldn't be a relevant use case?  

It just seems like something different then it did before, defeats the purpose of making it more efficient?

ryan-williams: AFAIK there's no way to apply the same trick I applied elsewhere (handling multiple independent samples in one RDD, by "keying" elements with their sample ID) to the mixture-model-building step; the mllib API we're using doesn't allow for building multiple models at once based on e.g. the key of a tuple, so this step is going to be happen in serial across samples either way.

My argument is that it's better to push that sequencing of multiple independent RDD pipelines to a different part of the stack than inside one Spark app like it's been, since Spark can't parallelize such operations at all.

Given all that, I've also added here the ability to build one model that incorporates data from multiple samples. I'm not saying that's an important use case, but more like a "might as well" after the initial step, described above, of reducing the contract here to only building one model per Spark app.

You're right that that's now how I described it in the PR message above; I'll go change that.

src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
```diff
@@ -179,30 +178,38 @@ object VAFHistogram {
...
       }
 
       if (args.cluster) {
-        val numClusters = args.numClusters
-        variantLoci.foreach(buildMixtureModel(_, numClusters))
-      }
+        val variantsPerSample =
+          variantLoci
+            .keyBy(_.sampleId)
+            .splitByKey(numVariantsPerSample)
```
ryan-williams: OK @arahuja, I took a new approach here that retains the mixture-per-sample behavior.

I wrote [a `RDD.splitByKey` function in magic-rdds](https://github.com/hammerlab/magic-rdds/blob/f7309c181589d37af9baae03627fbf085b208091/src/main/scala/org/hammerlab/magic/rdd/SplitByKeyRDD.scala#L7-L16) which will incur one shuffle in order to then pass a separate RDD to the mixture-API for each key (in this case sample ID).

Let me know if that assuages your concerns.

arahuja:
Probably easiest just to cut this from`guacamole`, have a new `guacamole` release and have one-off/custom jobs like this live independently and just rely on the pileup processing infrastructure in guac. 

Could take it back to #329 as well and have things like core processing primitives and calls and jobs live independently as well.  Either way, probably time for a new `guacamole` release anyways

ryan-williams:
I don't disagree with any of that, but it also seems orthogonal.

This should be back to feature-parity with the original code, plus some efficiency improvements.

ryan-williams:

PR #504, opened 2016-07-25
rm pileup overlapsLocus check

ryan-williams:
I could be wrong, but I can't see why we need the check in `Pileup.apply`, likewise the couple test cases that verify that non-overlapping reads aren't included in Pileups.

Let me know what you think!

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/504)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/pileup/Pileup.scala
+++ b/src/main/scala/org/hammerlab/guacamole/pileup/Pileup.scala
@@ -128,7 +128,7 @@ object Pileup {
   /**
    * Given reads and a locus, returns a [[Pileup]] at the specified locus.
    *
-   * @param reads Sequence of reads, in any order, that may or may not overlap the locus.
+   * @param reads Sequence of reads, in any order, that must overlap the locus.
    * @param contigName The contig these reads lie on.
    * @param locus The locus to return a [[Pileup]] at.
    * @param contigSequence The reference for this pileup's contig
@@ -138,8 +138,7 @@ object Pileup {
             contigName: ContigName,
             locus: Locus,
             contigSequence: ContigSequence): Pileup = {
-    //TODO: Is this call to overlaps locus necessary?
-    val elements = reads.filter(_.overlapsLocus(locus)).map(PileupElement(_, locus, contigSequence))
+    val elements = reads.map(PileupElement(_, locus, contigSequence))
     Pileup(contigName, locus, contigSequence, elements.toIndexedSeq)
   }
 }
--- a/src/test/scala/org/hammerlab/guacamole/pileup/PileupSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/pileup/PileupSuite.scala
@@ -211,6 +211,7 @@ class PileupSuite extends GuacFunSuite with TableDrivenPropertyChecks {
   test("Loci 10-19 deleted from half of the reads") {
     val pileup = loadPileup(sc, "same_start_reads.sam", locus = 0, reference = reference)
     val deletionPileup = pileup.atGreaterLocus(9, Seq.empty.iterator)
+
     deletionPileup.elements.map(_.alignment).count {
       case Deletion(bases, _) => {
         Bases.basesToString(bases) should equal("AAAAAAAAAAA")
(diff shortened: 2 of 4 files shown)


src/main/scala/org/hammerlab/guacamole/pileup/Pileup.scala
```diff
@@ -150,8 +150,7 @@ object Pileup {
    * @return A [[Pileup]] at the given locus.
    */
   def apply(reads: Seq[MappedRead], referenceName: String, locus: Long, referenceContigSequence: ContigSequence): Pileup = {
-    //TODO: Is this call to overlaps locus necessary?
-    val elements = reads.filter(_.overlapsLocus(locus)).map(PileupElement(_, locus, referenceContigSequence))
+    val elements = reads.map(PileupElement(_, locus, referenceContigSequence))
```
timodonnell: I think if you remove the filter then this doc string:

```
@param reads Sequence of reads, in any order, that may or may not overlap the locus.
```

Should change to something like "Sequence of reads, in any order, that must overlap the locus."

Were you seeing overhead in all the `overlaps` checks?

timodonnell:
After this change, what happens if you make a pileup using reads that do not overlap the locus? If it just throws an exception, then this change is ok by me. If we get any kind of correctness issue though then we should either leave as is or make it throw an exception.

src/main/scala/org/hammerlab/guacamole/pileup/Pileup.scala
ryan-williams: Good point about docstring, done!

> Were you seeing overhead in all the `overlaps` checks?

Nope, it's just an extra O(n) over all reads at a spot that may be one of the "tighter loops" in the codebase, so I was scrutinizing it for that reason.

Having thought about it more, though:
- all calling code makes sure to only throw reads at `Pileup.apply` that it knows should be there, and with good reason, since in the cases we currently handle, you wouldn't want to build a `Pileup` out of a bunch of reads whose proximity to the locus in question is not known.
- the `PileupElement` ctor calls `advanceToLocus` which throws an assertion error if a read doesn't overlap; I prefer that behavior here since in general we shouldn't be passing irrelevant reads to `Pileup`.

ryan-williams:
Good q, my answer is [above](https://github.com/hammerlab/guacamole/pull/504#discussion_r72139668), you always get assertion errors, so I think this is the right way to go, having paged this all back in now :) thanks!

ryan-williams:
woops I think I changed this to my version that verifies we get assertion errors, let me fix that

src/test/scala/org/hammerlab/guacamole/pileup/PileupSuite.scala
```diff
@@ -89,9 +86,6 @@ class PileupSuite extends GuacFunSuite with TableDrivenPropertyChecks {
       TestUtil.makeRead("TCGACCCTCGA", "4M3I4M", 1, "chr1", Some(Seq(10, 15, 20, 25, 5, 5, 5, 10, 15, 20, 25)))
     )
 
-    val noPileup = Pileup(reads, "chr1", 0, reference.getContig("chr1")).elements
-    noPileup.size should be(0)
-
```
arahuja: should these be assertion error tests?

ryan-williams: sure, good call, will add momentarily!

ryan-williams: heh, looks like [we already test that in `PileupSuite`](https://github.com/hammerlab/guacamole/blob/14a25a141e94e367a47fb1cd1ee18f112b43a523/src/test/scala/org/hammerlab/guacamole/pileup/PileupSuite.scala#L252-L273).

arahuja: Ah ok so a `PileupElement` can't be created, so therefore a `Pileup` can't be. Got it - feel free to add or not, up to you. More tests never hurt.

ryan-williams: fair enough, will add a direct test of Pileup-creation

ryan-williams: OK, added

timodonnell:
LGTM

PR #506, opened 2016-07-25
fix forced-loci loading

ryan-williams:
use the same logic as regular loci-loading

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/506)

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/main/SomaticJointCallerIntegrationTests.scala
+++ b/src/test/scala/org/hammerlab/guacamole/main/SomaticJointCallerIntegrationTests.scala
@@ -95,7 +95,7 @@ object SomaticJointCallerIntegrationTests extends SparkCommand[Arguments] with V
         args.out = resultFile
         args.paths = Seq(NA12878TestUtil.subsetBam).toArray
         args.loci = "chr1:0-6700000"
-        args.forceCallLociFromFile = NA12878TestUtil.expectedCallsVCF
+        args.forceCallLociFile = NA12878TestUtil.expectedCallsVCF
         args.referenceFastaPath = NA12878TestUtil.chr1PrefixFasta
         SomaticJoint.Caller.run(args, sc)
       }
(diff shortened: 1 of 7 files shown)


timodonnell:
LGTM, thanks @ryan-williams

arahuja:
As a note, this drops support of using a VCF as an input for loci

ryan-williams:
hm good point, let me add that back in, seems worth keeping around unless anyone feels otherwise

timodonnell:
Woah good catch @arahuja , yeah I definitely use that

src/test/scala/org/hammerlab/guacamole/main/SomaticJointCallerIntegrationTests.scala
```diff
@@ -95,7 +95,7 @@ object SomaticJointCallerIntegrationTests extends SparkCommand[Arguments] with V
         args.out = resultFile
         args.paths = Seq(NA12878TestUtil.subsetBam).toArray
         args.loci = "chr1:0-6700000"
-        args.forceCallLociFromFile = NA12878TestUtil.expectedCallsVCF
+        args.forceCallLociFile = NA12878TestUtil.expectedCallsVCF
```
arahuja: Yea - shouldn't have something failed here? Or because there is no assert this errors quietly?

ryan-williams: This test doesn't run as part of Travis / `mvn test`; cf. recent discussion at https://github.com/hammerlab/guacamole/issues/483#issuecomment-231597188.

Might make sense to have a test of the loci-from-VCF functionality though; I'll add that to this PR.

ryan-williams:
@arahuja @timodonnell this now has vcf-loading again and a toy test of that functionality

timodonnell:
LGTM

PR #508, opened 2016-07-26
add somatic-joint-caller end-to-end test

### LINKED ISSUE #507: run joint caller integration tests in travis
It would be useful to continuously verify that we can run the joint caller and get reasonable output. That test is currently quite slow though so it has been removed from the unit tests. It would be great to have it run as part of travis though. Some profiling and performance optimization of it might be useful too. 

See discussion at https://github.com/hammerlab/guacamole/issues/483#issuecomment-231597188

ryan-williams:
fixes #507

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/508)


timodonnell:
Thanks @ryan-williams . We can see if having this test is enough to keep the joint caller integration run unbroken going forward. I will probably change it at some point to have a more fine grained assertion than that the whole output file is unchanged, as we may be changing INFO columns etc often, but I can take care of that when needed. LGTM

ryan-williams:
yea sounds good. I just had to tweak it to skip over the `##reference` line of the `all.vcf` file since that points to a system-dependent path.

definitely open to doing more deliberate checking in this test

PR #514, opened 2016-08-03
rm mistakenly committed files

ryan-williams:
these snuck in to #501 somewhere in my rebasing saga

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/514)


arahuja:
Not sure where these were added, but looks good

ryan-williams:
yea, it was in #501, my bad

PR #515, opened 2016-08-04
SampleId/Name type aliases

ryan-williams:
use them more consistently

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/515)


arahuja:
👍

PR #516, opened 2016-08-04
ReadSets-args refactoring

ryan-williams:
Share sample-loading, filename/sample infra across all commands.

ReadSets now keeps sample names/paths around explicitly in a way that is tied to the corresponding RDD[Read]s.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/516)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
@@ -13,7 +13,7 @@ import org.hammerlab.guacamole.logging.DelayedMessages
 import org.hammerlab.guacamole.logging.LoggingUtils.progress
 import org.hammerlab.guacamole.pileup.Pileup
 import org.hammerlab.guacamole.reads.MappedRead
-import org.hammerlab.guacamole.readsets.ReadSets
+import org.hammerlab.guacamole.readsets.{ReadSets, SampleName}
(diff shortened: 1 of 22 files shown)


src/main/scala/org/hammerlab/guacamole/commands/jointcaller/Input.scala
```diff
@@ -16,10 +17,11 @@ import org.hammerlab.guacamole.readsets.{SampleId, SampleName}
  * @param analyte rna or dna
  */
 case class Input(index: SampleId,
-                 sampleName: SampleName,
```
timodonnell: we should just call this class `JointCallerInput` instead of having multiple `Input` classes in the code base

timodonnell:
Seems ok to me. As discussed off line, just don't want to follow this trend into refactoring joint caller stuff into the whole code base.

Did the usage for any commands change?

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/Input.scala
ryan-williams: We already do! Its name is `…jointcaller.Input`.

But seriously; defensively naming things to try to avoid conflicts at possible `import` sites is not good practice.

ryan-williams:
> don't want to follow this trend into refactoring joint caller stuff into the whole code base.

I don't follow what you mean here.

> Did the usage for any commands change?

No, afaik.

PR #517, opened 2016-08-04
rm unused Pileup.bySample

ryan-williams:
doesn't make sense for it to coexist with Pileup.sampleName, which
assumes a Pileup contains reads from only one sample, which is what
all callers in guacamole currently do.

this is a step towards clarifying that contract.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/517)


arahuja:
👍

PR #520, opened 2016-08-07
PartitionedRegions improvements

ryan-williams:
push the `PartitionedRegions` abstraction out to callers, so that they can start to separate loci+read partitioning plumbing from application logic.

recapitulates some of #502 but keying off `SampleId`s instead of `SampleName`s; still collapses the VAFHistogram mixture-model into one for all samples in an app, per https://github.com/hammerlab/guacamole/pull/502#discussion_r73574358

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/520)


src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
```diff
@@ -1,98 +1,214 @@
...
- * RDD, downstream code can iterate over the partitions emitted here and build full pileups for the loci that each
- * partition corresponds to.
- *
- * [[WindowFlatMapUtils]] calls this, and uses [[SlidingWindow]] to iterate over the partitioned regions in a manner
- * suitable for pileup-construction by [[PileupFlatMapUtils]].
+ * This means some regions will occur multiple times in the RDD (due to regions straddling partition boundaries, so it's
+ * important not to confuse this with a regular RDD[ReferenceRegion].
  */
+class PartitionedRegions[R <: ReferenceRegion: ClassTag](val numSamples: NumSamples,
+                                                         @transient keyedRegions: RDD[(SampleId, R)],
```
timodonnell: Need to explain what the `@transient` is for

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
```diff
@@ -1,98 +1,214 @@
...
+  assert(
+    keyedRegions.getNumPartitions == lociSetsRDD.getNumPartitions,
+    s"reads partitions: ${keyedRegions.getNumPartitions}, loci partitions: ${lociSetsRDD.getNumPartitions}"
+  )
+
+  def sc: SparkContext = keyedRegions.sparkContext
+
+  val numLoci = sc.accumulator(0L, "numLoci")
+
+  def mapPartitions[V: ClassTag](f: (Iterator[(SampleId, R)], LociSet) => Iterator[V]): RDD[V] =
```
timodonnell: need a header comment

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
```diff
@@ -1,98 +1,214 @@
...
+      )
+
+  private lazy val partitionLociSets: Array[LociSet] =
+    partitioning
+      .inverse
+      .toArray
+      .sortBy(_._1)
+      .map(_._2)
+
+  lazy val lociSetsRDD: RDD[LociSet] =
```
timodonnell: I'm confused that we're putting lociSets into an RDD. A comment here explaining why would be helpful

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
```diff
@@ -1,98 +1,214 @@
...
+   * @tparam R ReferenceRegion type.
+   */
+  def apply[R <: ReferenceRegion: ClassTag](regionRDDs: PerSample[RDD[R]],
+                                            loci: LociSet,
+                                            args: LociPartitionerArgs): PartitionedRegions[R] =
+    apply(regionRDDs, loci, args, halfWindowSize = 0)
+
+  /**
+   * @param regionRDDs RDD of regions to partition.
+   * @param loci A [[LociPartitioning]] describing which ranges of genomic loci should be mapped to each Spark
```
timodonnell: In the code lLoci` seems to be a `LociSet` not a `LociPartitioning`. (I also find that confusing though -- doesn't this want some kind of map from LociSets to partitions?) A one sentence intro comment to this function would be helpful

Also same for the override above.

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
```diff
@@ -1,98 +1,214 @@
...
+    val partitionedRegions: RDD[(SampleId, R)] =
+      (for {
+        // For each region…
+        (sampleId, region) <- keyedRegions
+
+        // Partitions that should receive a copy of this region.
+        partitions = partitioningBroadcast.value.getAll(region, halfWindowSize)
+
+        // Update the copies-per-region histogram accumulator.
+        _ = (regionCopiesHistogram += partitions.size)
```
timodonnell: This syntax confuses me, can it be made more explicit?

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
```diff
@@ -1,98 +1,214 @@
...
+
+        // For each destination partition…
```
timodonnell: I think some non-ascii ellipses have snuck in again

timodonnell:
Wow, there's a lot here. Left some comments. My main question is how this affects runtimes on the cluster, since it involves a lot of spark partitioning logic that I don't have a good understanding of the performance implications

Also a style point fwiw: I find it helpful for methods to have an intro describing what it does, in addition to the argument descriptions

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
ryan-williams: I don't think it's necessary, but defensive `@transient`ing seems like reasonable practice, since e.g. spark shells will pick up and serialize any variables that are in scope if a closure defined in the shell is passed to RDD.map or equivalent.

I'll add a comment saying as much.

ryan-williams: ah, I lied, it gets picked up by the `ClosureCleaner` when `mapPartitions` is executed below, so this is impt. will comment

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
ryan-williams: good point, done

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
ryan-williams: good call, added a comment explaining…

this is an RDD with one LociSet per partition, representing the loci that are assigned to that partition.

instead of broadcasting the LociSets for all partitions to every executor, like we did previously, this is just a way to send only each partition's LociSet just to it.

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
ryan-williams: good call, these scaladocs were not correct for this code, sorry. I've updated them now.

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
ryan-williams: Seems especially harmless (I would say desirable) in comments…

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
ryan-williams: yea, `for` comprehensions don't let you do no-ops (like this accumulator increment) directly, so I'm just assigning it to an unused `_` here.

per other discussion, I still believe this block to be much cleaner as a `for` than a `flatMap` and `map`

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
ryan-williams: It's about 8 years too late to be afraid of unicode characters in documentation 😈

timodonnell: Yeah I'm down to give up on the unicode debate, seems you feel stronger than I do. My main motivation was that some tools don't seem to handle it well (e.g. `git diff`, which pipes to less, on at least my terminal seems to display garbage characters for unicode). (It [appears](http://stackoverflow.com/questions/17320721/git-diff-shows-unicode-symbols-in-angle-brackets) that is `less`'s fault.) If you still think overall unicode is worth it though then let's do it

src/main/scala/org/hammerlab/guacamole/loci/map/LociMap.scala
```diff
@@ -61,6 +61,9 @@ case class LociMap[T](@transient private val map: SortedMap[ContigName, Contig[T
     mapOfBuilders.mapValues(_.result).toMap
   }
 
+  def getAll(r: ReferenceRegion, halfWindowSize: Int = 0): Set[T] =
```
timodonnell: could use a header comment

timodonnell:
Cool, thanks for the replies @ryan-williams . I took another pass, LGTM

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegions.scala
ryan-williams: Weird, `git diff` to `less` works fine for me:



Happy to help you debug that if you want.

[This SO](http://serverfault.com/a/414763) suggests `export LESSCHARSET=utf-8`, but I don't have that set so it's definitely not a necessary condition.

[The next answer there](http://serverfault.com/a/414762) does list a few vars that I have set:

```bash
export LANG=en_US.UTF-8
export LANGUAGE=en_US.UTF-8
export LC_ALL="en_US.UTF-8"
```

So you could try setting those.

src/main/scala/org/hammerlab/guacamole/loci/map/LociMap.scala
ryan-williams: good point, done

PR #521, opened 2016-08-07
rm Pileup.head accessor

ryan-williams:
not an API we should expose; tests can reach into elements directly
for now

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/521)


timodonnell:
lgtm

PR #523, opened 2016-08-07
expand some blocks over more lines for legibility

ryan-williams:
cf. https://github.com/hammerlab/guacamole/pull/496#discussion_r73795483

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/523)


timodonnell:
Comments basically the same as in the other PR. I'd rather not have new lines at the start of blocks or have lines with only ending parens on them, but other than that lgtm

ryan-williams:

PR #526, opened 2016-08-07
move some var declarations closer to their usages

ryan-williams:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/526)


timodonnell:
lgtm

PR #527, opened 2016-08-07
factor out SJC callPileups

ryan-williams:
cf. https://github.com/hammerlab/guacamole/pull/496/files#r73795569

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/527)


timodonnell:
lgtm

PR #529, opened 2016-08-07
name null-ish method params

ryan-williams:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/529)


timodonnell:
lgtm

PR #533, opened 2016-08-07
makePileup / PileupUtil helper

ryan-williams:
reduce test boilerplate around creating pileups

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/533)

### DIFF
--- a/src/test/scala/org/hammerlab/guacamole/commands/SomaticStandardCallerSuite.scala
+++ b/src/test/scala/org/hammerlab/guacamole/commands/SomaticStandardCallerSuite.scala
@@ -1,7 +1,7 @@
 package org.hammerlab.guacamole.commands
 
 import org.hammerlab.guacamole.filters.somatic.SomaticGenotypeFilter
-import org.hammerlab.guacamole.pileup.Pileup
+import org.hammerlab.guacamole.pileup.{Util => PileupUtil}
 import org.hammerlab.guacamole.reads.{MappedRead, ReadsUtil}
 import org.hammerlab.guacamole.reference.{ContigName, Locus, ReferenceBroadcast}
 import org.hammerlab.guacamole.util.{Bases, GuacFunSuite, TestUtil}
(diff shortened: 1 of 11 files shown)


timodonnell:
This overall seems more complicated to me than the code it replaces, especially the stuff involving `implicit`

ryan-williams:
A bunch of tests instantiate `Pileup`s to do various things. They are not intending to test anything related to `Pileup` construction; in particular, they always want to grab the `ContigSequence` for the `ContigName` that the `Pileup` is being created on. Moreover, they have an ambient `ReferenceBroadcast` laying around from which to get said `ContigSequence`.

Instrumenting test `Suite`s to reduce mindless boilerplate at the sub-test-case level (some tests make multiple pileups) reduces complexity overall; two possible ways of doing so are:

1. declare an abstract `def reference: ReferenceBroadcast` in `pileup.Util` that will be used in `Util.makePileup`, and implemented in each `Suite`.
2. have `makePileup` take the `ReferenceBroadcast` it needs `implicit`ly, and require `Suite`s to declare their ambient `ReferenceBroadcast`s to be `implicit`.

The former has two main drawbacks relative to the latter:
- it's not possible for a `Suite` to use one `ReferenceBroadcast` in one place and another elsewhere.
- `Suite`s must name their `ReferenceBroadcast`s the same thing (`reference` in this example).

The `implicit` route fixes both of these problems, and is equally as statically checked, which is to say basically completely: missing or extraneous `implicit ReferenceBroadcast`s will be compile-time errors.

So I disagree that this is more complex; it puts test-Pileup-creation in one place that is then reused many dozens of times, and a quick skim of the diff shows a lot of red: redundant code that has been eliminated. 

Lots of our tests are hard to follow for being walls of text due to poor decomposition of boilerplate setup code like this.

src/test/scala/org/hammerlab/guacamole/commands/SomaticStandardCallerSuite.scala
```diff
@@ -238,7 +242,7 @@ class SomaticStandardCallerSuite
         ("TCGTCGA", "3M1D4M", 0)
       )
 
-    val tumorPileup = Pileup(tumorReads, "chr1", 2, simpleReference.getContig("chr1"))
+    val tumorPileup = makePileup(tumorReads)
```
timodonnell: from looking at this it's not obvious to me what position the pileup will be at. Would be good to include those arguments in these places `"chr1"` and `2`

ryan-williams: sure, agreed this is a bridge too far, removing those default args

ryan-williams:
OK, I reworked things to make `reference` part of the `Util` mix-in. I also moved two other `TestUtil` methods involved in creating pileups into `pileup.Util`.

PR #534, opened 2016-08-07
Import some PFMU/WFMU methods directly in tests

ryan-williams:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/534)


timodonnell:
No objections

PR #535, opened 2016-08-07
use per-sample alias in PFMU test

ryan-williams:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/535)


timodonnell:
Seems overall an increase in complexity to me, but no strong opinions. Fine to merge if you really think it's better

ryan-williams:
cool, I hear that; fwiw the `PerSample` bit of this is orthogonal to the factoring out of `pileupsToElementStrings`, which I assume is the added complexity you're referring to; I think not repeating that inline function thrice is an improvement though, and worth factoring out, so yea I think I will go ahead with it.

PR #537, opened 2016-08-07
fold LikelihoodUtil into LikelihoodSuite

ryan-williams:
it's all just used in the one test and is not obviously going to be reused in other tests

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/537)


src/test/scala/org/hammerlab/guacamole/likelihood/LikelihoodSuite.scala
```diff
@@ -18,32 +18,74 @@
...
 import org.hammerlab.guacamole.reads.MappedRead
-import org.hammerlab.guacamole.util.{GuacFunSuite, TestUtil}
-import org.hammerlab.guacamole.variants.Genotype
+import org.hammerlab.guacamole.util.{Bases, GuacFunSuite, TestUtil}
+import org.hammerlab.guacamole.variants.{Allele, Genotype}
 import org.scalatest.prop.TableDrivenPropertyChecks
 
 class LikelihoodSuite extends GuacFunSuite with TableDrivenPropertyChecks {
 
+  implicit lazy val reference = TestUtil.makeReference(sc, Seq(("chr1", 1, "C")))
```
timodonnell: what does the implicit do here? I would rather avoid it if possible for clarity to non scala experts like me, unless it is doing something really useful (in which case a comment would help)

timodonnell:
seems fine to me

src/test/scala/org/hammerlab/guacamole/likelihood/LikelihoodSuite.scala
ryan-williams: ah, it is unused here, good catch!

it gets used by #533 but just snuck in here; anyway I'll document it over there as well

PR #539, opened 2016-08-07
Parends nits

ryan-williams:
rm a couple unnecessary parends, add some to calls that do non-trivial computation

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/539)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/GermlineAssemblyCaller.scala
@@ -153,7 +153,7 @@ object GermlineAssemblyCaller {
               )
 
             // Compute the number reads with variant bases from the reads overlapping the currentLocus
-            val pileupAltReads = (pileup.depth - pileup.referenceDepth)
+            val pileupAltReads = pileup.depth - pileup.referenceDepth
             if (currentLocusReads.isEmpty || pileupAltReads < minAltReads) {
               (lastCalledLocus, Iterator.empty)
             } else if (shortcutAssembly && !AssemblyUtils.isActiveRegion(currentLocusReads, referenceContig, minAreaVaf)) {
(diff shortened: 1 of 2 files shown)


timodonnell:
FWIW The style guide I read when I started writing scala (twitter's maybe?) said to use parens to indicate that a function has side effects. That's the convention I've been striving to follow. Not opposed to just merging this tho

ryan-williams:
Interesting! I guess I'm somewhat ambivalent between those; I really thought `SlidingWindow.currentRegions()` is worth parend-ing because it writes the whole queue to a vector each time it's called, which is one of the tightest-loop / critical-path candidates in the codebase, afaik; pushing the fact that it is not a `val` out to callers seems worthwhile given that.

Or, put another way, it has the potential "side effect" of a lot of runtime cost!

I'll wait and see if anyone else has thoughts on this.

arahuja:
Styleguide on parens: 
http://docs.scala-lang.org/style/naming-conventions.html#parentheses

> These are different methods at compile-time. While foo1 can be called with or without the parentheses, foo2 may not be called with parentheses.

> Thus, it is actually quite important that proper guidelines be observed regarding when it is appropriate to declare a method without parentheses and when it is not.

> Methods which act as accessors of any sort (either encapsulating a field or a logical property) should be declared without parentheses except if they have side effects.

ryan-williams:
Thx for the link. Seems like `nextLocusWithRegions` should be parend-less according to that, so I've changed that here.

I still feel like `currentRegions`' runtime implications are a serious side-effect and that the parends are reasonable/intuitive, but if I'm to infer that you are against that here then I'll drop them.

arahuja:
> but if I'm to infer that you are against that here then I'll drop them.

Indifferent on it really, so feel free to proceed as you see fit - as with most of these, my preference is to stick to a standard and automate how we check it/conform to it

ryan-williams:

PR #541, opened 2016-08-07
space out lots of joint caller code blocks

ryan-williams:
hopefully improving readability

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/541)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/commands/SomaticJointCaller.scala
+++ b/src/main/scala/org/hammerlab/guacamole/commands/SomaticJointCaller.scala
@@ -27,7 +27,8 @@ object SomaticJoint {
     @Args4jOption(name = "--out", usage = "Output path for all variants in VCF. Default: no output")
     var out: String = ""
 
-    @Args4jOption(name = "--out-dir",
+    @Args4jOption(
+      name = "--out-dir",
       usage = "Output dir for all variants, split into separate files for somatic/germline")
     var outDir: String = ""
 
@@ -283,34 +284,57 @@ object SomaticJoint {
 
       writeSome(path("all"), calls, inputs.items)
 
-      if (!onlySomatic) {
+      if (!onlySomatic)
         writeSome(
           path("germline"),
-          calls.filter(_.singleAlleleEvidences.exists(
-            evidence => evidence.isGermlineCall || anyForced(evidence))),
-          inputs.items)
-      }
+          calls.filter(
+            _.singleAlleleEvidences.exists(
+              evidence => evidence.isGermlineCall || anyForced(evidence)
+            )
+          ),
+          inputs.items
+        )
 
       val somaticCallsOrForced =
-        calls.filter(_.singleAlleleEvidences.exists(
-          evidence => evidence.isSomaticCall || anyForced(evidence)))
+        calls.filter(
+          _.singleAlleleEvidences.exists(
+            evidence => evidence.isSomaticCall || anyForced(evidence)
+          )
+        )
+
       writeSome(path("somatic.all_samples"), somaticCallsOrForced, inputs.items)
 
-      inputs.items.foreach(input => {
-        if (!onlySomatic) {
+      inputs.items.foreach(
+        input => {
+          if (!onlySomatic) {
+            writeSome(
+              path("all.%s.%s.%s".format(
+                input.sampleName, input.tissueType.toString, input.analyte.toString)),
+              calls,
+              Vector(input)
+            )
+          }
+
           writeSome(
-            path("all.%s.%s.%s".format(
-              input.sampleName, input.tissueType.toString, input.analyte.toString)),
-            calls,
-            Vector(input))
+            path(
+              "somatic.%s.%s.%s".format(
+                input.sampleName,
+                input.tissueType.toString,
+                input.analyte.toString
+              )
+            ),
+            somaticCallsOrForced,
+            Vector(input)
+          )
         }
-        writeSome(
-          path("somatic.%s.%s.%s".format(
-            input.sampleName, input.tissueType.toString, input.analyte.toString)),
-          somaticCallsOrForced,
-          Vector(input))
-      })
-      writeSome(path("all.tumor_pooled_dna"), somaticCallsOrForced, Vector.empty, includePooledTumor = Some(true))
+      )
+
+      writeSome(
+        path("all.tumor_pooled_dna"),
+        somaticCallsOrForced,
+        Vector.empty,
+        includePooledTumor = Some(true)
+      )
     }
   }
 }
--- a/src/main/scala/org/hammerlab/guacamole/jointcaller/AlleleAtLocus.scala
+++ b/src/main/scala/org/hammerlab/guacamole/jointcaller/AlleleAtLocus.scala
@@ -116,7 +116,14 @@ object AlleleAtLocus {
 
       subsequenceCounts.map(pair => (subsequenceToAllele(pair._1), requiredReads, pair._2))
     })
-    val result = alleleRequiredReadsActualReads.filter(tpl => tpl._3 >= tpl._2).map(_._1).distinct.toVector
+
+    val result =
+      alleleRequiredReadsActualReads
+        .filter(tpl => tpl._3 >= tpl._2)
+        .map(_._1)
+        .distinct  // Reduce to distinct alleles
+        .toVector
+
     if (atLeastOneAllele && result.isEmpty) {
       val allelesSortedByTotal = alleleRequiredReadsActualReads
         .groupBy(_._1)
--- a/src/main/scala/org/hammerlab/guacamole/jointcaller/VCFOutput.scala
+++ b/src/main/scala/org/hammerlab/guacamole/jointcaller/VCFOutput.scala
@@ -6,12 +6,13 @@ import htsjdk.samtools.SAMSequenceDictionary
 import htsjdk.variant.variantcontext._
 import htsjdk.variant.variantcontext.writer.VariantContextWriterBuilder
 import htsjdk.variant.vcf._
-import org.hammerlab.guacamole.jointcaller.Input.{ Analyte, TissueType }
-import org.hammerlab.guacamole.jointcaller.annotation.{ MultiSampleAnnotations, SingleSampleAnnotations }
+import org.hammerlab.guacamole.jointcaller.Input.{Analyte, TissueType}
+import org.hammerlab.guacamole.jointcaller.annotation.{MultiSampleAnnotations, SingleSampleAnnotations}
 import org.hammerlab.guacamole.jointcaller.evidence._
+import org.hammerlab.guacamole.jointcaller.pileup_summarization.PileupStats.AlleleMixture
 import org.hammerlab.guacamole.reference.ReferenceBroadcast
 
-import scala.collection.{ JavaConversions, mutable }
+import scala.collection.{JavaConversions, mutable}
 
 object VCFOutput {
   /**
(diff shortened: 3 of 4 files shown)


src/main/scala/org/hammerlab/guacamole/commands/jointcaller/VCFOutput.scala
```diff
@@ -114,100 +137,244 @@ object VCFOutput {
 
     def allelicDepthString(depths: Map[String, Int]): String = {
       val allAlleles = (Seq(allele.ref, allele.alt) ++ depths.keys.toSeq).distinct
-      allAlleles.map(x => "%s=%d".format(if (x.isEmpty) "." else x, depths.getOrElse(x, 0))).mkString(",")
+      allAlleles.map(
```
timodonnell: I find this increase in verbosity less clear actually. I'd much rather be able to see the code on one page then have to scroll around a ton. That's a personal preference though and I wouldn't reject new code that looks like this. I just don't think we should change existing code to be in this style, as it is not a meaningful improvement in my opinion

ryan-williams: Do you mean that specifically about the change at this line, or this whole PR?

I agree this line is a little more vertically stretched than it needs to be; will condense it back a bit.

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/VCFOutput.scala
```diff
@@ -114,100 +137,244 @@ object VCFOutput {
...
+
           val posteriors = samplesEvidence.perNormalSampleGermlinePosteriors(input.index)
+
           val alleleGenotype = posteriors.maxBy(_._2)._1
-          genotypeBuilder.alleles(JavaConversions.seqAsJavaList(
-            Seq(makeHtsjdkAllele(alleleGenotype._1), makeHtsjdkAllele(alleleGenotype._2))))
-          genotypeBuilder.attribute("RL",
-            posteriors.map(pair => "[%s/%s]=%.8g".format(pair._1._1, pair._1._2, pair._2)).mkString(","))
+
+          genotypeBuilder.alleles(
```
timodonnell: the `.alleles` call in this block seems harder to read to me than what it replaces. Denser code is often clearer to my eyes I guess. The `.attribute` call below seems like an improvement though since it now names the elements in the pair, so I'm down to make that replacement. Although, just fwiw, I often find scala's `for` harder to read than alternative ways of writing the same thing, so if there's a way to do this while keeping it a `map` that would be my preference

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/VCFOutput.scala
```diff
@@ -114,100 +137,244 @@ object VCFOutput {
...
           }
-          genotypeBuilder.alleles(JavaConversions.seqAsJavaList(sampleGenotype.map(makeHtsjdkAllele _)))
-          genotypeBuilder.attribute("RL",
-            posteriors.map(pair => "[%s]=%.8g".format(mixtureToString(pair._1), pair._2)).mkString(","))
+
+          genotypeBuilder.alleles(
+            JavaConversions.seqAsJavaList(
+              sampleGenotype.map(
+                makeHtsjdkAllele _
+              )
```
timodonnell: I'm not a fan of using lines for ending parens, would rather put them on together on the last line

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/VCFOutput.scala
```diff
@@ -114,100 +137,244 @@ object VCFOutput {
...
+              .map(allele => tumorEvidence.allelicDepths.getOrElse(allele, 0))
+              .toArray
+            )
             .DP(tumorEvidence.depth)
-        }
-        case (TissueType.Tumor, Analyte.RNA) => {
-          val tumorEvidence = samplesEvidence.allEvidences(input.index).asInstanceOf[TumorRNASingleSampleSingleAlleleEvidence]
+
+        case (TissueType.Tumor, Analyte.RNA) =>
+
```
timodonnell: why the newline?

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/evidence/MultiSampleSingleAlleleEvidence.scala
```diff
@@ -57,15 +63,19 @@ case class MultiSampleSingleAlleleEvidence(parameters: Parameters,
...
-    case 2                               => 0 // hom ref
-    case 1                               => parameters.germlineNegativeLog10HeterozygousPrior // het
-    case 0 if (alleles._1 == alleles._2) => parameters.germlineNegativeLog10HomozygousAlternatePrior // hom alt
-
-    // Compound alt, which we should not hit in the current version of the caller (we currently only consider mixtures
-    // involving a reference allele and a single alt allele)
-    case 0                               => throw new AssertionError("Compound alts not supported")
-  }
+  def germlinePrior(alleles: (String, String)): Double =
+    Seq(
```
timodonnell: Agree this is an improvement

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/evidence/MultiSampleSingleAlleleEvidence.scala
```diff
@@ -97,46 +125,62 @@ case class MultiSampleSingleAlleleEvidence(parameters: Parameters,
...
-    inputs.items.filter(_.tumorRNA).map(input => {
-      val likelihoods = allEvidences(input.index).asInstanceOf[TumorRNASingleSampleSingleAlleleEvidence].logLikelihoods
-      input.index -> likelihoods.map(kv => (kv._1 -> (kv._2 - somaticPriorRna(kv._1))))
-    }).toMap
+    inputs
+      .items
+      .filter(_.tumorRNA)
+      .map(
+        input => {
+
```
timodonnell: I would drop these newlines at the start of blocks

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/evidence/MultiSampleSingleAlleleEvidence.scala
```diff
@@ -97,46 +125,62 @@ case class MultiSampleSingleAlleleEvidence(parameters: Parameters,
...
   /** Maximum a posteriori somatic mixtures for each tumor sample. */
   val perTumorRnaSampleTopMixtures = perTumorRnaSampleSomaticPosteriors.mapValues(_.maxBy(_._2)._1)
 
   /** Indices of tumor rna samples with expression */
-  val tumorRnaSampleExpressed: PerSample[SampleId] = perTumorRnaSampleTopMixtures
-    .filter(pair => pair._2.keys.toSet != Set(germlineAlleles._1, germlineAlleles._2))
-    .keys.toVector
+  val tumorRnaSampleExpressed: PerSample[SampleId] =
+    perTumorRnaSampleTopMixtures
+      .filter(pair => pair._2.keys.toSet != Set(germlineAlleles._1, germlineAlleles._2))
```
timodonnell: Seems like a reasonable replacement to me. it seems like the right density to me. Having the `.filter` split up onto multiple lines would be too spacey, so I'm glad this replacement does not do that

timodonnell:
Made some comments. I am assuming none of this changes semantics. My main thought is cases where you have taken unnamed things like `pair._1` and given them names, are a big improvement, but cases where all that changed was spacing are more debatable, especially things involving parens on their own line or blank lines. I find it pretty useful to be able to look at a function in one screen full. I also am less into scala's `for` than you so I'd encourage you to not that change things to for when it is avoidable. Aside from comments I'm fine to merge this in.

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/VCFOutput.scala
ryan-williams: Echoing some of my motivation from a comment from #496:

> I like only having one thing happening on a given line; scala is so dense, and lines are so much wider than they are tall (in terms of pixels on screen), that things tend to be clearer the more lines they are split over, at least starting from where we are today.

The previous version of this line opens a bunch of `(` scopes and then closes 4 of them at once. One can do so much in Scala while passing params to a method (and indeed you have in this file and even on this line), each parameter-slot potentially being a whole code block that is arbitrarily complex, so it is really helpful to be explicit about which function-call's scope is opened and closed where in nested calls like this, is my feeling.

Parends in Scala end up grouping code in a more substantive way than we are used to, closer to what curly-braces do (in Scala and many other languages), which is why I like treating them more like we treat curly-braces; especially in code where we are not accumulating a lot of state in each expression (which is generally good!), curly braces often end up being superfluous while the parends actually carry all the weight of structuring the code.

For example, there are about 5 things I care about, as a reader, happening on this line:

```
Seq(makeHtsjdkAllele(alleleGenotype._1), makeHtsjdkAllele(alleleGenotype._2))
```

Start a `Seq`, make an allele, pass it the first element of the pair, make a second allele, pass it the second element, close the seq.

…all on a line that is about 30x as wide as it is tall, on my screen; that's just a bad way to lay out the information, imho.

> I often find scala's for harder to read than alternative ways of writing the same thing, so if there's a way to do this while keeping it a map that would be my preference

In cases with one vanilla `.map` and no destructuring, I agree with you, but any time you nest more than one map/flatMap:

```scala
foos.flatMap(foo => {
  bars.map(bar => {
    baz(bar)
  })
})
```

you incur much more cruft than:

```scala
for {
  foo <- foos
  bar <- bars
} yield
  baz(bar)
```

Similarly, when there is destructuring involved, as you outlined that you agree with about the case below, `for`s are a big win imho.

So that's basically where I come down: multiple nested `map`/`flatMap`s and/or ones involving destructuring are usually better as a `for`. That doesn't seem that far from what you seem to be saying.

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/VCFOutput.scala
ryan-williams: Per [my comment above](https://github.com/hammerlab/guacamole/pull/541/files#r73811376), closing a bunch of parends in a row is something I really dislike as a reader; just like we wouldn't end the last method in a class with a bunch of `}}}`s to save a few lines, I don't think we should do that with `)`s; each one is meaningfully changing the scope (and typically indent level) of the code.

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/VCFOutput.scala
ryan-williams: this 4-`case` `match` block is 150 lines long; I wanted to space out the division between `case`s here to make it easier to see.

Maybe later I'll factor it out to be less humongous, but I thought this was an improvement for now.

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/evidence/MultiSampleSingleAlleleEvidence.scala
ryan-williams: great, glad to hear it!

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/evidence/MultiSampleSingleAlleleEvidence.scala
ryan-williams: definitely less invested in this one than the one above; dropping it

src/main/scala/org/hammerlab/guacamole/commands/jointcaller/evidence/MultiSampleSingleAlleleEvidence.scala
ryan-williams: makes sense

ryan-williams:
Cool, I responded to all comments and changed a few things.

One thing I want to mention in response to:

> I find it pretty useful to be able to look at a function in one screen full.

is that I frequently use IntelliJ's code-block-collapsing capability to be able to view a file at higher and lower levels. In particular, ⌘⇧- to collapse everything then ⌘+ to expand just one level at a time is a great way to see just the top-level API of a class, and incrementally drill down:



In the last frame here I'm able to survey an almost 300-line method easily.

This is a good way to view a function on one screen as you've described, without us being constrained to pack logical expressions way further horizontally than is optimal for close-inspection use cases.

timodonnell:
Cool, thanks for the responses @ryan-williams . LGTM

PR #542, opened 2016-08-07
mv `jointcaller` pkg to top-level

ryan-williams:
joint caller itself stays in `commands` with the others, for now!

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/542)


timodonnell:
LGTM

PR #545, opened 2016-08-09
cache coverages

ryan-williams:
removed this during CR process previously but that was a mistake; partitioning debug stats end up performing this whole computation a second time

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/545)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/readsets/rdd/RegionRDD.scala
+++ b/src/main/scala/org/hammerlab/guacamole/readsets/rdd/RegionRDD.scala
@@ -9,6 +9,7 @@ import org.hammerlab.guacamole.readsets.iterator.{ContigCoverageIterator, Contig
 import org.hammerlab.guacamole.reference.{ContigName, NumLoci, Position, ReferenceRegion}
 import org.hammerlab.magic.rdd.RunLengthRDD._
 
+import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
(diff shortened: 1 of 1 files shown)


src/main/scala/org/hammerlab/guacamole/readsets/rdd/RegionRDD.scala
```diff
@@ -26,19 +27,23 @@ class RegionRDD[R <: ReferenceRegion: ClassTag](@transient rdd: RDD[R])
   def coverage(halfWindowSize: Int, loci: LociSet): RDD[(Position, Coverage)] =
     coverage(halfWindowSize, sc.broadcast(loci))
 
+  val _coverages = mutable.HashMap[(Int, LociSet), RDD[(Position, Coverage)]]()
```
timodonnell: `coveragesCache` or similar might be a better name

timodonnell: the cache can also be private, I'd imagine

timodonnell:
LGTM

timodonnell:
oops didn't mean to close that, LGTM

src/main/scala/org/hammerlab/guacamole/readsets/rdd/RegionRDD.scala
ryan-williams: good points, both done

PR #546, opened 2016-08-09
optionally error/warn on repeat contigs in coverage-depth computation

ryan-williams:
- `RegionRDD` → `CoverageRDD`, also removed `implicit` creation since it now takes in `requireRegionsGroupedByContig` bool

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/546)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/loci/map/LociMap.scala
+++ b/src/main/scala/org/hammerlab/guacamole/loci/map/LociMap.scala
@@ -18,13 +18,11 @@
 
 package org.hammerlab.guacamole.loci.map
 
-import java.io.{InputStream, OutputStream, PrintStream}
+import java.io.{OutputStream, PrintStream}
 
-import org.hammerlab.guacamole.loci.partitioning.LociPartitioner.PartitionIndex
 import org.hammerlab.guacamole.loci.set.{LociSet, Builder => LociSetBuilder}
 import org.hammerlab.guacamole.reference.{ContigName, ReferenceRegion}
 import org.hammerlab.guacamole.strings.TruncatedToString
-import org.hammerlab.magic.iterator.LinesIterator
 
 import scala.collection.immutable.TreeMap
 import scala.collection.{SortedMap, mutable}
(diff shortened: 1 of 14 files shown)


src/main/scala/org/hammerlab/guacamole/loci/map/LociMap.scala
```diff
@@ -99,6 +99,19 @@ object LociMap {
   }
 
   /**
+   * Build a map from some LociSets, mapping each set's loci to that sets index in `lociSets`.
+   */
+  def apply(lociSets: Iterable[LociSet]): LociMap[PartitionIndex] = {
```
timodonnell: Function looks fine, but might be a bit strange to have LociMap use stuff from `loci.partitioning.LociPartitioner`. Maybe just have it return a `LociMap[Int]`? Also this `apply` seems kind of niche to loci partitioning and not something that it's obvious from the signature what it does, so could be worth giving it a descriptive name

src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala
```diff
@@ -56,6 +62,13 @@ trait LociPartitionerArgs
     handler = classOf[BooleanOptionHandler]
   )
   var quiet: Boolean = false
+
+  @Args4JOption(
+    name = "--allow-ungrouped-contigs",
+    usage = "Loosen requirement that regions are grouped by contigs.",
```
timodonnell: Just out of curiosity, it's not obvious when / why you would or wouldn't use this. Is it just for debug? If you see a good place to document the higher level use case for this that could be useful

timodonnell:
Left some comments, feel free to address but fine to merge either way. LGTM

src/main/scala/org/hammerlab/guacamole/loci/map/LociMap.scala
ryan-williams: good point, `LociMap` should not know about `PartitionIndex`, but disguising it as an `Int` is not right either; the right solution here is to put this (and a couple other `LociMap` ctors that are specialized to the `LociMap[PartitionIndex]` a.k.a. `LociPartitioning` case into `LociPartitioning`, so I've done that, thanks for flagging this!

src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala
ryan-williams: yea, I thought it was simply less efficient but still correct on unsorted data, but digging back in, that is not true, so i've added verifications to the `ContigsIterator` and `ContigCoverageIterator` making sure these assumptions hold. thanks!

PR #547, opened 2016-08-09
Include spark application id in VCF metadata in joint caller

timodonnell:
Add the spark application ID to the metadata in the VCF in the joint caller. This way it can be included in benchmarking results so we can go back and get logs etc. for a benchmark

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/547)


ryan-williams:
lgtm

timodonnell:
Updated to hopefully fix the failure in SomaticJointCallerEndToEndSuite

PR #549, opened 2016-08-15
add load/save capability to PartitionedRegions

ryan-williams:
one step toward allowing the pileup-creation pipeline to save duplicate work during iteration on downstream logic, e.g. pileup-creation or variant-calling

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/549)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala
+++ b/src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala
@@ -3,6 +3,7 @@ package org.hammerlab.guacamole.loci.partitioning
 import org.apache.spark.rdd.RDD
 import org.hammerlab.guacamole.loci.LociArgs
 import org.hammerlab.guacamole.loci.set.LociSet
+import org.hammerlab.guacamole.readsets.rdd.PartitionedRegionsArgs
 import org.hammerlab.guacamole.reference.ReferenceRegion
 import org.hammerlab.magic.args4j.StringOptionHandler
 import org.kohsuke.args4j.spi.BooleanOptionHandler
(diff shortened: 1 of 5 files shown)


src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala
```diff
@@ -18,10 +18,12 @@ trait LociPartitionerArgs
...
   @Args4JOption(
     name = "--loci-partitioning",
-    usage = "If set: load a LociPartitioning from this path if it exists; write a computed partitioning to this path if it doesn't.",
+    usage = "Load a LociPartitioning from this path if it exists; else write a computed partitioning to this path.",
     handler = classOf[StringOptionHandler]
   )
-  var lociPartitioningPathOpt: Option[String] = None
+  protected var _lociPartitioningPathOpt: Option[String] = None
+
+  def lociPartitioningPathOpt: Option[String] = _lociPartitioningPathOpt
```
timodonnell: Might be worth a comment explaining what's going on here with the `def` and the `protected var`. I'm sure there's a good reason but it's not obvious to me.

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegionsArgs.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.kohsuke.args4j.{Option => Args4JOption}
+
+/**
+ * Command-line arguments related to partitioning regions according to a [[LociPartitioning]].
+ *
+ * Allows for configuring how/whether to persist the computed [[PartitionedRegions]] (its [[RDD]] of regions and/or its
+ * [[LociPartitioning]]).
+ */
+trait PartitionedRegionsArgs extends LociPartitionerArgs {
+  @Args4JOption(
```
timodonnell: Not obvious to me what's going with `--partitioning-dir` vs. `--partitioned-reads`.  A header comment explaining what the difference between these two options is would be useful

timodonnell:
Made some small comments, LGTM

src/main/scala/org/hammerlab/guacamole/readsets/rdd/PartitionedRegionsArgs.scala
ryan-williams: Cool added more comments and usage info.

src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala
ryan-williams: makes sense, added! the getter interface allows for different path-resolution logic in subclasses, e.g. PartitionedRegionsArgs

PR #550, opened 2016-08-15
clarify "quiet" param governing partitioning debug info

### LINKED ISSUE #544: minor thing: consider changing `--quiet` to indicate performance implications
```
  @Args4JOption(
    name = "--quiet",
    aliases = Array("-q"),
    usage = "Whether to compute additional statistics about the partitioned reads (default: false).",
  )
  var quiet: Boolean = false
```

If seems to be a substantial performance difference so may want to make the default not to do that computation and then have an arg like `--partitioning-debug-info` to turn it on. `--quiet` to me implies it just affects what gets printed not what gets computed.

ryan-williams:
fixes #544 

also removes a harmless joint-caller println's gating behind this param

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/550)


timodonnell:
LGTM

Issue #205, opened 2014-11-04
Add comments and tests for Filters package


timodonnell:
Some classes could use some more explanation:
- GenotypeFilter.scala - need a file-level header comment describing high-level idea
- PileupFilter.scala - need file-level header comment. These filters seem to be filtering PileupElements, not Pileups. What is the difference between this and PileupElementsFilter.scala?
- SomaticGenotypeFilter.scala - need file-level header comment
- FishersExactTest - comment needed. Maybe move to a math utils file?

PR #554, opened 2016-08-15
filters cleanup

ryan-williams:
related to #205, doesn't fix it but makes some progress

- move each filter to its own file
- change multi-class files to packages that contain multiple files
- simplify math expressions in FishersExact

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/554)


src/main/scala/org/hammerlab/guacamole/filters/FishersExactTest.scala
```diff
@@ -18,22 +18,20 @@
...
   /** Fisher's exact test, returned as -1 * log base 10 probability (i.e. a positive number). */
-  def asLog10(totalA: Int, totalB: Int, conditionA: Int, conditionB: Int): Double = {
-    (ArithmeticUtils.binomialCoefficientLog(totalA, conditionA) +
-      ArithmeticUtils.binomialCoefficientLog(totalB, conditionB) -
-      ArithmeticUtils.binomialCoefficientLog(totalA + totalB, conditionA + conditionB)
-      / Math.log(10))
-  }
-}
\ No newline at end of file
+  def asLog10(totalA: Int, totalB: Int, conditionA: Int, conditionB: Int): Double =
```
timodonnell: could use a header comment

timodonnell:
LGTM

src/main/scala/org/hammerlab/guacamole/filters/FishersExactTest.scala
ryan-williams: done!

PR #555, opened 2016-08-19
split more things out of TestUtil

ryan-williams:
- reference things to `ReferenceUtil`,
- reads-rdd things to `ReadsRDDUtil`, 
- factor temp-file utilities to magic-rdds helper that cleans up afterward
- also rm unused `LikelihoodUtil`

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/555)


src/test/scala/org/hammerlab/guacamole/commands/SomaticStandardCallerRealDataSuite.scala
```diff
@@ -117,21 +118,21 @@ class SomaticStandardCallerRealDataSuite
...
         sc,
         "synthetic.challenge.set1.tumor.v2.withMDTags.chr2.syn1fp.sam",
         "synthetic.challenge.set1.normal.v2.withMDTags.chr2.syn1fp.sam"
       )
 
     val negativePositions = Array[Long](
       216094721,
-      3529313,
-      8789794,
+        3529313,
```
timodonnell: spacing change intentional?

src/test/scala/org/hammerlab/guacamole/commands/SomaticStandardCallerRealDataSuite.scala
```diff
@@ -149,8 +150,8 @@ class SomaticStandardCallerRealDataSuite
     val negativePositions = Array[Long](
       148487667,
       134307261,
-      90376213,
-      3638733,
+       90376213,
+        3638733,
```
timodonnell: intentional?

timodonnell:
Did a fast pass no obvious issues. From an organization standpoint I think it'd be clearest to have all utilities intended only to be used by tests to reside together (like in their own sub package) or have the word Test in their name, but I'm fine with this approach too. LGTM

src/test/scala/org/hammerlab/guacamole/commands/SomaticStandardCallerRealDataSuite.scala
ryan-williams: yea i right-aligned a few large lists of loci integers like this is a few places, thought it was clearer

src/test/scala/org/hammerlab/guacamole/commands/SomaticStandardCallerRealDataSuite.scala
ryan-williams: ditto

ryan-williams:
Thanks for the comments.

> From an organization standpoint I think it'd be clearest to have all utilities intended only to be used by tests to reside together (like in their own sub package)

Yea that's an interesting consideration; there's "used by tests" as in "… not by app/main", and also "… as opposed to being tests themselves".

The former is already distinguished at the highest level, by the `src/{main,test}` divide.

The latter is analogous to the discussion about whether, in `src/main`, we should have packages like `commands` separate from [all the code that services those commands]. That state of affairs is only bearable in `src/main` since there are relatively few commands, and even so it's a bit of a smell there: imagining the per-package dependency graph, something like `commands` ends up depending on everything, so the assembly caller can't be built/run independently of the joint caller, though that's a more common and logical distinction in the codebase than "build/run all the callers" vs. "build/run all the code that services all the callers". 

In `src/test`, where the "commands" (the tests themselves) are myriad and there is even more impetus to build/run them independently of one another, it's correspondingly better to elevate the "reads" vs. "pileups" -type distinctions over the "runnable cmds" vs. "everything else" distinctions, so that's why I did it this way and think that that's the right approach.

> … or have the word Test in their name

Being is `src/test` should be sufficient to make it clear that they are scoped for tests only. Within `src/test` our current convention is `*Suite` for tests themselves and `*Util` for utilities servicing those tests. If either was going to be changed to `*Test` it should be the former, not the latter, so I am going to leave it as is.

PR #556, opened 2016-08-19
take sample-id awareness out of read-partitioning code

ryan-williams:
better separation of concerns:
- put a "sample ID" back on read objects
- partition reads without worrying about sample ID or "keying" reads with it
- make downstream code (e.g. SplitIterator) not require `Int` keys for splitting.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/556)


src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
```diff
@@ -236,7 +237,8 @@ object VAFHistogram {
    * @param printStats Print descriptive statistics for the variant allele frequency distribution
    * @return RDD of VariantLocus, which contain the locus and non-zero variant allele frequency
    */
-  def variantLociFromReads(partitionedReads: PartitionedReads,
+  def variantLociFromReads(numSamples: NumSamples,
```
timodonnell: should add `numSamples` to header comment

src/main/scala/org/hammerlab/guacamole/distributed/WindowFlatMapUtils.scala
```diff
@@ -120,14 +127,16 @@ object WindowFlatMapUtils {
    *
    *  (3) The results of the provided function are concatenated into an RDD, which is returned.
    *
+   * @param numSamples number of samples / input-files whose reads are in @partitionedReads.
```
timodonnell: is the header comment still correct about what this function does?

src/main/scala/org/hammerlab/guacamole/windowing/SplitIterator.scala
```diff
@@ -13,9 +13,9 @@ object SplitIterator {
    * @tparam V
    * @return a buffered iterator for each key in 0 .. num.
    */
-  def split[V](num: Int, iterator: Iterator[(Int, V)]): Vector[BufferedIterator[V]] = {
+  def split[V](num: Int, iterator: Iterator[V], idxFn: V => Int): Vector[BufferedIterator[V]] = {
```
timodonnell: need to update header comment about the new parameter (and `V`)

timodonnell:
I took a high-level scan and did not spot any obvious correctness issues, so LGTM

src/main/scala/org/hammerlab/guacamole/commands/VAFHistogram.scala
ryan-williams: good call, done

src/main/scala/org/hammerlab/guacamole/distributed/WindowFlatMapUtils.scala
ryan-williams: hm no it is pretty out of date, updated, thanks!

PR #564, opened 2016-08-22
travis scoverage attempt

### LINKED ISSUE #244: add coveralls support
@danvk says:

Just chiming in from the sidelines here: You'll probably never look at your coverage data unless you post it to coveralls.io from Travis and stick a badge in the project's README.md. Here's a sample PR from igv-httpfs that does this.

ryan-williams:
fixes #244 

This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/564)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,22 @@
 language: java
 sudo: false
+
 cache:
   directories:
   - .autoconf
   - $HOME/.m2
+  - $HOME/.zinc
+
+script: mvn scoverage:report
+
 after_success:
-- .travis/publish_docs.sh
+  - .travis/publish_docs.sh
+  - mvn coveralls:report
+
 env:
   global:
     secure: ffCremJXhnrGEaP0nRgChQiaZrfDJpG8yyMckri/dcTYZsESa0qcyyu9gNXwtW3UkEZpmwFbMSUqkMHgIsYBMDMD9zZLHUpMLswN6VDl3/WrE+4o7tJq8PGDB/idPnM+vr4YPO8uopUGGIhEiAHsvBGqybYsHdGQa1TKmbH0x+k=
+
 notifications:
   slack:
     secure: MpSRYjcajUOegCpk5dzpG0lCe5YBwzlVYTo4b0MD3gu7HFHGFscf2CW1DlP3QJTdaTjxU2oA9nEOvrxq1RZ9J91hwDV5PtZXY5CdVRl9B6VrfulSGT35qeZoaiL0ETfNhfklPTEEiE7xSRG0sYsisqlCl5xN0QqBqrE0yzVHsRA=
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,17 +4,25 @@ sudo: false
 cache:
   directories:
   - .autoconf
+
+  # Local Maven repository.
   - $HOME/.m2
+
+  # Local builds of zinc (Scala incremental compiler).
   - $HOME/.zinc
 
+# Run the tests and generate coverage reports in one go.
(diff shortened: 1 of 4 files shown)


e5c:
LGTM!

---

Reviewed 3 of 3 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/hammerlab/guacamole/564#-:-KPxPDTNOQTCxdkNwlE7:b-85i648)*

.travis.yml
```diff
@@ -1,14 +1,22 @@
 language: java
 sudo: false
+
 cache:
   directories:
   - .autoconf
   - $HOME/.m2
+  - $HOME/.zinc
+
+script: mvn scoverage:report
```
arahuja: Not sure if this intended or not and it doesn't seem like it makes a difference, but travis now seems to run the tests as a side-effect of the scoverage run rather than running the tests explicitly?

before travis ran:
```
mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
mvn test -B
```

now it runs

```
mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
mvn scoverage:report
```

ryan-williams: Yea, that's intentional, though maybe there's a clearer way to do it, and/or I can document things better. Any ideas?

Maybe I'll try leaving `script` as the default `mvn test -B`, and doing:

```yml
after_success:
- mvn scoverage:report -DskipTests coveralls:report
```

or something like that.

By default `mvn scoverage:report` runs the tests.

arahuja:
Did you have to add something locally to run this?

I tried 
```
mvn clean
mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
mvn scoverage:report
```
I get 

```
[ERROR] No plugin found for prefix 'scoverage' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/Users/arahuja/.m2/repository), central (https://repo.maven.apache.org/maven2)] -> [Help 1]
```

ryan-williams:
You need `-Pcoverage` for `scoverage:report` or `coveralls:report` to be availble.

Also `coveralls:report` will need coveralls `<repoToken/>`, which we're not supposed to put into the git history, but you can get from https://coveralls.io/github/hammerlab/guacamole (I'm assuming only hammerlab org can see the token on that page…).

arahuja:
> You need -Pcoverage for scoverage:report

Ah interesting, any reason the travis script doesn't need that?

ryan-williams:
Yea, sry I should have mentioned that `-Pcoverage` [is also activated by `env.TRAVIS`](https://github.com/hammerlab/guacamole/pull/564/files#diff-600376dffeb79835ede4a0b285078036R406); that predates this change (though that whole profile has been defunct for a long time).

.travis.yml
ryan-williams: OK, I played with this locally and can only get a non-empty coverage report when the tests are run as part of `scoverage:report`.

Things that didn't work:

```
mvn clean -Pcoverage test  # also without -Pcoverage
mvn -Pcoverage -DskipTests scoverage:report
mvn -Pcoverage coveralls:report
```

So I think we should leave it like I have it here; I've added comments to `.travis.yml` about this state of affairs though.

arahuja:
@ryan-williams All set with this?

ryan-williams:
I believe so! Sorry, didn't see a "lgtm" or equivalent so wasn't sure if you wanted another look

PR #567, opened 2016-09-02
Revert somatic standard tests to use table properties

arahuja:
This is reverting part of https://github.com/hammerlab/guacamole/commit/8c471d5545b17b67daca660ea4f06950b4fb1a3d to use the table property check, otherwise the failing locus is not printed and the message just states `false was not true`

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/567)


ryan-williams:
lgtm, thanks!

PR #569, opened 2016-09-06
Upgrade Hadoop BAM for better interval filtering

### LINKED ISSUE #399: Upgrade HadoopBAM to use BAM index

arahuja:
This brings back #433 and fixes #399 

@ryan-williams I think you tested switching the Java version on a single Spark submission, can we add something to the README on that? This PR will move the repository to Java 8 only

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/569)


ryan-williams:
OK, sorry for the delay here @arahuja, this lgtm.

I just sent #578 that resolves the conflicts here and adds a couple of things.

I am agnostic about whether you incorporate any/all of it or just rebase this and merge it in yourself.

Lmk if you have any questions about what I'd proposed over on #578.

src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
```diff
@@ -321,14 +320,15 @@ object ReadSets extends Logging {
...
     var result = reads
-    for {
-      overlapsLoci <- filters.overlapsLociOpt
-    } {
-      val contigLengths = getContigLengths(sequenceDictionary)
-      val loci = overlapsLoci.result(contigLengths)
-      val broadcastLoci = reads.sparkContext.broadcast(loci)
-      result = result.filter(_.asMappedRead.exists(broadcastLoci.value.intersects))
-    }
+    filters
```
ryan-williams: I find the `for` to be cleaner, do you not?

arahuja: No - not particularly?

arahuja:
I re-based this PR, and then added your commit on top. 

I added one more commit to change a matching on `Option` to be a `fold` instead as I think matching on `Option` is not recommended.

I similarly changed a `for` on option to `foreach`.

ryan-williams:
Sounds good.

> matching on Option is not recommended.

Interesting; any reason?

arahuja:
>Interesting; any reason? 

Just boilerplate already built-in to `map`, `fold` or `foreach`

PR #568, opened 2016-09-04
Move likelihood code to breeze

arahuja:
Reviving an old branch here that moved the likelihood calculations to `breeze`

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/568)

PR #571, opened 2016-09-08
Update likelihood code to handle non-uniform allelic mixtures

arahuja:
This PR contains the following changes:
- Misc updates from the code review on #568 
- Changes `Genotype` from a collection of alleles to a map of `Allele` -> Allele Fraction
- Updates `Likelihood.scala` and `LikelihoodSuite.scala` to handle this

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/571)

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/variants/Genotype.scala
+++ b/src/main/scala/org/hammerlab/guacamole/variants/Genotype.scala
@@ -9,7 +9,8 @@ package org.hammerlab.guacamole.variants
  */
 case class Genotype(alleleMixture: Map[Allele, Double]) {
 
-  assume(alleleMixture.values.sum <= 1, "Allele fractions are larger than 1")
+  // The allele fractions should sum up to approximately one
+  assume(alleleMixture.values.sum == 1, s"Allele should sum to 1, but sum to ${alleleMixture.values.sum}")
 
   val alleles = alleleMixture.keySet
 
--- a/src/main/scala/org/hammerlab/guacamole/variants/Genotype.scala
+++ b/src/main/scala/org/hammerlab/guacamole/variants/Genotype.scala
@@ -1,28 +1,19 @@
 package org.hammerlab.guacamole.variants
 
-import com.esotericsoftware.kryo.io.{Input, Output}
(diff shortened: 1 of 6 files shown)


src/main/scala/org/hammerlab/guacamole/variants/Genotype.scala
```diff
@@ -1,28 +1,19 @@
...
  */
-case class Genotype(alleles: Allele*) {
-  /**
-   * The ploidy of the organism is the number of alleles in the genotype.
-   */
-  val ploidy = alleles.size
+case class Genotype(alleleMixture: Map[Allele, Double]) {
 
-  lazy val uniqueAllelesCount = alleles.toSet.size
+  assume(alleleMixture.values.sum <= 1, "Allele fractions are larger than 1")
```
ryan-williams: Will `alleleMixture` contain a reference allele so that we'd expect the sum to be ≈1?

arahuja: Yes, this test was originally `= 1` but was thinking it might be more likely that we assume `.5` for the reference and then something like `.2` for the variant if it is a low purity sample. Could possibly come back and adjust this.

ryan-williams: gotcha… I feel like the existence of this runtime-check raises more questions than it answers; is this a failure mode we're worried about? I'm ok leaving it in but maybe add a comment explaining the thinking, it's a peculiar and somewhat unnerving thing to feel the need to assert!

ryan-williams:
lgtm modulo adding a comment, and also some tweaks I sent in #574, which you can incorporate here or which I can pull out as a separate PR, uty!

src/main/scala/org/hammerlab/guacamole/variants/Genotype.scala
arahuja: Hm ok, maybe to clarify - the mixture should be a true mixture where the weights sum to 1. The reason for the check is to ensure that the entered values are at least fractions ... will add some comments around this

PR #574, opened 2016-09-09
likelihood nits

ryan-williams:
goes on top of #571 

- private-ize a few methods
- simplify alignment-inclusion API in likelihood calculation: simple
  boolean instead of arbitrary function
- remove "Likelihood." from call-sites
- misc line-spacing

I can make this a separate PR if you don't want to combine it with #571 / things here are controversial, @arahuja, lmk!

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/574)


arahuja:
Thanks!

PR #576, opened 2016-09-14
add somatic-standard end-to-end test

ryan-williams:
Individual commits stand alone, but a couple of additional notes:

- fold common genotype/variant logic into `GenotypeOutputCaller`, implemented by SomaticStandard and GermlineAssembly
- add hook in `SparkCommand` for validating arguments
- remove superfluous inheritance of "args" types from e.g. `DebugLogHelpers`
- remove tumor/normal API from ReadSets, per discussion on #575

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/576)


ryan-williams:
Hm sort of lame that coveralls says "unknown" here, as this PR increases coverage ~3% (previously 76%), which is one of its selling points…

src/main/scala/org/hammerlab/guacamole/variants/Concordance.scala
```diff
@@ -1,154 +0,0 @@
-package org.hammerlab.guacamole.variants
```
arahuja: Any reason for the removing this? I was hoping to use this again.

arahuja:
LGTM - would prefer it if we kept `Concordance`

src/main/scala/org/hammerlab/guacamole/variants/Concordance.scala
ryan-williams: opportunistically removed since I didn't think it'd been used in years, has no tests, etc… I've put it back in though

PR #577, opened 2016-09-14
Move commons-io back to standard dep

arahuja:
At some point, `commons-io` was moved to a test-only dependency, but this is necessary as a dependency when running without `spark-submit` for Hadoop output.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/577)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -320,6 +320,11 @@
       <artifactId>grizzled-slf4j_${scala.version.prefix}</artifactId>
       <version>1.0.3</version>
     </dependency>
+    <dependency>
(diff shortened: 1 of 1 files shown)


ryan-williams:
heh, just did this locally as well.

Seems like it being present as a test dep was enough to omit it from being picked up as a trans-dep by all dependencies (including Spark, which needs it), but then it doesn't make it into the packaged JAR because it's only a test dep.

Feels like broken decisions in the Maven stack, but yea lgtm

PR #579, opened 2016-09-15
Scripts overhaul

ryan-williams:
Paraphrasing the commit msg from https://github.com/hammerlab/guacamole/commit/52666e1774b580fb0c8fd94f2dfa59faed2f3b06:

- make scripts/guacamole wrap spark-submit
- make scripts/guacamole-shell, a wrapper for spark-shell
- both delegate to jar-finder and spark-env-setup scripts, `.jars-rc` and `.spark-rc`.
- orient spark configuration around "spark properties" files, pointed to by `$GUAC_SPARK_CONFS` env var, which can take multiple paths to properties files which will be concatenated

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/579)


pom.xml
```diff
@@ -320,6 +320,11 @@
       <artifactId>grizzled-slf4j_${scala.version.prefix}</artifactId>
       <version>1.0.3</version>
     </dependency>
+
+    <!--
+      Only used directly in tests, but adding test-scope blocks transitive inclusion via Spark, resulting in runtime
```
arahuja: Not sure I get this comment - isn't this necessary to write a VCF when running without the `spark-assembly` jar?

arahuja:
LGTM

pom.xml
ryan-williams: I thought we just un-`test`-scoped this on #577, not because we need it in non-`test`-scope, but because Maven was excluding it as a transitive dep (which Spark and other libraries need) due to its presence in our POM, but then not including it at all in the assembly JAR since it was scoped `test`.

This comment just attempts to explain that state of affairs. lmk if it would be more clear to say something different, or if my understanding of the situation is incorrect.

arahuja: AFAIK, the reason we need it in test or non-test is the same - to use Hadoop outputs?

ryan-williams: My understanding was that something under `src/test` (and nothing under `src/main`) imported something from this dep, which is why we had it `test`-scoped, in our pom.

We were forced to widen the scope due to Maven excluding it from being packaged as a trans-dep from other deps of ours, like Spark.

After digging just now, I find that nothing in our code-base uses this dep, so I've removed it; I don't know why it was there originally.

arahuja: `commons-io` is needed for Hadoop output when running locally and not using `spark-submit`

ryan-williams: If Hadoop needs commons-io, then it should depend on it; lo and behold, it does:

```
$ mvn dependency:tree -Dverbose -Dincludes="commons-io:" | perl -pe 's/^\[INFO\] //'
…
org.hammerlab.guacamole:guacamole:jar:0.0.1-SNAPSHOT
+- org.bdgenomics.adam:adam-core_2.10:jar:0.19.0:compile
|  \- (commons-io:commons-io:jar:1.3.2:compile - omitted for conflict with 2.5)
+- org.apache.spark:spark-core_2.10:jar:1.6.1:compile
|  \- org.tachyonproject:tachyon-client:jar:0.8.2:compile
|     \- (commons-io:commons-io:jar:2.4:compile - omitted for conflict with 1.3.2)
+- org.apache.hadoop:hadoop-client:jar:2.7.0:compile
|  +- org.apache.hadoop:hadoop-common:jar:2.7.0:compile
|  |  \- (commons-io:commons-io:jar:2.4:compile - omitted for conflict with 1.3.2)
|  +- org.apache.hadoop:hadoop-hdfs:jar:2.7.0:compile
|  |  \- (commons-io:commons-io:jar:2.4:compile - omitted for conflict with 1.3.2)
|  \- org.apache.hadoop:hadoop-mapreduce-client-core:jar:2.7.0:compile
|     \- org.apache.hadoop:hadoop-yarn-common:jar:2.7.0:compile
|        \- (commons-io:commons-io:jar:2.4:compile - omitted for conflict with 1.3.2)
+- commons-io:commons-io:jar:2.5:compile
\- com.holdenkarau:spark-testing-base_2.10:jar:1.6.1_0.4.4:test (scope not updated to compile)
   +- org.apache.hadoop:hadoop-hdfs:jar:tests:2.6.0:test
   |  \- (commons-io:commons-io:jar:2.4:test - omitted for conflict with 2.5)
   +- org.apache.hadoop:hadoop-common:jar:tests:2.6.0:test
   |  \- (commons-io:commons-io:jar:2.4:test - omitted for conflict with 2.5)
   +- org.apache.hadoop:hadoop-yarn-server-tests:jar:tests:2.6.0:test
   |  \- org.apache.hadoop:hadoop-yarn-server-resourcemanager:jar:2.6.0:test
   |     \- (commons-io:commons-io:jar:2.4:test - omitted for conflict with 2.5)
   \- org.apache.hadoop:hadoop-minicluster:jar:2.6.0:test
      +- org.apache.hadoop:hadoop-common:test-jar:tests:2.6.0:test
      |  \- (commons-io:commons-io:jar:2.4:test - omitted for conflict with 2.5)
      \- org.apache.hadoop:hadoop-hdfs:test-jar:tests:2.6.0:test
         \- (commons-io:commons-io:jar:2.4:test - omitted for conflict with 2.5)
…
```

(This is our dep tree before this PR).

Our top-level dep on commons-io:2.5 was apparently a kludge to resolve the conflict between hadoop and spark's commons-io:2.4 dep with adam-core's commons-io:1.3.2 dep.

The right way to resolve that is to exclude adam-core from bringing in its older version and competing with spark and hadoop; afaict ADAM runs fine with 2.4, but hadoop and spark miss `Charsets` (added in 2.3) when 1.3.2 wins.

I've now done that:

```
org.hammerlab.guacamole:guacamole:jar:0.0.1-SNAPSHOT
+- org.apache.spark:spark-core_2.10:jar:1.6.1:compile
|  \- org.tachyonproject:tachyon-client:jar:0.8.2:compile
|     \- commons-io:commons-io:jar:2.4:compile
+- org.apache.hadoop:hadoop-client:jar:2.7.0:compile
|  +- org.apache.hadoop:hadoop-common:jar:2.7.0:compile
|  |  \- (commons-io:commons-io:jar:2.4:compile - omitted for duplicate)
|  +- org.apache.hadoop:hadoop-hdfs:jar:2.7.0:compile
|  |  \- (commons-io:commons-io:jar:2.4:compile - omitted for duplicate)
|  \- org.apache.hadoop:hadoop-mapreduce-client-core:jar:2.7.0:compile
|     \- org.apache.hadoop:hadoop-yarn-common:jar:2.7.0:compile
|        \- (commons-io:commons-io:jar:2.4:compile - omitted for duplicate)
\- com.holdenkarau:spark-testing-base_2.10:jar:1.6.1_0.4.4:test (scope not updated to compile)
   +- org.apache.hadoop:hadoop-hdfs:jar:tests:2.6.0:test
   |  \- (commons-io:commons-io:jar:2.4:test - omitted for duplicate)
   +- org.apache.hadoop:hadoop-common:jar:tests:2.6.0:test
   |  \- (commons-io:commons-io:jar:2.4:test - omitted for duplicate)
   +- org.apache.hadoop:hadoop-yarn-server-tests:jar:tests:2.6.0:test
   |  \- org.apache.hadoop:hadoop-yarn-server-resourcemanager:jar:2.6.0:test
   |     \- (commons-io:commons-io:jar:2.4:test - omitted for duplicate)
   \- org.apache.hadoop:hadoop-minicluster:jar:2.6.0:test
      +- org.apache.hadoop:hadoop-common:test-jar:tests:2.6.0:test
      |  \- (commons-io:commons-io:jar:2.4:test - omitted for duplicate)
      \- org.apache.hadoop:hadoop-hdfs:test-jar:tests:2.6.0:test
         \- (commons-io:commons-io:jar:2.4:test - omitted for duplicate)
```

ryan-williams:
Added in conf/kryo file and made it always get mixed in to the spark properties file.

pom.xml
ryan-williams: As a nice post-script to this discussion, the more recent ADAM snapshot we moved to in #581 depends on commons-io:2.4, so the exclusion is now unnecessary, and I've removed it.

PR #580, opened 2016-09-15
rm "-fasta" from reference-param name

ryan-williams:
- also rm unused ReferenceArgs

it can be a `.fasta.gz`, making the previous name a bit ambiguous, and both previous ref args were unnecessarily verbose

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/580)


arahuja:
This PR seems like it was merged in already? The diff is empty except for an empty line removed?

ryan-williams:
Weird, yea [this commit](https://github.com/hammerlab/guacamole/pull/579/commits/55b27e3df952c148742b0f2bc0fdc0598c6ce72c) ended up on #579; I don't think that was intentional / know when it was added there relative to your review, sorry!

Let me know if you have any comments on it!

PR #581, opened 2016-09-15
"upgrade" adam to forked "0.20.0"

ryan-williams:
this ports us on to most of the non-trivial changes that have happened in ADAM since 0.19.0, using my forked "0.20.0".

ADAM's VCF-writing now requires [`Sample.sampleId`](https://github.com/bigdatagenomics/bdg-formats/blob/bdg-formats-0.9.0/src/main/resources/avro/bdg.avdl#L1195) to match [`Genotype.sampleId`](https://github.com/bigdatagenomics/bdg-formats/blob/bdg-formats-0.9.0/src/main/resources/avro/bdg.avdl#L627), and I am using the "sampleName" pulled from [readsets.args.Base](https://github.com/hammerlab/guacamole/blob/a6771eeafed073e6858979e01b214b214d965a1f/src/main/scala/org/hammerlab/guacamole/readsets/args/Base.scala#L13) / [ReadSets](https://github.com/hammerlab/guacamole/blob/a6771eeafed073e6858979e01b214b214d965a1f/src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala#L35) rather than [the loaded-from-data `sampleName` on `Read`](https://github.com/hammerlab/guacamole/blob/a6771eeafed073e6858979e01b214b214d965a1f/src/main/scala/org/hammerlab/guacamole/reads/Read.scala#L32).

The latter is now unused so I removed it. It might make sense to keep it around explicitly in some way… seems like what ADAM now does with RecordGroups is the right thing, maybe we'll want to use that if we care about that data again.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/guacamole/581)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -155,7 +155,7 @@
           <reportsDirectory>${project.build.directory}/scalatest-reports</reportsDirectory>
           <junitxml>.</junitxml>
           <filereports>GuacamoleTestSuite.txt</filereports>
-          <argLine>-Xmx4g -XX:MaxPermSize=256m</argLine>
+          <argLine>-Xmx2g -XX:MaxPermSize=256m</argLine>
 
           <!-- Print full stack traces when tests fail. -->
           <stdout>F</stdout>
--- a/pom.xml
+++ b/pom.xml
@@ -16,16 +16,11 @@
   <name>guacamole: variant caller</name>
 
   <properties>
-    <adam.version>0.19.0</adam.version>
-    <bdg-formats.version>0.7.0</bdg-formats.version>
     <java.version>1.8</java.version>
     <scala.version>2.10.5</scala.version>
     <scala.version.prefix>2.10</scala.version.prefix>
     <hadoop.version>2.7.0</hadoop.version>
     <spark.version>1.6.1</spark.version>
-    <maven.release.plugin.version>2.4.1</maven.release.plugin.version>
-    <shade.plugin.version>2.1</shade.plugin.version>
-    <scoverage.plugin.version>1.1.1</scoverage.plugin.version>
   </properties>
 
   <licenses>
@@ -50,7 +45,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
-        <version>${shade.plugin.version}</version>
+        <version>2.1</version>
         <configuration>
           <transformers>
             <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
@@ -177,7 +172,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-release-plugin</artifactId>
-        <version>${maven.release.plugin.version}</version>
+        <version>2.4.1</version>
         <configuration>
           <branchName>branch-@{project.version}</branchName>
           <goals>deploy</goals>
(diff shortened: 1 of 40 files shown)


src/main/scala/org/hammerlab/guacamole/variants/CalledSomaticAllele.scala
```diff
@@ -41,6 +41,9 @@ case class CalledSomaticAllele(sampleName: SampleName,
       .newBuilder
       .setAlleles(seqAsJavaList(Seq(GenotypeAllele.Ref, GenotypeAllele.Alt)))
       .setSampleId(sampleName)
+      .setContigName(contigName)
```
arahuja: Weird, what was happening before?

pom.xml
```diff
@@ -16,16 +16,11 @@
   <name>guacamole: variant caller</name>
 
   <properties>
-    <adam.version>0.19.0</adam.version>
```
arahuja: Any reason you don't like these as properties?

src/test/resources/tough.golden.vcf
```diff
@@ -16,7 +16,92 @@
 ##INFO=<ID=ReadPosRankSum,Number=1,Type=Float,Description="Z-score from Wilcoxon rank sum test of Alt vs. Ref read position bias">
 ##INFO=<ID=VQSLOD,Number=1,Type=Float,Description="Log odds ratio of being a true variant versus being false under the trained gaussian mixture model">
 ##INFO=<ID=culprit,Number=1,Type=String,Description="The annotation which was the worst performing in the Gaussian mixture model, likely the reason why the variant was filtered out">
-#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	tumor.chr20
+##contig=<ID=1,length=249250621,assembly=GRCh37>
```
arahuja: Is this just what is generated now? Otherwise, don't really need all the contigs listed?

pom.xml
ryan-williams: I'm going with the heuristic that if it's used twice, it should be factored out, otherwise it should just be on the dependency-declaration.

Factoring out versions that only appear once seems more verbose and less clear, to me.

lmk if you disagree.

src/test/resources/tough.golden.vcf
ryan-williams: yea, it's what is generated now. 

if there was a more narrow a way to test this i'd be open to it, but this seems like the most straightforward way to have an end-to-end test.

src/main/scala/org/hammerlab/guacamole/variants/CalledSomaticAllele.scala
ryan-williams: i don't know, otomh!

the semantics of the downstream adam and htsjdk models here have changed in this upgrade, so a few things have to specified differently than they were before.

it's also possible that these aren't needed now either; I added them while debugging an issue where variant's INFO fields were coming out blank, which turned out to be due to the `sampleId` not being set consistently throughout.

I can dig in further if you want.

pom.xml
arahuja: I'd agree that it's not worth it to factor out for a single dependency, but once there seems useful, easy to get a read on the major dependencies and their version ... but probably fine as is.

src/test/resources/tough.golden.vcf
arahuja: No that makes sense, OOC  - did these positions match the test positions before or are they different/whatever was output?  It's possible more positions were output than the test positions.

pom.xml
ryan-williams: Yea, I don't think any of these qualify as "major dependencies".

The two that do, spark and hadoop, coincidentally (or not!) are used in multiple places.

ryan-williams:
No, coveralls is clearly confused on this PR.

It just occurred to me that it's likely to do with my caching `target` here; coverage-info from previous runs is hanging around.

Let me either make sure that doesn't happen, or remove that change.

In general I think zinc is reliable enough at this point that we could save a lot of time not recompiling everything every build, and only rarely (if ever) having to clear caches; having repros for such zinc bugs as cause us to need to clear caches could be useful anyway.

src/test/resources/tough.golden.vcf
ryan-williams: This predates this PR, but all 25 positions that [used to be checked in SomaticStandardRealDataSuite](https://github.com/hammerlab/guacamole/commit/6711f4adb203d96fa14ba7403b1a4ba68ad0ba37#diff-14a7269ee8be3937ec72f27c586af285L90) are here, as well as 9 more.

ryan-williams:
OK, I think I fixed coverealls.

ryan-williams:
> I'd say just wait for/ask for a new ADAM release then managing one though

There's no "managing" to be done; we have the equivalent of a `SNAPSHOT` but that won't break under us.

PR #587, opened 2016-09-27
improvements / bugfix in capped-regions partitioner

ryan-williams:
- surface alternate coverage-computation strategy, "exploding" reads into per-base, depth-1 tuples, controllable by a flag
  - opens a part of the performance-space where memory-use is not proportional to maximum depth of reads, partly inspired by discussions on #572 
- fix a bug with how read-starts are tallied given `loci` strings that start and stop in the interior of contigs
- rename the loci-partitioners

### DIFF
--- a/src/main/scala/org/hammerlab/guacamole/readsets/iterator/ContigCoverageIterator.scala
+++ b/src/main/scala/org/hammerlab/guacamole/readsets/iterator/ContigCoverageIterator.scala
@@ -7,9 +7,21 @@ import org.hammerlab.guacamole.reference.{ContigIterator, Interval, Locus}
 import scala.collection.mutable
 
 /**
- * [[SkippableLociIterator]] that consumes reference regions (sorted by start-position, and all on the same contig) and
- * emits a [[Coverage]] at each covered locus, indicating how many reads overlap (and start at) that locus. Overlaps
- * are given a grace-margin of `halfWindowSize`.
+ * [[SkippableLociIterator]] that consumes [[Interval]]s (sorted by start-position) and emits a [[Coverage]] at each
+ * covered locus, indicating how many intervals overlap (and start at) that locus. Overlaps are given a grace-margin of
+ * `halfWindowSize`.
+ *
+ * One ([[Locus]], [[Coverage]]) tuple is emitted for each locus where the underlying intervals (and `halfWindowSize`)
+ * yield a non-empty [[Coverage]]: intervals contribute 1 to [[Coverage.depth]] for each locus they overlap
+ * (± halfWindowSize) and 1 to [[Coverage.starts]] for the first such locus.
+ *
+ * In cases where this iterator's [[SkippableLociIterator.skipTo]] method is called while it is emitting elements (e.g.
+ * when this iterator is made to proceed through loci discontinuously, to restrict calculation to specific loci-ranges),
+ * [[ContigCoverageIterator]] can land in the middle of some of its input [[Interval]]s, emitting their first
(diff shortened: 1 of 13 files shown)


src/main/scala/org/hammerlab/guacamole/readsets/iterator/ContigCoverageIterator.scala
```diff
@@ -53,11 +52,28 @@ case class ContigCoverageIterator(halfWindowSize: Int,
       else
         lastStart = start
 
-      ends.enqueue(end)
-      numAdded += 1
+      if (end + halfWindowSize > locus) {
```
arahuja: I think I follow this change but not 100%, could you add some comments here about what should be happening?

src/main/scala/org/hammerlab/guacamole/readsets/rdd/CoverageRDD.scala
```diff
@@ -120,24 +114,55 @@ class CoverageRDD[R <: ReferenceRegion: ClassTag](@transient rdd: RDD[R])
...
+    // Compute the bounds of loci that this region should contribute 1 unit of coverage-depth to.
+    val lowerBound = math.max(0, start - halfWindowSize)
+    val upperBound = end + halfWindowSize
 
-      // Compute the bounds of loci that this region should contribute 1 unit of coverage-depth to.
-      contigLength = contigLengthsBroadcast.value(contigName)
-      lowerBound = math.max(0, start - halfWindowSize)
-      upperBound = math.min(contigLength, end + halfWindowSize)
+    // Iterator over loci spanned by the current region, including the half-window buffer on each end.
+    val readLocusIterator = new LociIterator(Iterator(Interval(lowerBound, upperBound)).buffered)
```
arahuja: nit: mix of `region` and `read` here ... I don't love our use of `region` as our name for things with reference coordinates ... but until there is something better.

src/test/scala/org/hammerlab/guacamole/readsets/rdd/CoverageRDDSuite.scala
```diff
@@ -26,88 +24,41 @@ class CoverageRDDSuiteRegistrar extends KryoTestRegistrar {
   }
 }
 
-class RegionRDDSuite
+class CoverageRDDSuite
   extends GuacFunSuite
     with RegionsRDDUtil
     with ContigLengthsUtil {
 
   override def registrar = "org.hammerlab.guacamole.readsets.rdd.CoverageRDDSuiteRegistrar"
```
arahuja: Doesn't matter much here since this class is defined in this file, but you can do `classOf[CoverageRDDSuiteRegistrar].getCanonicalName` to protect against re-namings or packages moving around.

src/main/scala/org/hammerlab/guacamole/readsets/iterator/ContigCoverageIterator.scala
ryan-williams: yea, good idea, added a bunch!

also while explaining myself, found an edge case that I still wasn't handling correctly here, so fixed and added a test for it. thanks for prompting!

src/main/scala/org/hammerlab/guacamole/readsets/rdd/CoverageRDD.scala
ryan-williams: yea good point, done

src/test/scala/org/hammerlab/guacamole/readsets/rdd/CoverageRDDSuite.scala
ryan-williams: great idea, changed here and elsewhere

PR #589, opened 2016-09-28
lots of pom, shading, scripts improvements

### LINKED ISSUE #588: Can't use guacamole in spark-shell due to guava-shading nuances
[`Interval` exposes `com.google.common.collect.Range` in its public API](https://github.com/hammerlab/guacamole/blob/5fff3ea1bfbe877870a87d68d49b5b09c0a8a7ef/src/main/scala/org/hammerlab/guacamole/reference/Interval.scala#L37), which in some way means that that reference doesn't get [relocated by the shade plugin](https://github.com/hammerlab/guacamole/blob/5fff3ea1bfbe877870a87d68d49b5b09c0a8a7ef/pom.xml#L209-L215).

I don't fully understand why; when I package an assembly (or guac) JAR with `-Puber` (resp. `-Pguac`), and extract the class files from it, I see an `org/hammerlab/guacamole/reference/Interval.class` that contains, in ASCII, `toJavaRange.()Lorg/hammerlab/guavarelocated/collect/Range`, and no instances of `google`, which I'd have thought meant that the shading was proceeding correctly / as intended.

Yet, when I run:

```bash
spark-shell --jars target/guacamole-with-deps-0.0.1-SNAPSHOT.jar
```

then in the shell I get this error:

```
scala> org.hammerlab.guacamole.reference.Interval(10, 20)
error: bad symbolic reference. A signature in Interval.class refers to term collect
in package com.google.common which is not available.
It may be completely missing from the current classpath, or the version on
the classpath might be incompatible with the version used when compiling Interval.class.
```

I think this has to do with the "classes exposed in public API can't be shaded" issue that [I've heard tell of around Spark](https://issues.apache.org/jira/browse/SPARK-4809?focusedCommentId=14578457&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14578457). I've tried addressing it by *not* shading `Range`, or by shading but not relocating it, and in both cases doing the same to a various groups of classes around it, but in the end it seems that [our use of `Range.closedOpen`](https://github.com/hammerlab/guacamole/blob/58ac26f8477b1cce1b608732f7d5eaa456922faf/src/main/scala/org/hammerlab/guacamole/loci/set/Contig.scala#L27) (which only showed up in Guava 14.0.1, while Hadoop 2.* pins us with an un-shaded Guava 11.0.2), means that it's impossible to make this work.

One possible escape-hatch is [the "user classpath first" Spark configs](http://spark.apache.org/docs/latest/configuration.html#runtime-environment), but even in the latest Spark release they are deemed experimental, and IME they've accordingly introduced other problems when I've tried to use them in similar situations in the past, so I'm not eager to hitch ourselves to them.

My only other idea for proceeding is to remove that `Range` reference from `Interval`'s public API, further characterizing what kinds of things we can and can't shade.

I'd love to get to the bottom of how the shell is even seeing a reference to `com.google.common.collect`, allegedly in a `.class` file that doesn't seem to contain any such references, but that's further in the weeds so not my first plan of attack.

ryan-williams:
- remove Guava `Range` from public APIs, fix/limit guava shading; fixes #588 
- fix conflicts between and some incorrect shading across `guac`, `deps`, and `uber` profiles.
- make Spark and Hadoop `provided` by default, which reduces JAR size significantly.
  - expose `local` profile that un-provides them, useful only for running via `java` instead of `spark-*` wrappers.
- make default `guac` profile output paths to all transdeps to `target/dependencies`.
  - rework all scripts to build a full guac+deps classpath from thin guac jar plus this dependencies-file.
- rm misc other obsolete params/classes.


pom.xml
```diff
@@ -21,37 +21,43 @@
...
         <version>2.4.3</version>
         <configuration>
+
           <transformers>
+            <!-- Set "main" class. -->
             <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
               <mainClass>org.hammerlab.guacamole.Main</mainClass>
             </transformer>
-
-            <!-- Akka (which Spark depends on) needs this reference.conf munging in order to work correctly in -->
```
arahuja: Did you test with these removals?

arahuja: Nevermind, see this is just done later

arahuja:
Looked fine to me - but looks like there are some travis issues?

pom.xml
ryan-williams: yea we only need this in JARs that shade in spark/hadoop, so I put it in the `local` profile since they're `provided` otherwise

ryan-williams:
Yea, something about how a dep conflict was being resolved was apparently shaken loose with making the spark/hadoop deps `provided` by default; I think I've got it explicitly set in a consistent/working way now… thanks for the review, lmk if you have any thoughts about how I resolved it.

PR #593, opened 2016-09-30
use coarser loci-ranges in TakeLociIterator

ryan-williams:
improves performance of capped-regions partitioner in cases where coverage is sparse; pathological cases would end up generating `LociSet`s with many 1000s of ranges comprising them.


src/main/scala/org/hammerlab/guacamole/loci/set/TakeLociIterator.scala
```diff
@@ -97,8 +103,9 @@ class TakeLociIterator(it: BufferedIterator[(Position, Coverage)],
    * @return The number of covering regions, and a [[Contig]] with the taken loci ranges.
    */
   def takeLoci(it: ContigIterator[(HasLocus, Coverage)],
-            numRegions: Int,
-            dropAbove: Int): Option[(Int, Contig)] = {
+               numRegions: Int,
+               dropAbove: Int,
+               trimRanges: Boolean): Option[(Int, Contig)] = {
```
arahuja: Add this to the doc above?

arahuja: Add this to the doc above?

src/main/scala/org/hammerlab/guacamole/loci/set/TakeLociIterator.scala
```diff
@@ -112,16 +119,16 @@ class TakeLociIterator(it: BufferedIterator[(Position, Coverage)],
...
+      if (intervals.isEmpty)
+        None
+      else
+        Some(
+          curNumRegions ->
+            Contig(
+              it.contigName,
+              intervals
+            )
+        )
```
arahuja: Not sure what the diff is here? just spacing?

src/main/scala/org/hammerlab/guacamole/loci/set/TakeLociIterator.scala
ryan-williams: it was extraneous, this value comes in to the class via the constructor, I've removed it, good catch

src/main/scala/org/hammerlab/guacamole/loci/set/TakeLociIterator.scala
ryan-williams: yea this `if`/`else` was not indented from the `def` it implements before, and it should be

ryan-williams:
> `trimRanges` is added by never set to true other than the tests - is that expected?

weird, I thought I'd added a cmdline-flag for it, but can't find it any of my branches, maybe I hadn't at all; anyway, I added it now, thanks for catching!

Issue #551, opened 2016-08-15
Remove `progress` helper, use Spark/log4j


ryan-williams:
Spark's log4j wrapper is already used in a few places for e.g. logging warnings, and we should fold our logging into that infrastructure, where we currently have [a tiny homegrown `progress` helper](https://github.com/hammerlab/guacamole/blob/b638a11561686162245779673900c19e8f11fb63/src/main/scala/org/hammerlab/guacamole/logging/LoggingUtils.scala#L13).

The former supports a lot more customizability (setting levels per package/class-name, lazily evaluating expressions iff they are being logged at a level that is displayable, etc.).

PR #595, opened 2016-09-30
LociPartitioning cleanups

ryan-williams:
see commits


src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala
```diff
@@ -53,24 +60,31 @@ trait LociPartitionerArgs
...
   }
 
   @Args4JOption(
     name = "--partitioning-stats",
-    usage = "Compute additional statistics about the partitioned loci and reads; causes additional Spark jobs to be run, " +
-      "so should be disabled in performance-critical environments. Default: false.",
+    usage =
+      "Compute additional statistics about the partitioned loci and reads; causes additional Spark jobs to be run, " +
+        "so should be disabled in performance-critical environments. Default: false.",
     handler = classOf[BooleanOptionHandler]
```
arahuja: OOC - does this add something on top of the usual boolean arg situation?

src/main/scala/org/hammerlab/guacamole/loci/partitioning/CappedRegionsPartitioner.scala
```diff
@@ -131,4 +73,69 @@ class CappedRegionsPartitioner[R <: ReferenceRegion: ClassTag](regions: RDD[R],
...
+        val total = runs.map(_._1.toLong).sum
+        s"${runs.length} runs, $total loci (avg %.1f): $rs".format(total.toDouble / runs.length)
+      }
+    }
+
+    val totalLoci = validLoci + invalidLoci
+
+    val loci = lociBroadcast.value
+
+    progress(
```
arahuja: Any reason to come back to `progress` here vs going more in the direction of #551

src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala
ryan-williams: ha no I guess that is the default handler anyway, removed, good catch!

src/main/scala/org/hammerlab/guacamole/loci/partitioning/CappedRegionsPartitioner.scala
ryan-williams: well, this `progress` statement was already there, and it's still consistently used around the codebase, don't want to start mixing log-formats, just going to switch it all in one fell swoop at some point

PR #597, opened 2016-09-30
fold halfWindowSize into args plumbing

ryan-williams:


ryan-williams:
Sure, sorry I didn't provide more context.

I'm diving back into some loci-partitioning stuff to explore sending only the necessary reference material to each partition, rather than Broadcasting it per-contig.

The halfWindowSize treatment through the loci-partitioning logic felt like it was not encapsulated correctly:
- one of the three loci-partitioners use it while the other two don't,
- all callers of loci-partitioning code except for GermlineAssemblyCaller set it to `0` without even offering a flag to configure it,
- yet it was being plumbed through as a top-level parameter in the LociPartitioner and PartitionedRegions APIs, and hard-coded to `0` in callers that don't care about it.

The API I'm trying to support for loci-partitioners is:
- everything specific to each partitioner gets passed to that partitioner as a ctor arg, and is hidden from the general LociPartitioner interface.
- things that are common to all partitioners can be part of the shared interface.
  - currently, that's just [the `LociSet => LociPartitioning` shape](https://github.com/hammerlab/guacamole/blob/64845185cd79fefad5e84b6151041d38ab5ab9d5/src/main/scala/org/hammerlab/guacamole/loci/partitioning/LociPartitioner.scala#L77).

Assembly callers want to map their `--assembly-window-range` parameter to the half-window-size, so I've done this by having CappedRegionsPartitionerArgs mix-in a `halfWindowSize` method that `AssemblyArgs` overrides and maps that flag to.

In a forthcoming PR I add a thin "partition-loci" top-level guac command that wants to do the same thing but with a different flag, so mixing-in and overriding `HalfWindowConfig` is a way to let multiple commands plumb half-windows through the loci-partitioning logic while allowing them to configure the that value through CLI flags (or other logic) of their choice.

PR #600, opened 2016-10-02
fold AffineGapPenaltyAlignment into ReadAlignment

ryan-williams:
AffineGapPenaltyAlignment was a standalone object with one public
method (in spirit; scoreAlignmentPaths was public but shouldn't/needn't have
been. I made it package-private here so that tests can access it) that
was the main way of creating a ReadAlignment; arguably that makes it
just a ReadAlignment constructor, so I've folded it into ReadAlignment
accordingly.

I also package-privated the direct ReadAlignment constructor and moved AlignmentState to its own file.


ryan-williams:
merging this in sans review

PR #603, opened 2016-10-03
add `partition-loci` cmd

ryan-williams:
reuses all the same arguments, partitions loci (optionally writing it to `--loci-partitioning` path), and exits


ryan-williams:
Yea, they're somewhat debug-y, but the logical things they do make sense to expose downstream, e.g. to workflow managers that might be interested in finer-grained composition of these pieces, or QC/viz tools that might do analytics on the data they generate; keeping several logical workflow-stages coupled within larger Spark apps starts to push workflow-management into Spark, which is not where we want to be.

Plus, we don't have a ton of Commands bearing out the space of possibilities atm, so I think a couple composable proof-of-concept units in there is reasonable.

PR #605, opened 2016-10-04
Remove PileupFilter and revamp InputFilters

### LINKED ISSUE #135: QualityAlignedReadsFilter is wasteful and hurts read/task distribution
The current QualityAlignedReadsFilter removes all reads below some alignment quality threshold, but does so on a pileup by pileup basis.  We could remove all of these reads upfront.  This would help avoid pileups we will skip anyways and help with read distribution across tasks (as we would have a more accurate count of reads that will actually be processed)

The caveat is, even if we remove these reads, we use these reads to compute a measure of complexity - % reads below some alignment quality defines complexity currently.  But this filter hasn't proved to be that useful, and can probably be replaced with something else.

arahuja:
This PR:
- removes `PileupFilter` files and usage from `SomaticStandard`
- Add `ReadFilterArgs` which has flags to set properties of `InputFilters`
  - This includes moving around some of the arg traits to have `LociArgs` and `ReadFilterArgs` tied to read inputs hierarchy as opposed to partitioning.
  - Reaching way back -  closes #135!
- Revamping some of `InputFilters` to tie in with arg-parsing and removing some defaults
   - Part of this includes having `LociParser` result in `None` when no loci are specified and removing the `fallback` arguments as sometimes this strings were incorrectly specified.
  - Another piece was removing the default `InputFilters` in a few cases. In https://github.com/hammerlab/guacamole/commit/3bcca667bd506a8eba5f85951f2856ccbe899f70 `SomaticStandard` went from removing duplicate and failed reads to including them because of a default filter argument, so I wanted to avoid that.


ryan-williams:
Looks great, thanks for cleaning all this up.

While reviewing I was inspired to do a further loci-parsing refactor that I'd attempted and failed a few times in the past, but finally cracked this time: #606.

A couple small things in #606 are separable and more like bona-fide tweaks to what is here, but on the whole it's a separate PR and can be reviewed/merged as such.

I opened #606 against this in case you want to merge it in here, but if not you can just merge this and I'll send out #606 as a fresh PR.

PR #606, opened 2016-10-06
loci-parsing refactor

ryan-williams:
PR on top of #605:

- LociParser → ParsedLoci, removed state+mutation pattern, leaving an
  object that just wraps loci, after we've parsed strings, but before
  we have contig-lengths / a LociSet
  - uses ParsedLociRange and ParsedLociRanges classes, which
    encapsulate "all loci" special-casing.
  - parseLoci method is moved in there, since it's basically a
    constructor for a ParsedLoci.
- removed all non-test calls of unsafe ParsedLoci.result and
  LociSet(String), that assume no contig-lengths are needed because
  provided ranges will be closed.
- changed some callers to reflect that the ReadSets() call now
  integrates building loci and filters in addition to loading
  read-sets.
- move a couple of InputFilters ctors that were only used in tests to
  TestInputFilters


src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
```diff
@@ -291,7 +291,7 @@ object ReadSets extends Logging {
       result =
         result.filter(
           _.asMappedRead
-                .forall(_.alignmentQuality >= filters.minAlignmentQuality)
+                .exists(_.alignmentQuality >= filters.minAlignmentQuality)
```
arahuja: This is slightly different, this would mean that unmapped reads would be filtered.  I thought filtering unmapped reads should be separated through the range filters or passing `--mapped-only`

src/test/scala/org/hammerlab/guacamole/readsets/io/TestInputFilters.scala
```diff
@@ -0,0 +1,30 @@
+package org.hammerlab.guacamole.readsets.io
+
+import org.hammerlab.guacamole.loci.parsing.ParsedLoci
+
+object TestInputFilters {
```
arahuja: 👍  This was annoyingly kept around even though it was only used in the tests, but I wasn't up for refactoring all of those!

src/main/scala/org/hammerlab/guacamole/readsets/ReadSets.scala
ryan-williams: good point, my mistake! fixed.

PR #608, opened 2016-10-21
Add loci set intersection and difference methods

arahuja:


ryan-williams:
I just tried restarting the travis build that had failed; confused by what the two builds being run are and why that one failed.

arahuja:
@ryan-williams I think this failed due to the github issues Friday. I restarted one yesterday which passed and then forgot to restart the other

ryan-williams:
lgtm. sent #609 with some nits

PR #612, opened 2016-11-04
Upgrades and default settings to SomaticStandard

arahuja:
This PR updates `SomaticStandard` to:
- compute likelihoods for tumor variants by taking into account empirical (possibly low) VAFs
- add separate thresholds for tumor and normal LOD
- sets reasonable default on filters and threshold to ensure a good variant set
- updates the tests to reflect these


ryan-williams (changes requested):
lgtm, just 2 q's about the `1e-10`s laying around, also sent https://github.com/arahuja/guacamole/pull/1 with some subjective cleanups

src/main/scala/org/hammerlab/guacamole/likelihood/Likelihood.scala
```diff
@@ -157,7 +157,7 @@ object Likelihood {
             case (allele, alleleFraction) =>
               alleleElementProbabilities(alleleToIndex(allele), ::) * alleleFraction
           }
-          sum( log( sum(alleleRows) )) + math.log(prior(genotype)) * depth
+          sum( log( sum(alleleRows) + 1e-10 )) + math.log(prior(genotype))
```
arahuja: IIRC, if `sum(alleleRows) == 0` (possible if you are using the alignment likelihood and all are empty) then `log` gives a NaN that propagates.

ryan-williams: OK, different question here: it seems to me that the inner `sum` here is being applied to probabilities, not in log-space, which I don't understand. Is that a bug or am I missing something?

ryan-williams: I'm still chewing on this `1e-10`; I notice that removing it allows the `epsilon` error in `LikelihoodSuite` to be `1e-12` instead of `1e-5`. The latter seems unacceptably high for the computations we're doing here.

ryan-williams: At the very least I think we should put together a test case for the `sum(alleleRows) == 0` scenario, lmk if you have ideas about how to arrange that.

ryan-williams: OK, I've convinced myself that `sum(alleleRows) == 0` should be impossible unless there are what I'd call bugs elsewhere; what did you mean by "and all are empty" above?

ryan-williams:
Superseded by #613

arahuja:
Is https://github.com/arahuja/guacamole/pull/2 not needed then? Was this merged in in #613?

ryan-williams:
That's right, sry I didn't garbage-collect!

